Issue #35, opened 2017-05-26
Configurable `JsonFactory` and `HttpTransport`


hiranya911:
Admin Java SDK uses a default `JsonFactory` and a default `HttpTransport` everywhere. But users should be allowed to configure these where appropriate. For instance, when the SDK is deployed behind a proxy, it becomes unable to make any HTTP requests, since the default `HttpTransport` is setup without proxy support. If this was configurable, users can init a transport with proxy support and pass it in.

I believe this needs to happen in 2 places:
* Expose the credential factory methods that accept `JsonFactory` and `HttpTransport` arguments. These methods already exist, but are package-protected.
* Allow setting a `JsonFactory` and `HttpTransport` via `FirebaseOptions`. These can be plumbed into everywhere else in the SDK.

It is possible to have the credentials also take the parameters from `FirebaseOptions`, but that's going to be a breaking change, and we probably can live without it.

hiranya911:
Released in [v5.2.0](https://firebase.google.com/support/release-notes/admin/java#5.2.0)

PR #38, opened 2017-06-06
Making JsonFactory and HttpFactory Configurable

hiranya911:
* Exposed protected factory methods in `FirebaseCredentials`, which accept `JsonFactory` and `HttpTransport` arguments.
* Added `setJsonFactory()` and `setHttpTransport()` methods to `FirebaseOptions`.

Related to #35

### DIFF
--- a/src/main/java/com/google/firebase/FirebaseOptions.java
+++ b/src/main/java/com/google/firebase/FirebaseOptions.java
@@ -16,7 +16,6 @@
 
 package com.google.firebase;
 
-import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.api.client.googleapis.util.Utils;
@@ -146,6 +145,8 @@ public final class FirebaseOptions {
       databaseUrl = options.databaseUrl;
       firebaseCredential = options.firebaseCredential;
       databaseAuthVariableOverride = options.databaseAuthVariableOverride;
+      httpTransport = options.httpTransport;
+      jsonFactory = options.jsonFactory;
     }
 
     /**
--- a/src/main/java/com/google/firebase/FirebaseOptions.java
+++ b/src/main/java/com/google/firebase/FirebaseOptions.java
@@ -19,6 +19,9 @@ package com.google.firebase;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
+import com.google.api.client.googleapis.util.Utils;
+import com.google.api.client.http.HttpTransport;
+import com.google.api.client.json.JsonFactory;
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Objects;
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/firebase/FirebaseOptions.java
```diff
@@ -71,6 +76,26 @@ FirebaseCredential getCredential() {
...
+   * @return A Google API client JsonFactory instance.
+   */
+  @NonNull
+  public JsonFactory getJsonFactory() {
+    return jsonFactory;
+  }
+
   @Override
   public boolean equals(Object obj) {
     if (!(obj instanceof FirebaseOptions)) {
```
mikelehen: Should equals / hashCode / toString need to be updated?

src/main/java/com/google/firebase/auth/FirebaseCredentials.java
```diff
@@ -80,8 +79,26 @@ public static FirebaseCredential applicationDefault() {
...
+   *
+   * <p>See <a
+   * href="https://developers.google.com/identity/protocols/application-default-credentials">Google
+   * Application Default Credentials</a> for details on Google Application Deafult Credentials.
+   *
+   * <p>See <a href="/docs/admin/setup#initialize_the_sdk">Initialize the SDK</a> for code samples
+   * and detailed documentation.
+   *
+   * @param transport HttpTransport used to communicate with the remote authentication server.
+   * @param jsonFactory JsonFactory used to parse JSON responses from the remote authentication
```
mikelehen: from the remote authentication ....   server?

src/main/java/com/google/firebase/FirebaseOptions.java
hiranya911: It looks like these were implemented to support some persistence requirements, which are not really used in the SDK anymore. In the future I can see us adding even more options to this class, and having to implement all these methods each time we make a change is a bit of a problem. WDYT? Should we just get rid of the implementations?

`FirebaseApp` also implements these methods, but at least it performs all comparisons based on the name of the app.

src/main/java/com/google/firebase/auth/FirebaseCredentials.java
hiranya911: Done

hiranya911:
Fixed the typo, and left a follow up question.

src/main/java/com/google/firebase/FirebaseOptions.java
mikelehen: Deleting it sounds good to me.

hiranya911: I will do this in a separate PR. The existing implementations in `FirebaseOptions` are ok for now.

mikelehen: Please do this sooner than later.  Leaving broken equals / hashCode code around is a good recipe for unhappiness.

PR #44, opened 2017-06-09
Add some useful badges with automatic version detection: Maven and Javadoc

valery1707:


hiranya911:
Thanks for pull request. The [javadoc.io link](https://www.javadoc.io/doc/com.google.firebase/firebase-admin) points to a page with an `AccessDenied` error for me. Any idea why that is? Perhaps we need to fix our javadoc distribution for this to work.

hiranya911:
Our javadoc distribution has all the generated HTML files nested  in a `docs/reference` subdirectory. I think we need to get rid of that.

valery1707:
@hiranya911 fixed

PR #51, opened 2017-06-29
Google Cloud Storage API [WIP]

hiranya911:
Exposes Google Cloud Storage API via the Java Admin SDK:
* Take a dependency on GCS Java client
* Introducing `StorageClient` to hold an instance of `com.google.cloud.storage.Storage`.
* Expose the GCS buckets via a new `getBucket()` method.
* Expose a new `getDownloadUrl()` method (pending review).
* Implement support for a new `storageBucket` app option.

go/firebase-admin-gcs
go/admin-gcs-download-url [Pending review]


pom.xml
```diff
@@ -330,6 +330,11 @@
             <version>20.0</version>
         </dependency>
         <dependency>
+            <groupId>com.google.cloud</groupId>
+            <artifactId>google-cloud-storage</artifactId>
+            <version>1.2.1</version>
```
hiranya911: At the final API review it was decided to make this a hard dependency. If in the future we change our mind about that decision, all we have to do is to add `<scope>provided</scope>` to the above config.

src/main/java/com/google/firebase/FirebaseOptions.java
```diff
@@ -139,6 +151,22 @@ public Builder setDatabaseUrl(@Nullable String databaseUrl) {
...
+     * used to access the bucket.
+     *
+     * <p>See <a href="/docs/admin/setup#initialize_the_sdk">Initialize the SDK</a> for code samples
+     * and detailed documentation.
+     *
+     * @param storageBucket The name of an existing Google Cloud Storage bucket.
+     * @return This <code>Builder</code> instance is returned so subsequent calls can be chained.
+     */
+    public Builder setStorageBucket(String storageBucket) {
+      this.storageBucket = storageBucket;
```
hiranya911: Done

src/main/java/com/google/firebase/auth/FirebaseCredentials.java
```diff
@@ -50,7 +50,8 @@
       ImmutableList.of(
           "https://www.googleapis.com/auth/firebase.database",
           "https://www.googleapis.com/auth/userinfo.email",
-          "https://www.googleapis.com/auth/identitytoolkit");
+          "https://www.googleapis.com/auth/identitytoolkit",
+          "https://www.googleapis.com/auth/devstorage.full_control");
```
hiranya911: Done

src/main/java/com/google/firebase/cloud/StorageClient.java
```diff
@@ -0,0 +1,144 @@
...
+import com.google.firebase.internal.FirebaseService;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+/**
+ * StorageClient provides access to Google Cloud Storage APIs. You can specify a cloud storage
+ * bucket via {@link com.google.firebase.FirebaseOptions}, and then get a reference to it by calling
+ * {@link StorageClient#getBucket()}. Or if you know the bucket name at runtime you can
+ * directly call {@link StorageClient#getBucket(String)}.
```
hiranya911: Done

src/main/java/com/google/firebase/cloud/StorageClient.java
```diff
@@ -0,0 +1,144 @@
...
+ *
+ * <p>This class requires Google Cloud Storage libraries for Java. Make sure the artifact
+ * google-cloud-storage is in classpath along with its transitive dependencies.
```
hiranya911: Added "the"

src/main/java/com/google/firebase/cloud/StorageClient.java
```diff
@@ -0,0 +1,144 @@
...
+ */
+public class StorageClient {
+
+  private static final String DOWNLOAD_TOKENS = "firebaseStorageDownloadTokens";
```
hiranya911: Done

src/main/java/com/google/firebase/cloud/StorageClient.java
```diff
@@ -0,0 +1,144 @@
...
+
+    StorageClientService(FirebaseApp app) {
+      super(SERVICE_ID, new StorageClient(app));
+    }
+
+    @Override
+    public void destroy() {
+      // NOTE: We don't explicitly tear down anything here, but public methods of StorageClient
+      // will now fail because calls to getOptions() and getToken() will hit FirebaseApp,
+      // which will throw once the app is deleted.
```
hiranya911: `getBucket()` calls `app.getOptions()`. Also `StorageClient` is authenticated with a `FirebaseOAuthCredentials`, which calls `app.getToken()` internally. These `app` invocations fail when the app is deleted.

src/main/java/com/google/firebase/internal/FirebaseOAuthCredentials.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+
+/**
+ * A Google cloud credential implementation that uses OAuth2 access tokens obtained from a
+ * <code>FirebaseApp</code> to authenticate cloud API calls. This essentially acts as a bridge
+ * between Firebase Admin SDK APIs and the Google cloud <code>Credentials</code> API.
+ */
+public final class FirebaseOAuthCredentials extends Credentials {
```
hiranya911: Done

src/main/java/com/google/firebase/cloud/StorageClient.java
```diff
@@ -0,0 +1,144 @@
...
+   * @param blob The blob to generate a download URL for.
+   * @return a URL string.
+   */
+  public String getDownloadUrl(Blob blob) {
+    checkNotNull(blob, "Blob must not be null");
+    Map<String, String> metadata = blob.getMetadata();
+    if (metadata == null) {
+      metadata = new HashMap<>();
+    }
+    String downloadTokens = metadata.get(DOWNLOAD_TOKENS);
```
hiranya911: Download tokens can be any random string in Firebase Storage. The Firebase console uses the same exact logic as here to generate and store download tokens. When a user sends a request to a generated download URL, the Firebase Storage simply checks if the token embedded in the URL is also available in the object's metadata.

I think provided that only a privileged user can update an object's metadata, I don't think there's a security issue here. The only situation that may cause problems is an admin user accidentally removing or modifying this property. I don't feel like that's something to be too worried about. But to be in the safe side, I've updated the code here to URL escape the token string.

src/main/java/com/google/firebase/FirebaseOptions.java
```diff
@@ -139,6 +151,22 @@ public Builder setDatabaseUrl(@Nullable String databaseUrl) {
     }
 
     /**
+     * Sets the name of the Google Cloud Storage bucket for reading and writing application data.
+     * The same credential used to initialize the SDK (see {@link Builder#setCredential}) will be
+     * used to access the bucket.
+     *
+     * <p>See <a href="/docs/admin/setup#initialize_the_sdk">Initialize the SDK</a> for code samples
```
hiranya911: Changed to an absolute URL (here and elsewhere in the same file)

hiranya911:
Addressed most of the suggestions, and left a couple of follow up comments.

src/main/java/com/google/firebase/cloud/StorageClient.java
mikelehen: Cool, and since this key isn't listed in https://firebase.google.com/docs/storage/web/file-metadata#file_metadata_properties I assume normal Firebase Storage clients can't write it.

src/main/java/com/google/firebase/cloud/StorageClient.java
```diff
@@ -0,0 +1,149 @@
...
+ * storage bucket via {@link com.google.firebase.FirebaseOptions}, and then get a reference to this
+ * default bucket by calling {@link StorageClient#getBucket()}. Or you can get a reference to a
+ * specific bucket at any time by calling {@link StorageClient#getBucket(String)}.
+ *
+ * <p>This class requires Google Cloud Storage libraries for Java. Make sure the artifact
+ * google-cloud-storage is in the classpath along with its transitive dependencies.
+ */
+public class StorageClient {
+
+  // Key used tp store previously-created download tokens in a Blob's metadata.
```
mikelehen: tp => to

src/main/java/com/google/firebase/cloud/StorageClient.java
```diff
@@ -113,16 +118,16 @@ public String getDownloadUrl(Blob blob) {
...
       downloadTokens = UUID.randomUUID().toString();
-      metadata.put(DOWNLOAD_TOKENS, downloadTokens);
+      metadata.put(DOWNLOAD_TOKENS_METADATA_KEY, downloadTokens);
       blob.toBuilder().setMetadata(metadata).build().update();
     }
 
     String token = downloadTokens.split(",")[0];
     return String.format("https://firebasestorage.googleapis.com/v0/b/%s/o/%s?alt=media&token=%s",
-        blob.getBucket(), blob.getName(), token);
+        blob.getBucket(), blob.getName(), UrlEscapers.urlFormParameterEscaper().escape(token));
```
mikelehen: I think the blob name should be url-encoded too?  You may as well do the bucket too, though I think it's limited to characters that are already url-safe.

src/main/java/com/google/firebase/cloud/StorageClient.java
hiranya911: Done

src/main/java/com/google/firebase/cloud/StorageClient.java
hiranya911: Escaped the object name. Bucket names are URL safe as per https://cloud.google.com/storage/docs/naming#requirements

src/main/java/com/google/firebase/cloud/StorageClient.java
hiranya911: It actually goes in the `customMetadata` field. It seems the server simply ignores any updates made to the download token metadata field from a client SDK. 

@katfang could you please confirm?

src/main/java/com/google/firebase/cloud/StorageClient.java
```diff
@@ -0,0 +1,150 @@
...
+import com.google.firebase.FirebaseApp;
+import com.google.firebase.ImplFirebaseTrampolines;
+import com.google.firebase.internal.FirebaseCloudCredentials;
+import com.google.firebase.internal.FirebaseService;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+/**
+ * StorageClient provides access to Google Cloud Storage APIs. You can specify a default cloud
```
katfang: Does it have the ability to get *the* default bucket without having to pass in any information?

hiranya911: Not exactly. The developer has to either call `FirebaseOptions.setStorageBucket()` when initializing the app, or pass the bucket name explicitly to `StorageClient.getBucket()`

src/main/java/com/google/firebase/cloud/StorageClient.java
katfang: Yeah. If you try to write to it:

$ curl "https://firebasestorage.googleapis.com/v0/b/bucketfot.appspot.com/o/any%2Fthing" -X PATCH  -d '{"metadata":{"firebaseStorageDownloadTokens":"pleasetellmeicannnot"}}' -H "Content-Type: application/json"
{
  "error": {
    "code": 400,
    "message": "Not allowed to set custom metadata for firebaseStorageDownloadTokens"
  }
}

And when you clear metadata normally, the server clears all other metadata.

I think this API will allow the admin user to set/clear the custom metadata field so that there are no tokens (separate from this method). But I think that's allowable, and possibly reasonable if they for some reason wanted to set them to something specific. (One token that gives access to 40 files?)

PR #53, opened 2017-07-06
SLF4J Integration

hiranya911:
We wish to use [SLF4J](https://www.slf4j.org/) as the logging interface in the Admin SDK. 

* Replaced the `Log` class in the internal package with calls to SLF4J
* Deprecated the logging interfaces in the database package. These will be removed in a future release.
* Modified `LogWrapper` to log to SLF4J when available, and log to the old interface when not. This is a temporary solution to avoid the breaking change. When we remove the deprecated logging stuff in the future, we can get rid of this as well.


src/main/java/com/google/firebase/database/logging/LogWrapper.java
```diff
@@ -18,19 +18,36 @@
 
 import java.io.PrintWriter;
 import java.io.StringWriter;
+import org.slf4j.LoggerFactory;
 
-/** User: greg Date: 6/12/13 Time: 2:23 PM */
+/**
+ * Legacy logging interface for database implementation. This class attempts to reconcile SLF4J
+ * with the old logging implementation of Admin SDK. When SLF4J is available, logs using that
+ * API. Otherwise falls back to the old logging implementation. This class will removed in a
```
mikelehen: I assume we could just log to both but we're avoiding that because it would potentially result in duplicate logs for people?  Perhaps worth spelling that out in the comment.

src/main/java/com/google/firebase/database/FirebaseDatabase.java
```diff
@@ -263,6 +263,7 @@ public void goOffline() {
    * diagnostic logging, and {@link Logger.Level#NONE NONE} to disable all logging.
    *
    * @param logLevel The desired minimum log level
+   * @deprecated Use SLF4J-based logging. This method will be removed in a future release.
```
hiranya911: Done. Updated the text and added a link to SLF4J user manual.

src/main/java/com/google/firebase/database/logging/LogWrapper.java
hiranya911: Done.

Issue #56, opened 2017-07-12
FR: Implement Phone Auth Support in User Management API


hiranya911:
Work is currently underway in the Node Admin SDK to implement phone auth support (see [branch](https://github.com/firebase/firebase-admin-node/tree/admin_node_phone)). This needs to be ported to the Java Admin SDK. This will provide:
* ability to associate a phone number with a user account
* ability look up users by phone numbers
* ability to update user accounts with phone numbers

hiranya911:
This was released today: https://firebase.google.com/support/release-notes/admin/java#5.3.0

PR #60, opened 2017-08-02
Implementing Phone Auth Support

hiranya911:
* Added `FirebaseAuth.getUserByPhoneNumber()` method.
* Added `setPhoneNumber()` method to `CreateRequest` and `UpdateRequest` types.
* Exposing phone number via `UserInfo` and `UserRecord` types.

go/firebase-java-phone-auth

### DIFF
--- a/src/main/java/com/google/firebase/auth/UserRecord.java
+++ b/src/main/java/com/google/firebase/auth/UserRecord.java
@@ -196,7 +196,7 @@ public class UserRecord implements UserInfo {
     // normalize accordingly.
     checkArgument(!Strings.isNullOrEmpty(phoneNumber), "phone number cannot be null or empty");
     checkState(phoneNumber.matches("^\\+.*[a-zA-Z0-9].*$"),
-        "phone number must begin with a '+' sign and contain at least one alphanumeric character");
+        "phone number must be a valid, E.164 compliant identifier string");
   }
 
   private static void checkPassword(String password) {
--- a/src/main/java/com/google/firebase/auth/UserRecord.java
+++ b/src/main/java/com/google/firebase/auth/UserRecord.java
@@ -18,6 +18,7 @@ package com.google.firebase.auth;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
 
 import com.google.common.base.Strings;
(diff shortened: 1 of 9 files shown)


hiranya911:
Related to #56

src/main/java/com/google/firebase/auth/UserRecord.java
```diff
@@ -177,6 +191,14 @@ private static void checkEmail(String email) {
     checkArgument(email.matches("^[^@]+@[^@]+$"));
   }
 
+  private static void checkPhoneNumber(String phoneNumber) {
+    // Phone number verification is very lax here. Backend will enforce E.164 spec compliance, and
+    // normalize accordingly.
+    checkArgument(!Strings.isNullOrEmpty(phoneNumber), "phone number cannot be null or empty");
+    checkState(phoneNumber.matches("^\\+.*[a-zA-Z0-9].*$"),
```
schmidt-sebastian: If I read this correctly,  this merely verifies that the phone number starts with '+' and contains at least one 'normal' character. So +&&&a&&& might be a valid number? If that is our level of verification, then we might as well let the server handles it.

src/main/java/com/google/firebase/auth/UserRecord.java
```diff
@@ -324,6 +357,19 @@ public UpdateRequest setEmail(String email) {
     }
 
     /**
+     * Updates the phone number associated with this user.
+     *
+     * @param phone a valid phone number string or null.
+     */
+    public UpdateRequest setPhoneNumber(String phone) {
```
schmidt-sebastian: Do we not want to use Nullable here? (and everywhere else).

I feel we shouldn't be calling this out in our comments but rather in our code. Furthermore, it is not clear from the comment what setting the number to 'null' entails.

hiranya911: Done. Added `Nullable` to appropriate method arguments, and also updated the javadoc of `setPhoneNumber()`.

src/main/java/com/google/firebase/auth/UserRecord.java
hiranya911: I simplified this to check only for the `+` prefix. Updated the error message accordingly. We can rely on the backend server for more comprehensive validation of phone numbers.

hiranya911:
Made the suggested changes.

PR #64, opened 2017-09-05
Proper handling of exceptions when verifying ID tokens

hiranya911:


src/main/java/com/google/firebase/auth/internal/FirebaseTokenVerifier.java
```diff
@@ -58,7 +58,8 @@
   private static final String ISSUER_PREFIX = "https://securetoken.google.com/";
   private static final String FIREBASE_AUDIENCE =
       "https://identitytoolkit.googleapis.com/google.identity.identitytoolkit.v1.IdentityToolkit";
-  private static final String ERROR_CODE = "ERROR_INVALID_CREDENTIAL";
+  private static final String ERROR_INVALID_CREDENTIAL = "ERROR_INVALID_CREDENTIAL";
+  private static final String ERROR_RUNTIME_EXCEPTION = "ERROR_RUNTIME_EXCEPTION";
```
mikelehen: Are there more of these? Should they all be defined in a single place?

hiranya911: There are few more in [`FirebaseUserManager`](https://github.com/firebase/firebase-admin-java/blob/master/src/main/java/com/google/firebase/auth/FirebaseUserManager.java#L53). I don't feel like there's enough of them to start extracting them into one place. WDYT? Perhaps introduce a new `internal.Constants` class, which would hold all shared constants?

mikelehen: I think in general it's useful to centralize error codes, for auditing and documentation purposes and whatnot. But I don't feel strongly that that needs to happen in this PR. So feel free to just consider it for the future.

Issue #55, opened 2017-07-10
FR: Revising the Threading Model and Async APIs


hiranya911:
There are several issues in the Admin SDK with respect to how it creates and uses threads:

1. Most public APIs are asynchronous, and returns `Task` instances. This tends to make implementing simple use cases unnecessarily complex (e.g. creating a custom token and sending back as a servlet response).
2. Main thread pool is initialized statically, and not configurable.
3. `ThreadFactory` is initialized statically, and not configurable.
4. `Tasks` API is visible to user code. A user can kick off arbitrary background tasks using this API, which will get executed on the main thread pool of the SDK, un-regulated.

In order to address these limitations, we can consider the following changes:

* Expose synchronous/blocking variants of public APIs where it makes sense (most notably at `FirebaseAuth`).

```
BlockingFirebaseAuth auth = FirebaseAuth.getBlockingInstance();
String customToken = auth.createCustomToken("user1");
```

* Make the thread pool and thread factory configurable via app options.
* Hide or decommission the `Tasks` API. We can consider using the `ApiFutures` from Google Cloud Commons (this is similar to Guava's `ListenableFutures`) as a replacement.

```
FirebaseAuth auth = FirebaseAuth.getInstance();
ApiFuture<String> customTokenFuture = auth.createCustomToken("user1");

// The following should not be supported 
// Tasks.call(new Callable(){...}); 
```

hiranya911:
We are making some headway with this issue. We are thinking about a 2-phase roll out plan:

**Phase 1 (No breaking changes)**
* Introduce new async methods that return `ApiFuture` instances (e.g. `ApiFuture<String> createCustomToken()`.
* Deprecate the `Task` API

**Phase 2 (breaking changes)**
* Remove the methods that return `Task` instances.
* Replace them with a set of blocking methods (e.g. `String createCustomToken()`).

nilsonfreitas:
I believe that firebase is preventing my appengine basic scaling instances to be evicted. It seems that as long as a background thread is alive, instance will also be alive and won't respect the idle-timeout setting.

Also, every 12 hours I can see a crash on logs while appengine calls /_ah/background with bellow trace:

```java.lang.RuntimeException: Restarting Firebase Worker Thread
	at com.google.firebase.internal.RevivingScheduledExecutor.<clinit>(RevivingScheduledExecutor.java:43)
[...]

PR #66, opened 2017-09-08
Deprecated the Task API and introduced ApiFutures

hiranya911:
This is the first phase of migration to `ApiFutures` (go/firebase-admin-java-threads). This PR:

1. Deprecates the `Task` API.
2. Adds a set of new methods to `FirebaseAuth` that returns `ApiFutures`. 
3. Updates the test cases to use `ApiFutures` where appropriate.

I'm not touching the `FirebaseCredential` interface, since that is going to be deprecated separately.

`DatabaseReference` and `OnDisconnect` classes also have some methods that return `Tasks`. Any ideas on how to deal with them will be appreciated. We can introduce a set of new methods that return `ApiFutures` here as well.  But that results in the following migration path which is not ideal:

```
// Phase 1 (No breaking changes)
class DatabaseReference {
   public Task<Void> setValue(Object value);
   public ApiFuture<Void> setValueAsync(Object value);
}
```

```
// Phase 2 (Breaking changes)
class DatabaseReference {
   public ApiFuture<Void> setValue(Object value);
}
```

The other option is to not touch them at the moment, and continue to support only `Tasks` in those APIs. We will simply switch them to return `ApiFutures` in phase 2.

### DIFF
--- a/src/main/java/com/google/firebase/database/DatabaseReference.java
+++ b/src/main/java/com/google/firebase/database/DatabaseReference.java
@@ -16,6 +16,7 @@
 
 package com.google.firebase.database;
 
+import com.google.api.core.ApiFuture;
 import com.google.firebase.database.core.CompoundWrite;
 import com.google.firebase.database.core.DatabaseConfig;
 import com.google.firebase.database.core.Path;
@@ -32,6 +33,7 @@ import com.google.firebase.database.utilities.PushIdGenerator;
 import com.google.firebase.database.utilities.Utilities;
 import com.google.firebase.database.utilities.Validation;
 import com.google.firebase.database.utilities.encoding.CustomClassMapper;
+import com.google.firebase.internal.TaskToApiFuture;
 import com.google.firebase.tasks.Task;
 
 import java.io.UnsupportedEncodingException;
@@ -172,13 +174,11 @@ public class DatabaseReference extends Query {
    * <code>Map&lt;String, MyPOJO&gt;</code>, as well as null values.
    *
    * @param value The value to set at this location
-   * @return The {@link Task} for this operation.
+   * @return The ApiFuture for this operation.
    */
-  public Task<Void> setValue(Object value) {
-    return setValueInternal(value, PriorityUtilities.parsePriority(null), null);
+  public ApiFuture<Void> setValueAsync(Object value) {
+    return new TaskToApiFuture<>(setValue(value));
   }
-
-  // Set priority
 
   /**
    * Set the data and priority to the given values. Passing null to setValue() will delete the data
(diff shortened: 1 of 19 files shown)


hiranya911:
Related to #55

src/main/java/com/google/firebase/internal/TaskToApiFuture.java
```diff
@@ -0,0 +1,80 @@
...
+      public void onComplete(Task<T> task) {
+        runnable.run();
+      }
+    });
+  }
+
+  @Override
+  public boolean cancel(boolean mayInterruptIfRunning) {
+    // Cannot be supported with Tasks
+    return false;
```
mikelehen: I wonder if we should throw an exception instead since this will never do anything useful.  

Also, according to https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Future.html#cancel(boolean), "After this method returns, subsequent calls to isDone() will always return true." and I suspect that means that after calling cancel(), we shouldn't call their completion listener, etc. even if the actual work wasn't cancelled.

So if we don't throw, I think we need to do more work to follow the contract.

src/main/java/com/google/firebase/database/DatabaseReference.java
```diff
@@ -212,7 +212,30 @@ public DatabaseReference push() {
...
    * @param value The value to set at this location
    * @param priority The priority to set at this location
+   * @return The ApiFuture for this operation.
+   */
+  public ApiFuture<Void> setValueAsync(Object value, Object priority) {
+    return new TaskToApiFuture<>(setValue(value, priority));
+  }
+
+  /**
+   * Similar to {@link #setValueAsync(Object)} but returns a Task.
```
mikelehen: For FirebaseAuth, you replicated the javadoc comments, but here you are just saying "Similar to ..."  Intentional?

src/main/java/com/google/firebase/tasks/Task.java
```diff
@@ -25,6 +25,7 @@
  * Represents an asynchronous operation.
  *
  * @param <T> the type of the result of the operation
+ * @deprecated Use ApiFuture interface instead
```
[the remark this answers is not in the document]
hiranya911: Done

src/main/java/com/google/firebase/database/DatabaseReference.java
hiranya911: Good point. Applied the "Similar to.." convention in FirebaseAuth too.

src/main/java/com/google/firebase/internal/TaskToApiFuture.java
hiranya911: I think the use of `cancel()` is commonplace enough, we need to do a bit better here than throwing an exception. I'm adding a boolean member to the class, and fixing the code so we correctly implement the contract of `isDone()`. 

The `cancel()` method only "attempts" to stop the task, and it is ok to not be able to stop the execution according to the contract. So I don't think we should try to stop any callbacks from firing.

hiranya911:
Merged to m19 branch.

Issue #57, opened 2017-07-17
FR: Migrate Admin SDK to google-auth-library-java


hiranya911:
Currently the Admin SDK depends on the `google-api-client` library for all credential processing. But GCP at large is moving towards the new https://github.com/google/google-auth-library-java library. This library has better, cleaner APIs for parsing credential files, and integrates well with other gcloud SDKs. We should migrate the Admin SDK to use this new library.

This dependency is still at v0.6 and the API is subject to change. It might be a good idea to wait until they release v1.0.

hiranya911:
I've done some prototyping related to this in the hkj-auth-creds branch. It seems we can get rid of a significant amount of code by doing this migration. I'm also trying to get https://github.com/google/google-auth-library-java/issues/117 as part of this, which will make things even easier for us.

hiranya911:
This was released today: https://firebase.google.com/support/release-notes/admin/java#5.4.0

PR #71, opened 2017-09-12
Migrating to GoogleCredentials

hiranya911:
### Migrating to `GoogleCredentials`
1. Deprecated the `FirebaseCredential` and `FirebaseCredentials` interfaces.
2. Added `FirebaseOptions.Builder.setCredentials(GoogleCredentials)` method, which is the new recommended way to pass credentials to the SDK.
3. Using `GoogleCredentials` directly in `FirebaseAuth`, `CloudStorageClient` and other places that make authenticated HTTP calls.
4. Removed old logic for token caching, and token change notifications. All this is now handled by `GoogleCredentials`.

### Project ID Lookup
1. Added `FirebaseOptions.Builder.setProjectId(String)` method, which can be used to initialize the SDK with an explicitly specified project ID.
2. 3-level (options, credentials and environment) project ID lookup mechanism in `ImplOnlyFirebaseTrampolines`


hiranya911:
Resolves #57

src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -427,28 +338,49 @@ FirebaseService getService(String id) {
    * its own instance of this class. This class is not thread safe. The caller (FirebaseApp) must
    * ensure that methods are called serially.
    */
-  static class TokenRefresher {
+  static class TokenRefresher implements CredentialsChangedListener {
```
mikelehen: It seems like you've reworked the way this works significantly (we schedule refreshes which trigger changes which cause us to schedule refreshes). I think it's probably worth revisiting the class comment to make sure it explains the intended approach.

src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -427,28 +338,49 @@ FirebaseService getService(String id) {
...
+    public final synchronized void onChanged(OAuth2Credentials credentials) throws IOException {
+      if (closed) {
+        return;
+      }
+
+      AccessToken accessToken = credentials.getAccessToken();
+      long refreshDelay = accessToken.getExpirationTime().getTime()
+          - System.currentTimeMillis() - TimeUnit.MINUTES.toMillis(5);
+      if (refreshDelay > 0) {
+        scheduleRefresh(refreshDelay);
```
mikelehen: Do we need some kind of protection against having multiple outstanding scheduled refreshes?  Since any token change will cause us to schedule a new one, it seems like we could end up with an infinite number in degenerate cases (somebody just calls refresh on the credential a bunch of times or something).

On that note, what motivated this approach?  It seems more complicated / dangerous to trigger off of credential changes rather than the old approach.

mikelehen (commented):
I haven't gotten that far through reviewing this, but here's some initial feedback for discussion.

src/main/java/com/google/firebase/ImplFirebaseTrampolines.java
```diff
@@ -50,18 +73,9 @@ public static String getPersistenceKey(String name, FirebaseOptions options) {
...
-  public static void removeAuthStateChangeListener(
-      @NonNull FirebaseApp app, @NonNull AuthStateListener listener) {
-    app.removeAuthStateListener(listener);
-  }
-
-  public static Task<GetTokenResult> getToken(@NonNull FirebaseApp app, boolean forceRefresh) {
-    return app.getToken(forceRefresh);
+  public static void addCredentialsChangedListener(
+      @NonNull FirebaseApp app, @NonNull CredentialsChangedListener listener) {
+    app.addCredentialsChangedListener(listener);
```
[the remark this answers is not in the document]
hiranya911: Fair enough. Done.

src/main/java/com/google/firebase/database/core/JvmAuthTokenProvider.java
```diff
@@ -35,51 +35,43 @@
...
-                  listener.onError(task.getException().toString());
-                }
-              }
-            });
+    GoogleCredentials credentials = ImplFirebaseTrampolines.getCredentials(firebaseApp);
+    try {
+      if (forceRefresh) {
+        credentials.refresh();
+      } else {
+        credentials.getRequestMetadata();
```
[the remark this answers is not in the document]
hiranya911: Added a comment explaining what's going on here. Basically `getRequestMetadata()` gets a new token if one doesn't already exist, or refreshes the token if it is expired (a conditional refresh so to speak).

src/main/java/com/google/firebase/database/core/JvmAuthTokenProvider.java
```diff
@@ -104,14 +96,15 @@ private AuthStateListener wrap(TokenChangeListener listener) {
     }
 
     @Override
-    public void onAuthStateChanged(final GetTokenResult tokenResult) {
+    public void onChanged(OAuth2Credentials credentials) throws IOException {
       // Notify the TokenChangeListener on database's thread pool to make sure that
       // all database work happens on database worker threads.
+      final AccessToken accessToken = credentials.getAccessToken();
```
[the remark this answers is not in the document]
hiranya911: When `onChanged()` event fires, it is guaranteed that `credentials.getAccessToken()` can return a valid OAuth2 token. Added a comment to explain.

src/main/java/com/google/firebase/FirebaseApp.java
hiranya911: Good catch. Updated the class-level javadoc.

src/main/java/com/google/firebase/FirebaseApp.java
hiranya911: The `scheduleRefresh()` helper method cancels the last scheduled refresh event before scheduling a new one. I added this bit of info to the javadoc. Moreover:

1. `TokenRefresher` is now thread-safe. It will only handle one token change event at a time. This is further guaranteed by `GoogleCredentials` whose `refresh()` method and the event notification logic are all synchronized.
2. The new approach is not too far off from what we used to have. We used to cache the token, and run the following bit of code whenever we refreshed the credential:

```
if (currentToken.compareAndSet(oldToken, newToken)) {
      listenersCopy = ImmutableList.copyOf(authStateListeners);
      long refreshDelay  = googleOAuthToken.getExpiryTime() - clock.now()
             - TimeUnit.MINUTES.toMillis(5);
      if (refreshDelay > 0) {
             tokenRefresher.scheduleRefresh(refreshDelay);
      } 
     ....
}
```

Now that token caching and refreshing happens within `GoogleCredentials`, we rely on the `onChanged()` event to schedule future refresh events.

src/main/java/com/google/firebase/ImplFirebaseTrampolines.java
```diff
@@ -34,8 +34,31 @@
...
-    return app.getOptions().getCredential();
+  public static GoogleCredentials getCredentials(@NonNull FirebaseApp app) {
+    return app.getOptions().getCredentials();
+  }
+
+  public static String getProjectId(@NonNull FirebaseApp app) {
+    return getProjectId(app.getOptions());
+  }
+
+  public static String getProjectId(@NonNull FirebaseOptions options) {
```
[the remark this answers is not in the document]
hiranya911: Done. Moved the logic to `FirebaseApp`.

src/main/java/com/google/firebase/database/core/JvmAuthTokenProvider.java
```diff
@@ -35,51 +35,47 @@
...
+    GoogleCredentials credentials = ImplFirebaseTrampolines.getCredentials(firebaseApp);
+    try {
+      if (forceRefresh) {
+        credentials.refresh();
+      } else {
+        // getRequestMetadata() will refresh only if the credentials instance does not already
+        // have a valid (unexpired) OAuth2 token. GoogleCredentials does not provide a clearly
+        // labeled method for conditionally refreshing the underlying token. getRequestMetadata()
+        // is the closest thing to what we need.
+        credentials.getRequestMetadata();
```
mikelehen: Thanks for adding the comment. I now understand what's going on a lot better and I'd actually suggest you change this again (sorry) so that we /always/ call getRequestMetadata() before calling getAccessToken() [this should be harmless in the case that we just called .refresh() since it will detect it doesn't need to refresh again] and then change the comment to something like:

// The typical way to use a GoogleCredentials object is to call getRequestMetadata() and include the metadata in your request. Since we are accessing the token directly via getAccessToken(), we must first call getRequestMetadata() to ensure the token is available (refreshed if necessary).

And while you're at it, I'd explicitly do "String accessToken = credentials.getAccessToken();" rather than embed it inside the listener.onSuccess() call. It's important enough to warrant its own line of code. :-)

src/main/java/com/google/firebase/FirebaseApp.java
mikelehen: Cool, thanks for explaining.

src/main/java/com/google/firebase/database/core/JvmAuthTokenProvider.java
```diff
@@ -35,51 +35,47 @@
...
-              @Override
-              public void onComplete(@NonNull Task<GetTokenResult> task) {
-                if (task.isSuccessful()) {
-                  listener.onSuccess(wrapOAuthToken(firebaseApp, task.getResult()));
-                } else {
-                  listener.onError(task.getException().toString());
-                }
-              }
-            });
+    GoogleCredentials credentials = ImplFirebaseTrampolines.getCredentials(firebaseApp);
```
mikelehen: Like my feedback elsewhere, it's tempting to cache the credentials object as a private member rather than read it every time (to make it more obvious that it won't change from one request to the next).

mikelehen (changes requested):
Thanks! I have a few more suggestions, but generally this looks good to me. It's great to clean up all the token fetching / refreshing code.

If you haven't already, can you do a sanity check that the proactive refresh and database credential change handling code is all working properly?  I think you can just write a little RTDB app that listens to .info/connected, let it run for an hour, and make sure the connection doesn't bounce (if the token expires without being refreshed we'll end up closing / re-opening the connection).

src/main/java/com/google/firebase/FirebaseOptions.java
```diff
@@ -179,9 +196,15 @@ public Builder setStorageBucket(String storageBucket) {
      * @param credential A <code>FirebaseCredential</code> used to authenticate the SDK. See {@link
      *     FirebaseCredentials} for default implementations.
      * @return This <code>Builder</code> instance is returned so subsequent calls can be chained.
+     * @deprecated Use {@link FirebaseOptions.Builder#setCredentials(GoogleCredentials)}.
```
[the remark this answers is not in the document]
hiranya911: We will essentially update our [samples](https://firebase.google.com/docs/admin/setup#initialize_the_sdk) to show how to initialize with different types of `GoogleCredentials` (service accounts, refresh tokens and app default). We can have a link here to the respective page in documentation.

src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -86,28 +84,22 @@
...
   /** Default constructor. */
-  private FirebaseApp(String name, FirebaseOptions options,
-      TokenRefresher.Factory factory, Clock clock) {
+  private FirebaseApp(String name, FirebaseOptions options, TokenRefresher.Factory factory) {
     checkArgument(!Strings.isNullOrEmpty(name));
     this.name = name;
     this.options = checkNotNull(options);
     this.tokenRefresher = checkNotNull(factory).create(this);
-    this.clock = checkNotNull(clock);
+    this.options.getCredentials().addChangeListener(this.tokenRefresher);
```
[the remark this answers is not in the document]
hiranya911: Done

src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -423,32 +353,57 @@ FirebaseService getService(String id) {
...
       cancelPrevious();
       scheduleNext(
-          new Callable<Task<GetTokenResult>>() {
+          new Callable<Void>() {
             @Override
-            public Task<GetTokenResult> call() throws Exception {
-              return firebaseApp.getToken(true);
+            public Void call() throws Exception {
+              logger.debug("Refreshing OAuth2 credential");
+              GoogleCredentials credentials = firebaseApp.getOptions().getCredentials();
```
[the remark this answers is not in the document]
hiranya911: Done

src/main/java/com/google/firebase/database/core/JvmAuthTokenProvider.java
hiranya911: Done

src/main/java/com/google/firebase/database/core/JvmAuthTokenProvider.java
hiranya911: Done

src/main/java/com/google/firebase/auth/FirebaseAuth.java
```diff
@@ -128,29 +130,26 @@ public static synchronized FirebaseAuth getInstance(FirebaseApp app) {
...
       final String uid, final Map<String, Object> developerClaims) {
-    FirebaseCredential credential = ImplFirebaseTrampolines.getCredential(firebaseApp);
-    if (!(credential instanceof FirebaseCredentials.CertCredential)) {
+    GoogleCredentials googleCredentials = ImplFirebaseTrampolines.getCredentials(firebaseApp);
+    if (!(googleCredentials instanceof ServiceAccountCredentials)) {
       return Tasks.forException(
           new FirebaseException(
-              "Must initialize FirebaseApp with a certificate credential to call "
+              "Must initialize FirebaseApp with a service account credential to call "
                   + "createCustomToken()"));
```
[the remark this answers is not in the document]
hiranya911: Fixed. While working on this I also realized that since we are no longer getting tokens via `FirebseApp`, some `FirebaseAuth` methods can continue to work even after deleting the app. Fixed this problem too.

hiranya911:
Made the suggested changes. I did a long running test several days ago, which worked out nicely. I started another just to be sure. Will update this issue with the results.

hiranya911:
[firebase.txt](https://github.com/firebase/firebase-admin-java/files/1307202/firebase.txt)
I just completed a long running test (debug log attached). Results summary:

* Test duration: ~17 hours
* Unexpected connection losses: 0
* Periodic refresh events:  18 (in near perfect 55 minute intervals)

Disconnect events (The initial `false` status is from when the app is starting up):
```
grep "CONNECTION STATUS" -B 1 firebase.txt 
2017-09-14 17:54:57 [FirebaseDatabaseWorker] DEBUG com.google.firebase.database.core.view.EventRaiser - Raising 1 event(s)
CONNECTION STATUS: false
--
2017-09-14 17:54:57 [FirebaseDatabaseEventTarget] DEBUG com.google.firebase.database.core.view.EventRaiser - Raising /.info/connected: VALUE: true
CONNECTION STATUS: true
```

Refresh events:
```
grep "Refreshing OAuth2 credential" firebase.txt 
2017-09-14 18:49:57 [pool-4-thread-1] DEBUG com.google.firebase.FirebaseApp - Refreshing OAuth2 credential
2017-09-14 19:44:57 [pool-4-thread-1] DEBUG com.google.firebase.FirebaseApp - Refreshing OAuth2 credential
2017-09-14 20:39:57 [pool-4-thread-1] DEBUG com.google.firebase.FirebaseApp - Refreshing OAuth2 credential
2017-09-14 21:34:57 [pool-4-thread-1] DEBUG com.google.firebase.FirebaseApp - Refreshing OAuth2 credential
2017-09-14 22:29:57 [pool-4-thread-1] DEBUG com.google.firebase.FirebaseApp - Refreshing OAuth2 credential
2017-09-14 23:24:57 [pool-4-thread-1] DEBUG com.google.firebase.FirebaseApp - Refreshing OAuth2 credential
2017-09-15 00:19:58 [pool-4-thread-1] DEBUG com.google.firebase.FirebaseApp - Refreshing OAuth2 credential
2017-09-15 01:14:58 [pool-4-thread-1] DEBUG com.google.firebase.FirebaseApp - Refreshing OAuth2 credential
2017-09-15 02:09:58 [pool-4-thread-1] DEBUG com.google.firebase.FirebaseApp - Refreshing OAuth2 credential
2017-09-15 03:04:58 [pool-4-thread-1] DEBUG com.google.firebase.FirebaseApp - Refreshing OAuth2 credential
2017-09-15 03:59:58 [pool-4-thread-1] DEBUG com.google.firebase.FirebaseApp - Refreshing OAuth2 credential
2017-09-15 04:54:58 [pool-4-thread-1] DEBUG com.google.firebase.FirebaseApp - Refreshing OAuth2 credential
2017-09-15 05:49:58 [pool-4-thread-1] DEBUG com.google.firebase.FirebaseApp - Refreshing OAuth2 credential
2017-09-15 06:44:59 [pool-4-thread-1] DEBUG com.google.firebase.FirebaseApp - Refreshing OAuth2 credential
2017-09-15 07:39:59 [pool-4-thread-1] DEBUG com.google.firebase.FirebaseApp - Refreshing OAuth2 credential
2017-09-15 08:34:59 [pool-4-thread-1] DEBUG com.google.firebase.FirebaseApp - Refreshing OAuth2 credential
2017-09-15 09:29:59 [pool-4-thread-1] DEBUG com.google.firebase.FirebaseApp - Refreshing OAuth2 credential
2017-09-15 10:24:59 [pool-4-thread-1] DEBUG com.google.firebase.FirebaseApp - Refreshing OAuth2 credential
```

mikelehen (approved):
Changes look good. Thanks!

Issue #73, opened 2017-09-17
FirebaseApp.delete() still has dangling thread


nathan-ca:
### [REQUIRED] Step 2: Describe your environment

  * Operating System version: Windows 10
  * Firebase SDK version: 5.3.1
  * Library version: ?
  * Firebase Product: app

### [REQUIRED] Step 3: Describe the problem

After app.delete(), there is still dangling thread. Bug can be reproduced on both 5.2.0 and latest 5.3.1 releases.

#### Steps to reproduce:

Run a trivial task, such as getUser, and call app.delete() after task complete. In my IDE, I observed three threads were created: pool-2-thread-1, pool-2-thread-2, pool-3-thread-1. After a while, pool-2-thread-[1,2] exited but pool-3-thread-1 is still dangling.

#### Relevant Code:
A trivial java test program
```
app = FirebaseApp.initializeApp(options);
Task<UserRecord> task = auth.getUser(uid);
task.addOnCompleteListener(new OnCompleteListener<UserRecord>() {

  @Override
  public void onComplete(Task<UserRecord> task) {
    app.delete();
  }
});
```

Test program won't exit after task completed.

Looks like TokenRefresher was not exiting. I have to add this hack after app.delete() call:

```
FirebaseExecutors.DEFAULT_SCHEDULED_EXECUTOR.shutdown();
```

hiranya911:
Fix for this issue was released today: https://firebase.google.com/support/release-notes/admin/java#5.4.0

When running on the regular JVM, all threads will be started as daemons, which won't cause the process to stay alive. The `FirebaseApp.delete()` method can also be invoked to further guarantee a clean graceful shutdown (although this is not required). Proactive token refresher will not start until RTDB code is invoked.

PR #74, opened 2017-09-18
Introducing the ThreadManager API

hiranya911:
Adding the new `ThreadManager` interface which controls how an app obtains its thread pools and thread factories. Developers can specify their own implementations of this interface via `FirebaseOptions`. We provide 2 implementations (one for basic JVM and one for GAE), one of which will be used by default.

This PR also ensures:
1. Calling `app.delete()` terminates any and all threads started by the SDK.
2. When running in a regular JVM (non-GAE), we start all threads as daemons. This ensures thread clean up even if the user ignores to call `app.delete()`.
3. Database uses the `ThreadFactory` provided by `ThreadManager` to start all of its threads (event target, run loop, socket reader/writer etc.)
4. The SDK doesn't use `TaskExecutors` for any thing. However, we continue to keep it in the code base for backward compatibility reasons.


hiranya911:
Related to #55 and #73

hiranya911:
I've been running some tests with this PR on App Engine, and I'll document my observations here.

### Manual Scaling

Works as usual. All async tasks (from auth, database and everything else) runs normally with the default `ThreadManager` chosen by the SDK. This is the preferred mode of deployment for accessing RTDB from App Engine environment.

### Automatic Scaling

Works as usual. All async tasks, except for the ones started by database, runs normally with the default `ThreadManager` chosen by the SDK. The database code requires long-lived threads, and hence cannot be used. The SDK will throw exceptions if attempted. But other components like auth can be used normally.

### Basic Scaling

This is similar to manual scaling. So everything, including database, will work as usual. However, the long-lived threads started by the database, and the proactive token refresher will prevent the instance from shutting down due to instance idle time (unless `FirebaseApp.delete()` is called). If this is important one must ensure that:

1. No database calls are made in the app, and
2. Proactive token refresher is disabled.

We will provide an API to achieve 2 in the future. In the meantime, one can use a custom executor (specified via `ThreadManager`). I had success with the following:

```
class MyExecutor extends ScheduledThreadPoolExecutor {
    public MyExecutor(ThreadFactory factory) {
        super(0, factory);
        setKeepAliveTime(60L, TimeUnit.SECONDS);
    }

    public <V> ScheduledFuture<V> schedule(Callable<V> command, long delay, TimeUnit unit) {
        if (delay > 0) {
            throw new UnsupportedOperationException();
        }
        return super.schedule(command, delay, unit);
    }

     public ScheduledFuture<?> schedule(Runnable command, long delay, TimeUnit unit) {
        if (delay > 0) {
            throw new UnsupportedOperationException();
        }
        return super.schedule(command, delay, unit);
    }
}
```

This essentially rejects all "scheduled" tasks, and only admits tasks for immediate execution (i.e. `delay == 0`). This will transparently fail the proactive token refresher, but execute everything else normally. 

I had my basic-scaled instance set to shutdown after 1 minute of idle time:

```
<basic-scaling>
  <max-instances>1</max-instances>
  <idle-timeout>1m</idle-timeout>
</basic-scaling>
```

With the above custom executor, I observed that my instance is shutting down after 2 minutes of idle time (1 extra minute due to the 60s keep alive time set on the executor).  So overall a successful test.

src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -84,9 +91,12 @@
   private final String name;
   private final FirebaseOptions options;
   private final TokenRefresher tokenRefresher;
+  private final ThreadManager threadManager;
 
   private final AtomicBoolean deleted = new AtomicBoolean();
   private final Map<String, FirebaseService> services = new HashMap<>();
+  private final AtomicReference<ListeningScheduledExecutorService> executorReference =
```
schmidt-sebastian: Since this is only ever accessed under lock, does it need to be an AtomicReference?

src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -332,6 +349,36 @@ private void checkNotDeleted() {
     checkState(!deleted.get(), "FirebaseApp was deleted %s", this);
   }
 
+  private ListeningScheduledExecutorService getExecutorService() {
```
schmidt-sebastian: How often is this not called? Can we initialize the executor in the constructor?

src/main/java/com/google/firebase/auth/internal/BaseCredential.java
```diff
@@ -59,14 +60,14 @@ public final GoogleCredentials getGoogleCredentials() {
...
-            accessToken.getExpirationTime().getTime());
-      }
-    });
+    try {
+      AccessToken accessToken = googleCredentials.refreshAccessToken();
+      GoogleOAuthAccessToken googleToken = new GoogleOAuthAccessToken(accessToken.getTokenValue(),
+          accessToken.getExpirationTime().getTime());
+      return Tasks.forResult(googleToken);
+    } catch (IOException e) {
+      throw new RuntimeException(e);
```
schmidt-sebastian: Does the Task library not have its own way of passing along errors?

src/main/java/com/google/firebase/database/tubesock/ThreadConfig.java
```diff
@@ -0,0 +1,42 @@
...
+
+  private final ThreadInitializer threadInitializer;
+  private final ThreadFactory threadFactory;
+
+  public ThreadConfig(ThreadInitializer threadInitializer, ThreadFactory threadFactory) {
+    this.threadInitializer = checkNotNull(threadInitializer);
+    this.threadFactory = checkNotNull(threadFactory);
+  }
+
+  ThreadInitializer getInitializer() {
```
schmidt-sebastian: It is somewhat hard to reason about (from the name alone) what the difference between a ThreadInitializer and a ThreadFactory is. Can you add comments (or rename the new ThreadInitializer)?

src/main/java/com/google/firebase/database/tubesock/WebSocket.java
```diff
@@ -44,6 +46,9 @@
  */
 public class WebSocket {
 
+  private static final ThreadConfig DEFAULT_THREAD_CONFIG = new ThreadConfig(
+      ThreadInitializer.defaultInstance, Executors.defaultThreadFactory());
```
schmidt-sebastian: Should this get the thread initializer from Platform.java?

src/test/java/com/google/firebase/ThreadManagerTest.java
```diff
@@ -0,0 +1,227 @@
...
+    assertEquals(1, threadManager.events.size());
+
+    // Deleting app should tear down threading resources.
+    app.delete();
+    assertEquals(2, threadManager.events.size());
+    event = threadManager.events.get(1);
+    assertEquals(Event.TYPE_RELEASE_EXECUTOR, event.type);
+    assertEquals(app, event.arguments[0]);
+  }
+
```
schmidt-sebastian: Can you add tests that exercise the "setup-teardown-setup-teardown" logic you implemented?

src/test/java/com/google/firebase/ThreadManagerTest.java
```diff
@@ -0,0 +1,227 @@
...
+  
+  private static class Event {
+    private static final int TYPE_GET_EXECUTOR = 100;
+    private static final int TYPE_RELEASE_EXECUTOR = 101;
+    private static final int TYPE_GET_THREAD_FACTORY = 102;
+
+    private final int type;
+    private final Object[] arguments;
+
+    public Event(int type, Object... arguments) {
```
schmidt-sebastian: I would prefer if you explicitly called out the arguments.

src/main/java/com/google/firebase/FirebaseApp.java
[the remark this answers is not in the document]
hiranya911: Done.

src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -316,6 +326,13 @@ public void delete() {
       }
       services.clear();
       tokenRefresher.cleanup();
+
+      // Clean up and terminate the thread pool
+      ScheduledExecutorService executor = executorReference.get();
+      if (executor != null) {
+        threadManager.releaseExecutor(this, executor);
+        executorReference.set(null);
```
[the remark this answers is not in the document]
hiranya911: Now that the AtomicReference is gone, this is a lot simpler.

src/main/java/com/google/firebase/FirebaseApp.java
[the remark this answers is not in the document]
hiranya911: Done.

src/main/java/com/google/firebase/FirebaseOptions.java
```diff
@@ -118,7 +122,12 @@ public JsonFactory getJsonFactory() {
     return jsonFactory;
   }
 
-  /** 
+  @NonNull
```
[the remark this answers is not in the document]
hiranya911: We already have `getCredentials()` which is package-protected. So there is some precedent here.

src/main/java/com/google/firebase/ThreadManager.java
```diff
@@ -0,0 +1,77 @@
...
+  final ListeningScheduledExecutorService getListeningExecutor(@NonNull FirebaseApp app) {
+    ScheduledExecutorService executor = getExecutor(app);
+    checkNotNull(executor, "ScheduledExecutorService must not be null");
+    return MoreExecutors.listeningDecorator(executor);
+  }
+
+  /**
+   * Returns the main thread pool for an app. Implementations may return the same instance of
+   * <code>ScheduledExecutorService</code> for multiple apps. The returned thread pool is used by
+   * all components of an app except for the Realtime Database. Database has far stricter and
```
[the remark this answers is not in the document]
hiranya911: I agree. Semantics of this API are not well defined at the moment. I'm putting together #77 to address this issue.  Once we do those changes we will have a clear separation:
* `getExecutor()` : Returns a thread pool for short-lived tasks
* `getThreadFactory()`: Returns a ThreadFactory for long-lived tasks (includes RTDB and other scheduled tasks)

src/main/java/com/google/firebase/ThreadManager.java
```diff
@@ -0,0 +1,77 @@
...
+   * app is deleted.
+   *
+   * @param app A {@link FirebaseApp} instance.
+   */
+  protected abstract void releaseExecutor(
+      @NonNull FirebaseApp app, @NonNull ScheduledExecutorService executor);
+
+  /**
+   * Returns the <code>ThreadFactory</code> to be used for creating any additional threads
+   * required by the SDK. This is used mainly to create the run loop, event target and web socket
```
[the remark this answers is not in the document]
hiranya911: Will do this as part of #77

src/main/java/com/google/firebase/auth/internal/BaseCredential.java
hiranya911: Done. On a related note that both `Task` and `FirebaseCredential` are now deprecated, and are on their way out.

src/main/java/com/google/firebase/database/tubesock/ThreadConfig.java
hiranya911: Done

src/main/java/com/google/firebase/database/tubesock/WebSocket.java
hiranya911: This constant is not used during normal execution. It is only used in the hidden `Websocket(URL url)` constructor, which is only called by some internal test classes we have implemented (a set of executable `main` methods!). So I believe this is ok. 

We can discuss later whether we need this at all. I think this and the test client code that use it can all be removed.

src/test/java/com/google/firebase/ThreadManagerTest.java
hiranya911: I think that's covered in the `FirebaseExecutorsTest`. Let me know if that's not what you were referring to.

src/test/java/com/google/firebase/ThreadManagerTest.java
hiranya911: Done

hiranya911:
Made most of the suggested changes. A few more changes are on their way in #77. In particular some of the complexities you pointed out in the semantics of `ThreadManager` will get improved for the better.

src/main/java/com/google/firebase/ThreadManager.java
```diff
@@ -66,8 +66,8 @@ protected abstract void releaseExecutor(
 
   /**
    * Returns the <code>ThreadFactory</code> to be used for creating any additional threads
-   * required by the SDK. This is used mainly to create the run loop, event target and web socket
-   * reader/writer threads of the Realtime Database client.
+   * required by the SDK. This is used mainly to create the long-lived worker threads for
```
schmidt-sebastian: missing "the"

src/main/java/com/google/firebase/database/tubesock/ThreadConfig.java
```diff
@@ -32,10 +37,25 @@ public ThreadConfig(ThreadInitializer threadInitializer, ThreadFactory threadFac
     this.threadFactory = checkNotNull(threadFactory);
   }
 
+  /**
+   * The {@link ThreadInitializer} implementation to be used for further configuring already
+   * created threads. In particular, this instance is used to rename the created threads, and turn
```
schmidt-sebastian: This suggests that it would work with threads that were created way back in the days. This class is used to configure newly created threads, while the factory is used to create new threads.

src/main/java/com/google/firebase/ImplFirebaseTrampolines.java
```diff
@@ -61,4 +68,12 @@ public static String getPersistenceKey(String name, FirebaseOptions options) {
     app.addService(service);
     return service;
   }
+
+  public static ThreadFactory getThreadFactory(@NonNull FirebaseApp app) {
+    return app.getThreadFactory();
+  }
+
+  public static <T> Task<T> submit(@NonNull FirebaseApp app, @NonNull Callable<T> command) {
```
schmidt-sebastian: submitCallable maybe? 'submit' seems generic in the context of this class.

src/main/java/com/google/firebase/ThreadManager.java
```diff
@@ -0,0 +1,77 @@
...
+   * <code>ScheduledExecutorService</code> for multiple apps. The returned thread pool is used by
+   * all components of an app except for the Realtime Database. Database has far stricter and
+   * complicated threading requirements, and thus initializes its own threads using the
+   * factory returned by {@link ThreadManager#getThreadFactory()}.
+   *
+   * @param app A {@link FirebaseApp} instance.
+   * @return A non-null {@link ScheduledExecutorService} instance.
+   */
+  @NonNull
+  protected abstract ScheduledExecutorService getExecutor(@NonNull FirebaseApp app);
```
schmidt-sebastian: Could we combine the two methods (this and above) and always return a ListeningScheduledExecutorService?

src/main/java/com/google/firebase/auth/internal/BaseCredential.java
```diff
@@ -59,14 +60,14 @@ public final GoogleCredentials getGoogleCredentials() {
...
-        return new GoogleOAuthAccessToken(accessToken.getTokenValue(),
-            accessToken.getExpirationTime().getTime());
-      }
-    });
+    try {
+      AccessToken accessToken = googleCredentials.refreshAccessToken();
+      GoogleOAuthAccessToken googleToken = new GoogleOAuthAccessToken(accessToken.getTokenValue(),
+          accessToken.getExpirationTime().getTime());
+      return Tasks.forResult(googleToken);
+    } catch (IOException e) {
```
schmidt-sebastian: Just Exception?

src/main/java/com/google/firebase/database/tubesock/ThreadConfig.java
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Configuration that governs how the websocket connections spawn and initialize threads
+ * for reading and writing data.
+ */
+public final class ThreadConfig {
+
+  private final ThreadInitializer threadInitializer;
+  private final ThreadFactory threadFactory;
+
+  public ThreadConfig(ThreadInitializer threadInitializer, ThreadFactory threadFactory) {
```
schmidt-sebastian: Nit: Swap the order here. First thread get 'factoried' and then initialized.

src/main/java/com/google/firebase/internal/FirebaseExecutors.java
```diff
@@ -16,20 +16,122 @@
...
+   * across all active apps. The executor service is initialized when the first app is initialized,
+   * and terminated when the last app is deleted. This class is thread safe.
+   */
+  abstract static class GlobalThreadManager extends ThreadManager {
+
+    private final Set<String> apps = new HashSet<>();
+    private ScheduledExecutorService executorService;
+
+    @Override
+    protected synchronized ScheduledExecutorService getExecutor(FirebaseApp app) {
```
schmidt-sebastian: I would suggest not to synchronize on the public facing object (since this can also be done by consumers). Can you use a private member (like executorService)?

src/main/java/com/google/firebase/internal/FirebaseExecutors.java
```diff
@@ -16,20 +16,122 @@
...
+    @Override
+    protected ScheduledExecutorService doInit() {
+      int cores = Runtime.getRuntime().availableProcessors();
+      // Create threads as daemons to ensure JVM exit when all foreground jobs are complete.
+      ThreadFactory threadFactory = new ThreadFactoryBuilder()
+          .setNameFormat("firebase-default-%d")
+          .setDaemon(true)
+          .setThreadFactory(getThreadFactory())
+          .build();
+      logger.debug("Initializing default executor with {} max threads", cores);
```
schmidt-sebastian: I don't believe there is any guarantee that ScheduledThreadPoolExecutor will try to schedule each thread on a dedicated core, so using this number here may not make much sense.

src/main/java/com/google/firebase/ThreadManager.java
hiranya911: Done

src/main/java/com/google/firebase/ThreadManager.java
hiranya911: I would prefer to keep `ThreadManager` API decoupled from any Guava types. That way developers don't have to learn about Guava APIs to extend this API.

src/main/java/com/google/firebase/ImplFirebaseTrampolines.java
hiranya911: Done

src/main/java/com/google/firebase/database/tubesock/ThreadConfig.java
hiranya911: Updated the text.

src/main/java/com/google/firebase/auth/internal/BaseCredential.java
hiranya911: Done

src/main/java/com/google/firebase/database/tubesock/ThreadConfig.java
hiranya911: Done

src/main/java/com/google/firebase/internal/FirebaseExecutors.java
hiranya911: Done

src/main/java/com/google/firebase/internal/FirebaseExecutors.java
hiranya911: Correct. I just used that as a max limit since `ScheduledThreadPoolExecutor` must be given some core thread pool count. I think we can go back to cached thread pool when we implement #77.

hiranya911:
Made the suggested changes. Over to @schmidt-sebastian for another round.

PR #77, opened 2017-09-25
Improving the ThreadManager and TokenRefresher APIs

hiranya911:
This is a follow up to #74. It attempts to solve several issues in the current design:

* `ThreadManager` interface returns a `ScheduledExecutorService`. This makes implementing the `ThreadManager` API cumbersome, since the default implementation provided in Java (`ScheduledThreadPoolExecutor`) is not very amenable to configuration.
* The different uses of the `getExecutor()` and `getThreadManager()` methods are not well defined in the `ThreadManager` interface. At the moment it is something along the lines of:
  - `getThreadManager()`: for RTDB
  - `getExecutor()`: for everything else
* Proactive token refresher always get started when calling `initializeApp()` (which starts long-lived threads). But only RTDB requires it in practice. This is not ideal when using the SDK only for auth operations.

To resolve these issue, this PR makes the following changes:

* Clearly define the contract of the `ThreadManager` interface.
  - `getThreadMaanger()`: Used to start long-lived threads (including RTDB, token refresher and anything else that would require long-lived threads in the future)
  - `getExecutor()`: Used for short-lived tasks.
* Return an `ExecutorService` from `ThreadManager`, whose default implementation (`ThreadPoolExecutor`) is much easier to configure.
* Invert the control of the proactive token refresher. Instead of always starting it from the app, let RTDB (or any other interested component) start it on demand.

The downside of this change is that we end up using a dedicated executor for proactive token refresher (we can use it for other scheduled tasks that we may add in the future -- but for now it's only used for token refresher). In GAE this will end up occupying a separate background thread. This shouldn't be a problem for those who are only using RTDB, but GAE users who access both RTDB and auth ends up spending another thread.  However, it is possible to provide a custom `ThreadManager` that uses request-scoped threads for short-lived tasks, which is one way to avoid this drawback.


hiranya911:
I did some testing on GAE for this PR. Auto and manual scaling works as before. Basic scaling also works, and idle instance shutdown can be easily enabled with a custom `ThreadManager`:

```
static class BasicScalingThreadManager extends ThreadManager {

    @Override
    protected ExecutorService getExecutor(@NonNull FirebaseApp firebaseApp) {
      // Use a single-threaded executor which keeps threads alive for 1 minute.
      return new ThreadPoolExecutor(0, 1,
          60L, TimeUnit.SECONDS, new SynchronousQueue<Runnable>(), getThreadFactory());
    }

    @Override
    protected void releaseExecutor(@NonNull FirebaseApp firebaseApp,
        @NonNull ExecutorService executorService) {
      executorService.shutdownNow();
    }

    @Override
    protected ThreadFactory getThreadFactory() {
      return com.google.appengine.api.ThreadManager.backgroundThreadFactory();
    }
}
```

My test GAE instance also has a 1 minute idle timeout. With this setup, when the app doesn't receive requests for 2 minutes (1 minute for the threads to die, and 1 more minute for the GAE idle timeout to kick in), the instance is stopped.

```
15:36:57.769  GET 404 270 B 7.2 s Unknown /_ah/start GET 404 270 B 7.2 s Unknown 
15:37:06.974  GET 200 2 B 71.6 s Unknown /_ah/background GET 200 2 B 71.6 s Unknown
15:37:16.326  GET 200 119 B 231 ms curl/7.53.1 /users?uid=jALVj6oEyvXmwLZuMcUCOeRWxhS2&count=1 GET 200 119 B 231 ms curl/7.53.1 
15:37:17.414  GET 200 119 B 230 ms curl/7.53.1 /users?uid=jALVj6oEyvXmwLZuMcUCOeRWxhS2&count=1 GET 200 119 B 230 ms curl/7.53.1
15:37:18.383  GET 200 119 B 213 ms curl/7.53.1 /users?uid=jALVj6oEyvXmwLZuMcUCOeRWxhS2&count=1 GET 200 119 B 213 ms curl/7.53.1
15:39:19.068  GET 200 2 B 93 ms Unknown /_ah/stop GET 200 2 B 93 ms Unknown
15:39:57.372  GET 200 119 B 13.3 s curl/7.53.1 /users?uid=jALVj6oEyvXmwLZuMcUCOeRWxhS2&count=1 GET 200 119 B 13.3 s curl/7.53.1
15:39:57.382  GET 404 270 B 8.3 s Unknown /_ah/start GET 404 270 B 8.3 s Unknown
15:40:06.735  GET 200 2 B 64 s Unknown /_ah/background GET 200 2 B 64 s Unknown
15:42:11.231  GET 200 2 B 93 ms Unknown /_ah/stop GET 200 2 B 93 ms Unknown 
```

hiranya911:
I've been conducting a long running test with this PR on App Engine (manual scaling). The test is coming up on 24 hours now and seem to be working fine.

src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -90,10 +93,12 @@
   private final FirebaseOptions options;
   private final TokenRefresher tokenRefresher;
   private final ThreadManager threadManager;
+  private final ThreadManager.FirebaseExecutor executor;
 
   private final AtomicBoolean deleted = new AtomicBoolean();
   private final Map<String, FirebaseService> services = new HashMap<>();
-  private final ListeningScheduledExecutorService executor;
+
+  private volatile ScheduledExecutorService scheduledExecutor;
```
schmidt-sebastian: It seems like 'ScheduledExecutorService' doesn't spawn its threads until it gets used for the first time. If you were to initialize it here, the code is this class could be drastically simplified.

src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -388,26 +423,30 @@ FirebaseService getService(String id) {
    */
   static class TokenRefresher implements CredentialsChangedListener {
 
+    private static final int STATE_READY = 0;
```
schmidt-sebastian: Can we use an Enum here and an AtomicReference below?

It may even be fine to just use a volatile enum: http://www.javamex.com/tutorials/synchronization_volatile.shtml

src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -446,14 +465,70 @@ protected void scheduleNext(Callable<Void> task, long delayMillis) {
...
+      credentials.addChangeListener(this);
+      AccessToken accessToken = credentials.getAccessToken();
+      long refreshDelay = 0L;
+      if (accessToken != null) {
+        refreshDelay = Math.max(getRefreshDelay(accessToken), refreshDelay);
       }
+      // If the access token is null, or is about to expire (i.e. expires in less than 5 minutes),
+      // schedule a refresh event with 0 delay. Otherwise schedule a refresh event at the token
+      // expiry time, minus 5 minutes.
+      scheduleRefresh(refreshDelay);
```
schmidt-sebastian: Subtract the 5 minutes here

src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -388,26 +423,30 @@ FirebaseService getService(String id) {
...
-      if (closed) {
+      if (state.get() != STATE_STARTED) {
         return;
       }
 
       AccessToken accessToken = credentials.getAccessToken();
-      long refreshDelay = accessToken.getExpirationTime().getTime()
-          - System.currentTimeMillis() - TimeUnit.MINUTES.toMillis(5);
+      long refreshDelay = getRefreshDelay(accessToken);
       if (refreshDelay > 0) {
```
schmidt-sebastian: This doesn't work for intervals <= 5 minutes.

src/main/java/com/google/firebase/ThreadManager.java
```diff
@@ -36,42 +36,71 @@
...
+   * creates dedicated executors using the <code>ThreadFactory</code> returned by the
+   * {@link #getThreadFactory()} method.
    *
    * @param app A {@link FirebaseApp} instance.
-   * @return A non-null {@link ScheduledExecutorService} instance.
+   * @return A non-null {@link ExecutorService} instance.
    */
   @NonNull
-  protected abstract ScheduledExecutorService getExecutor(@NonNull FirebaseApp app);
+  protected abstract ExecutorService getExecutor(@NonNull FirebaseApp app);
```
schmidt-sebastian: It's unclear to me when what the difference between and Executor and a FirebaseExecutor is.

src/main/java/com/google/firebase/ThreadManager.java
```diff
@@ -36,42 +36,71 @@
...
    */
   @NonNull
   protected abstract ThreadFactory getThreadFactory();
 
+  /**
+   * Wraps an ExecutorService in a ListeningExecutorService while keeping a reference to the
+   * original ExecutorService. This reference is used when it's time to release/cleanup the
+   * original ExecutorService.
+   */
+  static final class FirebaseExecutor {
```
schmidt-sebastian: This feels like it could be replaced by one line of code in the base class :)

src/main/java/com/google/firebase/internal/FirebaseExecutors.java
```diff
@@ -55,11 +55,10 @@
 
     private final Object lock = new Object();
     private final Set<String> apps = new HashSet<>();
-
-    private ScheduledExecutorService executorService;
+    private ExecutorService executorService;
 
     @Override
-    protected ScheduledExecutorService getExecutor(FirebaseApp app) {
+    protected synchronized ExecutorService getExecutor(FirebaseApp app) {
```
schmidt-sebastian: Don't lock on the main object (and don't lock twice).

src/main/java/com/google/firebase/internal/FirebaseExecutors.java
```diff
@@ -70,7 +69,7 @@ protected ScheduledExecutorService getExecutor(FirebaseApp app) {
     }
 
     @Override
-    protected void releaseExecutor(FirebaseApp app, ScheduledExecutorService executor) {
+    protected synchronized void releaseExecutor(FirebaseApp app, ExecutorService executor) {
       synchronized (lock) {
```
schmidt-sebastian: As above

src/main/java/com/google/firebase/internal/FirebaseExecutors.java
```diff
@@ -117,25 +116,50 @@ protected ThreadFactory getThreadFactory() {
...
+   * and RTDB can be supported as if running on the regular JVM.
+   *
+   * <p>Basic-scaling: Behavior is similar to manual-scaling. Since the threads are kept alive
+   * indefinitely, prevents the GAE idle instance shutdown. Developers are advised to use
+   * a custom ThreadManager implementation if idle instance shutdown should be supported. In
+   * general, a ThreadManager implementation that uses the request-scoped ThreadFactory, or the
+   * background ThreadFactory with specific keep-alive times can easily facilitate GAE idle
+   * instance shutdown. Note that this often comes at the cost of losing scheduled tasks and RTDB
+   * support. Therefore, for these features, manual-scaling is the recommended GAE deployment mode
+   * regardless of the ThreadManager implementation used.
```
schmidt-sebastian: Nice!

src/main/java/com/google/firebase/FirebaseApp.java
hiranya911: Done. Used enum. Kept the atomic reference for compare-and-set method.

src/main/java/com/google/firebase/FirebaseApp.java
hiranya911: This happens in the call to `getRefreshDelay()` a couple of lines earlier. I reorg'ed the code and comments a bit to make this clear.

src/main/java/com/google/firebase/FirebaseApp.java
hiranya911: That is correct. However, in practice this doesn't really happen. New tokens minted by Google auth have a 1 hour TTL. If it does happen (due to some issue in the remote token server), we don't want to schedule refresh events aggressively, and cause a feedback loop. So we simply log a warning and let things run course.  This is what we have done in the past releases too.

src/main/java/com/google/firebase/ThreadManager.java
hiranya911: The problem here is once we wrap the `ExecutorService` in a `ListeningExecutorService`, there's no way to reference the original executor again unless we explicitly store a reference to the original object. There's no way to get the delegate from the `ListeningExecutorService`.

On the other hand we want to make sure that the argument passed to user's `releaseExecutor()` is the same object returned by user's `getExecutor()`. So it looks like this is the easiest solution available. Or have I overlooked something obvious?

src/main/java/com/google/firebase/ThreadManager.java
hiranya911: `FirebaseExecutor` is just a holder for the `ListeningExecutorService` **and** the underlying `ExecutorService` returned by the user code. By keeping a reference to the original `ExecutorService` around we can make sure that the argument passed to `releaseExecutor()` is same as the one returned by `getExecutor()`.

src/main/java/com/google/firebase/internal/FirebaseExecutors.java
hiranya911: Ah! Missed it during the merge with #74. Thanks for pointing that out.

src/main/java/com/google/firebase/internal/FirebaseExecutors.java
hiranya911: Done

src/main/java/com/google/firebase/FirebaseApp.java
hiranya911: This will initialize the `ThreadFactory` though (`ThreadManager.getThreadFactory()`) will get called. On App Engine this attempts to start a no-op background thread to see if the background thread support is available.Probably not a big deal, but just wanted to get your opinion on it before I make the change.

schmidt-sebastian: I went through great lengths to make sure that we don't kick off random threads in the first place, so we should probably keep this as is. Thanks!

src/main/java/com/google/firebase/FirebaseApp.java
schmidt-sebastian: If you substract the 5 minutes here, then getRefreshDelay() will return exactly the delay that is specified in the token and then "refreshDelay > 0" check will work for all delays.

src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -446,20 +461,85 @@ protected void scheduleNext(Callable<Void> task, long delayMillis) {
...
+     * events, and schedules refresh events every time the OAuth2 token changes. If no active
+     * token is present, or if the available token is set to expire soon, this will also schedule
+     * a refresh event to be executed immediately.
+     *
+     * <p>This operation is idempotent. Calling it multiple times, or calling it after the
+     * refresher has been stopped has no effect.
+     */
+    final synchronized void start() {
+      // Allow starting only from the ready state.
+      if (!state.compareAndSet(State.READY, State.STARTED)) {
```
schmidt-sebastian: Nit: This might not be needed since state is only ever changed under lock.

src/main/java/com/google/firebase/ThreadManager.java
schmidt-sebastian: Can this class be renamed to "FirebaseExecutors", with two members: "userExecutor" & "listeningExecutor"?

src/main/java/com/google/firebase/ThreadManager.java
schmidt-sebastian: I don't know if we want to go down that road - but what if we required the user to provide a ListeningExecutorService in the first place? Would that be cleaner?

src/test/java/com/google/firebase/FirebaseAppTest.java
```diff
@@ -345,6 +350,99 @@ public void testProactiveTokenRefresh() throws Exception {
...
+      protected void cancelPrevious() {
+        cancelCalls.incrementAndGet();
+      }
+
+      @Override
+      protected void scheduleNext(Callable<Void> task, long delayMillis) {
+        scheduleCalls.incrementAndGet();
+      }
+    };
+    // stop() is allowed here, but since we didn't start(), no measurable state change
```
schmidt-sebastian: Nit: I know there are good reasons to keep this as a separate test case, but I would personally combine it with the previous test.

src/main/java/com/google/firebase/FirebaseApp.java
hiranya911: I'm also using `compareAndSet()` to ensure that we cannot start the refresher once it has been stopped.

src/main/java/com/google/firebase/ThreadManager.java
hiranya911: It would make things simpler for us. But we would be putting the onus of managing these references on the user. I would also prefer not to tie our API to a Guava type.

src/main/java/com/google/firebase/ThreadManager.java
hiranya911: Done. I renamed the existing `FirebaseExecutors` to `FirebaseThreadManagers`, and renamed this to `FirebaseExecutors`.

src/test/java/com/google/firebase/FirebaseAppTest.java
hiranya911: Done

hiranya911:
Made some of the suggested changes. Responded to all open comments. On to the next round.

PR #81, opened 2017-10-04
Serialize Collections

### LINKED ISSUE #78: FR: Add Support for Collections
Add support to the Firebase Admin SDK for any [Collection](https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html) - they all provide an [`iterator()`](https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html#iterator--) so I don't see why it is a problem to do this.

### [REQUIRED] Step 2: Describe your environment

  * Operating System version: Microsoft Windows 10 Pro 10.0.15063
  * Firebase SDK version: 5.3.1
  * Firebase Product: realtime database

### [REQUIRED] Step 3: Describe the problem

I am trying to serialize a [Set](https://docs.oracle.com/javase/8/docs/api/java/util/Set.html) using Firebase and it is throwing `Serializing Collections is not supported, please use Lists instead` due to [this throw statement](https://github.com/firebase/firebase-admin-java/blob/master/src/main/java/com/google/firebase/database/utilities/encoding/CustomClassMapper.java#L155)

#### Steps to reproduce:

Attempt to serialize a Set with the Firebase Admin SDK

#### Relevant Code:

Here is a JUnit 4.0 test that produces the error:
```
import com.google.firebase.database.DatabaseException;
import com.google.firebase.database.utilities.encoding.CustomClassMapper;
import org.junit.Test;

import java.util.Collections;
import java.util.HashMap;
import java.util.Map;

import static org.junit.Assert.assertEquals;

public class FirebaseTest {

    @Test
    public void firebaseCannotSerializeSets() throws Exception {
        Map<String, Object> data = new HashMap<>();
        data.put("myset", Collections.singleton("1"));

        try {
            CustomClassMapper.convertToPlainJavaTypes(data);
        } catch (DatabaseException e) {
            assertEquals("Serializing Collections is not supported, " +
                    "please use Lists instead", e.getMessage());
        }
    }
}

```

Rossiar:
Resolves #78 by removing the exception when a collection is not a `List` - since all `Collection` objects are `Iterable` they can be iterated over in a `for` loop and the serialized results added to the `result`, which I guess normalises to a `List` so that the JSON serializer can handle the serialization of the collection of objects? (since every object inside has already been serialized)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -61,42 +61,172 @@
         <skipUTs>${skipTests}</skipUTs>
     </properties>
 
-    <distributionManagement>
-        <repository>
-            <id>cloudbees-private-release-repository</id>
-            <url>https://repository-faceit.forge.cloudbees.com/release</url>
-        </repository>
-        <snapshotRepository>
-            <id>cloudbees-private-snapshot-repository</id>
-            <url>https://repository-faceit.forge.cloudbees.com/snapshot</url>
-        </snapshotRepository>
-    </distributionManagement>
     <scm>
-        <url>scm:git:https://github.com/faceit/${project.artifactId}.git</url>
+        <connection>scm:git:https://github.com/firebase/firebase-admin-java.git</connection>
+        <developerConnection>scm:git:ssh://git@github.com/firebase/firebase-admin-java.git</developerConnection>
+        <url>http://github.com/firebase/firebase-admin-java</url>
+        <tag>HEAD</tag>
     </scm>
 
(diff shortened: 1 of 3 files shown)


mikelehen:
Per https://github.com/firebase/firebase-admin-java/issues/78#issuecomment-334279965 the current behavior is intentional and I'm closing this PR.  Thanks for putting it together though!  We appreciate the engagement. :-)

PR #84, opened 2017-10-07
Load integration test credentials from well-known files

hiranya911:
Up until now we have required passing integration test credentials explicitly as system properties:

```
-Difrebase.it.certificate=path/to/key.json -Dfirebase.it.apikey=my_api_key
```

However, it is much easier if the test suite can simply load these parameters implicitly, from well-known files in the local file system. That way the development environment can be set up once with the required files, and the test suite can be invoked without additional parameters from there on.

This PR modifies the test suite to load the integration test credentials from `integration_cert.json` and `integration_apikey.txt` files. We've been following a similar convention in other admin SDKs, which has proven to be a lot easier to manage.

### DIFF
--- a/src/test/java/com/google/firebase/testing/IntegrationTestUtils.java
+++ b/src/test/java/com/google/firebase/testing/IntegrationTestUtils.java
@@ -16,10 +16,10 @@
 
 package com.google.firebase.testing;
 
-import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
-import com.google.common.base.Strings;
+import com.google.api.client.googleapis.util.Utils;
+import com.google.api.client.json.GenericJson;
 import com.google.common.collect.ImmutableList;
 import com.google.common.io.CharStreams;
 import com.google.firebase.FirebaseApp;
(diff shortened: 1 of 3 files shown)


src/test/java/com/google/firebase/testing/IntegrationTestUtils.java
```diff
@@ -40,26 +40,25 @@
...
-          "Service account certificate path not set. Set the "
-              + "firebase.it.certificate system property and try again.");
-      try (InputStreamReader reader = new InputStreamReader(new FileInputStream(certificatePath))) {
-        IT_SERVICE_ACCOUNT = new JSONObject(CharStreams.toString(reader));
+  private static synchronized GenericJson ensureServiceAccount() {
+    if (serviceAccount == null) {
+      try (InputStream stream = new FileInputStream(IT_SERVICE_ACCOUNT_PATH)) {
+        serviceAccount = Utils.getDefaultJsonFactory().fromInputStream(stream, GenericJson.class);
       } catch (IOException e) {
         throw new RuntimeException("Failed to read service account certificate", e);
```
mikelehen: Should we make this error more helpful?  Something like "Failed to read service account certificate from &lt;path&gt; see CONTRIBUTING.md for more details."

src/test/java/com/google/firebase/testing/IntegrationTestUtils.java
```diff
@@ -78,10 +77,14 @@ public static String getStorageBucket() {
...
-  public static String getApiKey() {
-    String apiKey = System.getProperty("firebase.it.apikey");
-    checkArgument(!Strings.isNullOrEmpty(apiKey), "API key not specified. Set the "
-        + "firebase.it.apikey system property and try again.");
+  public static synchronized String getApiKey() {
+    if (apiKey == null) {
+      try (InputStream stream = new FileInputStream(IT_API_KEY_PATH)) {
+        apiKey = CharStreams.toString(new InputStreamReader(stream)).trim();
+      } catch (IOException e) {
+        throw new RuntimeException("Failed to read api key file", e);
```
mikelehen: Ditto.

src/test/java/com/google/firebase/testing/IntegrationTestUtils.java
hiranya911: Done

src/test/java/com/google/firebase/testing/IntegrationTestUtils.java
hiranya911: Done

mikelehen:
Thanks, looks good!

Issue #75, opened 2017-09-20
Deserializing BigDecimal throws exception


CluEleSsUK:
Inside `CustomClassMapper` in method `deserializeToClass(Object obj, Class<T> clazz)`, the conditional statement wrongly assumes the value is of primitive type if `Number.class.isAssignableFrom(clazz)`.
BigDecimal returns true but is not primitive, and throws an `IllegalArgumentException` in `deserializeToPrimitive(Object obj, Class<T> clazz)`

hiranya911:
Thanks for pointing this out @CluEleSsUK. If you can please provide us a pull request to fix this. Otherwise I will take a look in a couple of days.

CluEleSsUK:
Sure thing, I'll have a look tomorrow!

ghajba:
A question to this topic: how should a BigDecimal/BigInteger be serialized?

I am currently facing the same issue, and I want to fix it, but I don't just want to hack something together.

schmidt-sebastian:
We are taking a look at this internally as this affects Android as well.

schmidt-sebastian:
I have pushed PRs to provide better error messages. Unfortunately, this will only marginally improve your experience with BigDecimal/BigInteger. Our backend does not support integer precision for numbers that exceed the 64bit signed range. If you do need to store larger numbers, please use "double" instead.

hiranya911:
@schmidt-sebastian are you going to implement any changes in the Admin SDK code?

schmidt-sebastian:
We are only updating our error messages: https://github.com/googleapis/google-cloud-java/pull/4675

hiranya911:
As of now we don't plan to support serialization of these types in RTDB. But we can revisit this issue if there's more interest.

PR #85, opened 2017-10-08
BigDecimal and BigInteger can be serialized

ghajba:
This is my solution to #75.

Because I didn't get any response and I need this change, I've solved it like this. But as mentioned in the related issue, I am open to suggestions.

And a question: when will be changes released that we can use the latest version instead of a snapshot?


ghajba:
I signed it!

actually with my primary email address: hajbag@gmail.com

2017-10-08 9:00 GMT+02:00 googlebot <notifications@github.com>:

src/main/java/com/google/firebase/database/snapshot/NodeUtilities.java
```diff
@@ -58,6 +60,10 @@ public static Node NodeFromJSON(Object value, Node priority) throws DatabaseExce
         return new DoubleNode((Double) value, priority);
       } else if (value instanceof Boolean) {
         return new BooleanNode((Boolean) value, priority);
+      } else if (value instanceof BigDecimal) {
+        return new DoubleNode(((BigDecimal) value).doubleValue(), priority);
```
hiranya911: This is a lossy conversion. What happens when the BigDecimal is too large to fit in a double?

src/main/java/com/google/firebase/database/snapshot/NodeUtilities.java
```diff
@@ -58,6 +60,10 @@ public static Node NodeFromJSON(Object value, Node priority) throws DatabaseExce
...
+      } else if (value instanceof BigInteger) {
+        return new LongNode(((BigInteger)value).longValue(), priority);
```
hiranya911: Same question applies here. What happens when the BigInteger is too large to fit in a long?

ghajba:
I know that these are lossy conversions. If the values don't fit, they'll be converted to infinity (positive or negative) for both types.

But I've already asked in #75 how the conversion should happen. After we clear this, I will adjust the code and write the tests too.

hiranya911:
I think you will have to implement a whole new Node type to support these types. In there you can just hold on to the string representation of the numbers.

ghajba:
Just to let you know: I am working on the changes but at a slow pace. I hope I can push my changes this week.

src/main/java/com/google/firebase/database/snapshot/BigDecimalNode.java
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  @Override
+  public String getHashRepresentation(HashVersion version) {
+    return null;
+  }
+
+  @Override
+  protected LeafType getLeafType() {
+    return LeafType.String;
```
hiranya911: I believe this should be `LeafType.Number`

src/main/java/com/google/firebase/database/snapshot/BigDecimalNode.java
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  @Override
+  public Object getValue() {
+    return this.value;
+  }
+
+  @Override
+  public String getHashRepresentation(HashVersion version) {
+    return null;
```
hiranya911: I think this should do something similar to what the `StringNode` does.

src/main/java/com/google/firebase/database/core/SyncTree.java
```diff
@@ -742,7 +742,8 @@ public Node calcCompleteEventCache(Path path, List<Long> writeIdsToExclude) {
       if (currentSyncPoint != null) {
         serverCache = currentSyncPoint.getCompleteServerCache(relativePath);
       }
-    } while (!pathToFollow.isEmpty() && serverCache == null);
+    }
```
hiranya911: Please revert to keep the change footprint small as possible.

src/main/java/com/google/firebase/database/snapshot/BigDecimalNode.java
```diff
@@ -0,0 +1,60 @@
+package com.google.firebase.database.snapshot;
```
hiranya911: Add license header.

src/main/java/com/google/firebase/database/snapshot/BigIntegerNode.java
```diff
@@ -0,0 +1,60 @@
+package com.google.firebase.database.snapshot;
```
hiranya911: Lot of the comments made on `BigDecimalNode` are also applicable here. Please revise as appropriate.

src/main/java/com/google/firebase/database/snapshot/BigDecimalNode.java
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  @Override
+  public Node updatePriority(Node priority) {
+    return new BigDecimalNode(this.value, priority);
+  }
+
+  @Override
+  public Object getValue() {
+    return this.value;
```
hiranya911: Should this return a `BigDecimal`? You can reconstruct the original value from the string.

src/main/java/com/google/firebase/database/snapshot/BigIntegerNode.java
ghajba: I'd add the tests, but I don't see the idea behind how you test your code. The structure is not clear to find an easy entry-point.

src/main/java/com/google/firebase/database/snapshot/BigDecimalNode.java
hiranya911: And as a heads up, you will have to revise how leaf nodes are compared in the parent `LeafNode.compareTo()` method. You will see that longs and doubles have been special cased there. Perhaps you can implement a new comparison method for all number types, which converts values to `BigDecimal` and then does the comparison.

src/main/java/com/google/firebase/database/snapshot/BigIntegerNode.java
hiranya911: `NodeTest` might be a good place to start. Also check `MapperTest`. See if you can serialize/deserialize some made up big integers and big decimals via the new code.

ghajba:
Thanks for the feedback. A big part of the codebase and what is expected is really unclear because of the missing JavaDoc. 
Anyway, I'll try my best to come up with a feasible solution for this issue and get it integrated :)

src/test/java/com/google/firebase/database/MapperTest.java
```diff
@@ -248,6 +252,31 @@ public void primitiveDeserializeLong() {
     } catch (DatabaseException e) { // ignore
     }
   }
+  
+  @Test
+  public void bigIntegerDeserialize() {
+    BigIntegerBean beanBigInteger = deserialize("{'value':'1'}", BigIntegerBean.class);
+    assertEquals(0,BigInteger.ONE.compareTo(beanBigInteger.getValue()));
+    
+    BigIntegerBean beanBigIntegerBig =
```
hiranya911: Can you use a value that is guaranteed to be larger then what can fit in a `long`? Something like:

```
BigInteger value = new BigInteger(String.valueOf(Long.MAX_VALUE)).add(BigInteger.ONE);
```

src/test/java/com/google/firebase/database/MapperTest.java
```diff
@@ -248,6 +252,31 @@ public void primitiveDeserializeLong() {
     } catch (DatabaseException e) { // ignore
     }
   }
+  
+  @Test
+  public void bigIntegerDeserialize() {
+    BigIntegerBean beanBigInteger = deserialize("{'value':'1'}", BigIntegerBean.class);
+    assertEquals(0,BigInteger.ONE.compareTo(beanBigInteger.getValue()));
```
hiranya911: Can we just do:

```
assertEquals(BigInteger.ONE, beanBigInteger.getValue());
```

src/test/java/com/google/firebase/database/MapperTest.java
```diff
@@ -248,6 +252,31 @@ public void primitiveDeserializeLong() {
...
+            .compareTo(beanBigIntegerBig.getValue()));
+  }
+  
+  @Test
+  public void bigDecimalDeserialize() {
+    BigDecimalBean beanBigDecimal = deserialize("{'value':'1.0'}", 
+            BigDecimalBean.class);
+    assertEquals(0,BigDecimal.ONE.compareTo(beanBigDecimal.getValue()));
+    
+    BigDecimalBean beanBigDecimalBig =
```
hiranya911: Same comments made on the test case above applies here.

hiranya911:
Instructions for integration tests can be found at: https://github.com/firebase/firebase-admin-java/blob/master/CONTRIBUTING.md#integration-testing

ghajba:
Want to give a note that I have not abandoned my PR, but had some busy days. I am back now and will add the required changes/tests and will take a look at the comparison method too @hiranya911 mentioned previously.

ghajba:
Well, the integration testing description seems not working. I updated my sources to the latest version of firebase and the FirebaseCredentialsTest, using a new authentication which is not described in the mentioned documentation, fail.
This makes the contribution experience not the best.

hiranya911:
The instructions look correct to me. I use them fairly regularly in multiple dev environments, and haven't had any issues. What is the code/test you're trying to run, and what is the error you get?

FWIW, `FirebaseCredentialsTest` is not an integration test. It's only a unit test.

ghajba:
These are the errors I get when running `mvn verify`:

```
Running com.google.firebase.auth.FirebaseAuthTest
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.298 sec <<< FAILURE! - in com.google.firebase.auth.FirebaseAuthTest
initializationError(com.google.firebase.auth.FirebaseAuthTest)  Time elapsed: 0.006 sec  <<< ERROR!
java.io.IOException: The Application Default Credentials are not available. They are available if running in Google Compute Engine. Otherwise, the environment variable GOOGLE_APPLICATION_CREDENTIALS must be defined pointing to a file defining the credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
        at com.google.firebase.auth.FirebaseAuthTest.createApplicationDefaultCredential(FirebaseAuthTest.java:143)
        at com.google.firebase.auth.FirebaseAuthTest.data(FirebaseAuthTest.java:108)
```

```
Running com.google.firebase.auth.FirebaseCredentialsTest
Tests run: 15, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.118 sec <<< FAILURE! - in com.google.firebase.auth.FirebaseCredentialsTest
defaultCredentialDoesntRefetch(com.google.firebase.auth.FirebaseCredentialsTest)  Time elapsed: 0.012 sec  <<< ERROR!
java.lang.RuntimeException: java.io.IOException: The Application Default Credentials are not available. They are available if running in Google Compute Engine. Otherwise, the environment variable GOOGLE_APPLICATION_CREDENTIALS must be defined pointing to a file defining the credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
        at com.google.firebase.auth.FirebaseCredentialsTest.defaultCredentialDoesntRefetch(FirebaseCredentialsTest.java:86)
Caused by: java.io.IOException: The Application Default Credentials are not available. They are available if running in Google Compute Engine. Otherwise, the environment variable GOOGLE_APPLICATION_CREDENTIALS must be defined pointing to a file defining the credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
        at com.google.firebase.auth.FirebaseCredentialsTest.defaultCredentialDoesntRefetch(FirebaseCredentialsTest.java:86)
```

And unit tests are executed with `mvn test` too.

hiranya911:
The test case does set the `GOOGLE_APPLICATION_CREDENTIALS` variable using some reflection magic:

```
    Map<String, String> environmentVariables =
        ImmutableMap.<String, String>builder()
            .put("GOOGLE_APPLICATION_CREDENTIALS", credentialsFile.getAbsolutePath())
            .build();
    TestUtils.setEnvironmentVariables(environmentVariables);
```

Looks like this doesn't work in your environment. What is your OS and JVM build? We test this on Linux and Mac, where it works fine.

ghajba:
:)
it's currently Windows. I'll try it on a Mac too

hiranya911:
@ghajba Feel free to ignore those 2 failures. It's not related to your changes anyway. I'll try to come up with a way to skip those tests on Windows.

src/main/java/com/google/firebase/database/snapshot/NodeUtilities.java
dimipaun: I think BigDecimals should be serialized to String. The entire point of using BigDecimals is to avoid the rounding errors inherent in floating point representation.

src/main/java/com/google/firebase/database/snapshot/NodeUtilities.java
dimipaun: Ditto for BigInteger -- we should serialize these to String to maintain the full representation.

dimipaun:
Hi @ghajba and @hiranya911,

Sorry for jumping late into this, but I think BigDecimal should serialize to String. The number 1 reason for using BigDecimals (such as in code that deals with money) is for it's _exact_ representation of the fraction, rather than for the large values that it holds.

Serializing it to double throws this out the window.

hiranya911:
Over to @schmidt-sebastian who's been looking at #75. Based on the comment there, I assume we are not going to implement these changes?

schmidt-sebastian:
Sorry for being slow to respond on this PR and thank you for your awesome contribution! 
As you have probably seen, for now we have updated the error messages for both the RTDB and Firestore. The RTDB client uses JSON as its interop format, and FIrestore uses signed Int64 types. Both of these representations do not allow us to express BigIntegers correctly. While serializing to String is an accurate solution for this particular platform, the type doesn't round-trip very well and our other clients would not have the ability to treat these types as numbers. For that reason, I am leaning towards making this more explicit and asking you (and all other affected customers) to directly use String types instead.

schmidt-sebastian:
Closing this PR for now. If we get significant feedback from a number of customers, we can revisit this, in which case this PR can probably just be merged as is.

PR #83, opened 2017-10-06
Firestore API

hiranya911:
Usage:

```
Firestore fs = FirestoreClient.getFirestore();
```

PR #91, opened 2017-10-23
WebSocket Implementation Based on Netty

hiranya911:
Java Admin SDK's RTDB implementation has been using a homegrown WebSocket implementation (the entire `com.google.firebase.databas.tubesock` package). I'm sure there were good reasons for doing it that way, but that code was pretty complicated and barely maintainable.

Starting from #83 (Firestore integration), we have a dependency on Netty, which is a popular library with excellent WebSocket support. This PR gets rid of our homegrown WebSocket code, and implements it using Netty's `NioEventLoop`. This results in a net removal of around 1600 lines of code. I believe the final outcome is much simpler and far more maintainable.

Unlike our old implementation which uses 2 threads (reader and writer) for WebSocket communication, Netty uses a single-threaded event loop. I have also tested this locally, and on GAE (java8) for several days straight without any problems.


hiranya911:
@mikelehen I simplified this PR as advised by removing some of the refactoring commits. It should be easier to review now. The PR states 23 files have been changed, but actually 15 of those files are total removals. Changes are localized to just 8 files. 

The class that has changed most is `WebsocketConnection`. The others have fairly small changes.

src/main/java/com/google/firebase/database/connection/NettyWebSocketClient.java
```diff
@@ -0,0 +1,202 @@
...
+    @Override
+    public void channelInactive(ChannelHandlerContext context) {
+      delegate.onClose();
+    }
+
+    @Override
+    public void channelRead0(ChannelHandlerContext context, Object message) throws Exception {
+      Channel channel = context.channel();
+      if (!handshaker.isHandshakeComplete()) {
+        checkState(message instanceof FullHttpResponse);
```
mikelehen: This is running on the firebase-websocket-worker thread, right? What happens when we throw an exception? Does it get handled, logged, etc?

src/main/java/com/google/firebase/database/connection/NettyWebSocketClient.java
```diff
@@ -0,0 +1,202 @@
...
+      }
+
+      if (message instanceof FullHttpResponse) {
+        FullHttpResponse response = (FullHttpResponse) message;
+        String error = String.format("Unexpected FullHttpResponse (status: %s; content: %s)",
+            response.status().toString(), response.content().toString(CharsetUtil.UTF_8));
+        throw new IllegalStateException(error);
+      }
+
+      WebSocketFrame frame = (WebSocketFrame) message;
```
mikelehen: Is this cast useful / necessary since we follow it with more specific instanceof checks?

src/main/java/com/google/firebase/database/connection/NettyWebSocketClient.java
```diff
@@ -0,0 +1,202 @@
...
+      if (frame instanceof TextWebSocketFrame) {
+        delegate.onMessage(((TextWebSocketFrame) frame).text());
+      } else if (frame instanceof CloseWebSocketFrame) {
+        delegate.onClose();
+      }
```
mikelehen: should we have an else case that throws just as a sanity check that we're covering all cases?

src/main/java/com/google/firebase/database/connection/WebsocketConnection.java
```diff
@@ -108,78 +99,86 @@ public void run() {
...
     if (logger.logsDebug()) {
-      logger.debug("HandleNewFrameCount: " + totalFrames);
+      logger.debug("HandleNewFrameCount: " + numFrames);
+    }
+    buffer.initialize(numFrames);
+  }
+
+  private void appendFrame(String message) {
+    if (buffer.append(message) > 0) {
+      return;
```
mikelehen: it wasn't obvious that buffer.append() would return the number of remaining frames, so maybe add a comment:
```
return; // waiting for more frames
```

or use an intermediate variable:

```
int framesRemaining = buffer.append(message);
if (framesRemaining > 0) {
  return;
}
```

or ignore the return value:

```
buffer.append(message);
if (buffer.hasRemaining()) {
  return;
}
```

src/main/java/com/google/firebase/database/connection/NettyWebSocketClient.java
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    @Override
+    public void channelInactive(ChannelHandlerContext context) {
+      delegate.onClose();
+    }
+
+    @Override
+    public void channelRead0(ChannelHandlerContext context, Object message) throws Exception {
+      Channel channel = context.channel();
```
schmidt-sebastian: How about just doing:

```
         try {
            if (message instanceof FullHttpResponse) {
              checkState(!handshaker.isHandshakeComplete());
              handshaker.finishHandshake(channel, (FullHttpResponse) message);
              delegate.onOpen();
            } else if (message instanceof TextWebSocketFrame) {
              delegate.onMessage(((TextWebSocketFrame) frame).text());
            } else {
              checkState(message instanceof CloseWebSocketFrame);
              delegate.onClose();
            }
          } catch (Exception e) {
            delegate.onError(e);
          }
```

src/main/java/com/google/firebase/database/connection/WebsocketConnection.java
```diff
@@ -242,94 +245,65 @@ public void run() {
...
   }
 
-  private boolean isBuffering() {
-    return frameReader != null;
+  /**
+   * Closes the low-level connection, and notifies the higher layer ({@link Connection}.
+   */
+  private void closeAndNotify() {
+    close();
+    delegate.onDisconnect(everConnected);
```
mikelehen: I wonder if there should be an isClosed check before calling the delegate, unless you're sure that closeAndNotify() can't be called more than once or that the delegate will be tolerant of multiple callbacks.

src/main/java/com/google/firebase/database/connection/WebsocketConnection.java
```diff
@@ -338,67 +312,102 @@ public void onClose() {
...
       executorService.execute(
           new Runnable() {
             @Override
             public void run() {
-              if (e.getCause() != null && e.getCause() instanceof EOFException) {
-                logger.debug("WebSocket reached EOF.");
-              } else {
-                logger.debug("WebSocket error.", e);
-              }
               onClosed();
```
mikelehen: I don't think NettyWebsocketConnection always closes the connection before calling onError, so I wonder if this should be closeAndNotify() instead?

src/main/java/com/google/firebase/database/connection/WebsocketConnection.java
```diff
@@ -338,67 +312,102 @@ public void onClose() {
...
     }
+  }
 
-    @Override
-    public void onLogMessage(String msg) {
-      if (logger.logsDebug()) {
-        logger.debug("Tubesock: " + msg);
-      }
-    }
+  private static class StringList {
```
mikelehen: Can you comment this a bit?  It's a bit wonky since it's expecting a specific number of elements, unlike a normal list...  You may want to make it a FrameBuffer or something to make it more obvious that it's special. But I don't feel strongly.

src/main/java/com/google/firebase/database/connection/WebsocketConnection.java
```diff
@@ -338,67 +312,102 @@ public void onClose() {
...
 
-    @Override
-    public void send(String msg) {
-      ws.send(msg);
+    private final AtomicInteger remaining = new AtomicInteger(0);
+    private List<String> buffer;
+
+    void initialize(int capacity) {
```
mikelehen: Why not make this the constructor and just create a new instance for each websocket message?  I don't see a strong reason for this to be a reusable class.

src/main/java/com/google/firebase/database/connection/WebsocketConnection.java
```diff
@@ -338,67 +312,102 @@ public void onClose() {
...
-      if (logger.logsDebug()) {
-        logger.debug("Tubesock: " + msg);
-      }
-    }
+  private static class StringList {
 
-    @Override
-    public void send(String msg) {
-      ws.send(msg);
+    private final AtomicInteger remaining = new AtomicInteger(0);
```
mikelehen: Why AtomicInteger?  This class should only be accessed by a single thread, right?  Using AtomicInteger suggests otherwise which makes the code less clear.

src/main/java/com/google/firebase/database/connection/WebsocketConnection.java
schmidt-sebastian: I thought for the longest time (up until this very line) that you were using an existing class. How much do we gain from having our own custom implementation, versus just doing it inline in the calling class?

And if you do decide to keep it, a small comment would be nice.

src/main/java/com/google/firebase/database/connection/WebsocketConnection.java
```diff
@@ -338,67 +312,102 @@ public void onClose() {
...
-      } catch (InterruptedException e) {
-        logger.error("Interrupted while shutting down websocket threads", e);
-      }
+    boolean hasRemaining() {
+      return remaining.get() > 0;
     }
 
-    @Override
-    public void connect() {
+    String combine() {
```
schmidt-sebastian: This deservers a better name or at least a comment :) Thanks!

src/main/java/com/google/firebase/database/core/RepoManager.java
```diff
@@ -54,6 +55,8 @@ public static void destroy(Context ctx) {
       instance.destroyInternal(ctx);
     } finally {
       ctx.stop();
+      // TODO(hkj): https://github.com/netty/netty/issues/7310
```
mikelehen: Please make this actionable or remove it. (but if you remove the TODO but keep the removeAll() call, please add a comment explaining what it's for)

src/main/java/com/google/firebase/database/connection/NettyWebSocketClient.java
```diff
@@ -0,0 +1,202 @@
...
+ * a single-threaded NIO event loop to read and write bytes from a WebSocket connection. Netty
+ * handles all the low-level IO, SSL and WebSocket handshake, and other protocol-specific details.
+ *
+ * <p>This implementation does not initiate connection close on its own. In case of errors or loss
+ * of connectivity, it notifies the higher layer ({@link WebsocketConnection}), which then decides
+ * whether to initiate a connection tear down.
+ */
+class NettyWebSocketClient implements WebsocketConnection.WSClient {
+
+  private static final int DEFAULT_WSS_PORT = 443;
```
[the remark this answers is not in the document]
hiranya911: It turns out if the URI string is specified as `https://test.firebaseio.com` (which is what most users do), [URI.getPort()](https://docs.oracle.com/javase/7/docs/api/java/net/URI.html#getPort()) returns -1.

src/main/java/com/google/firebase/database/connection/WebsocketConnection.java
hiranya911: We were using a custom [`StringListReader`](https://github.com/firebase/firebase-admin-java/blob/03bc30bc99fbd66068d6c3c3661cd972b6e16de1/src/main/java/com/google/firebase/database/connection/util/StringListReader.java) class for this purpose, which was far more complex than it needed to be. I just created a simpler version of it. Added a class comment to explain its purpose. I do believe the util class makes the core a bit more readable.

src/main/java/com/google/firebase/database/connection/WebsocketConnection.java
hiranya911: Added comment.

src/main/java/com/google/firebase/database/core/Context.java
```diff
@@ -290,4 +284,4 @@ private String buildUserAgent(String platformAgent) {
             .append(platformAgent);
     return sb.toString();
   }
-}
+}
```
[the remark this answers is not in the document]
hiranya911: Done :)

src/main/java/com/google/firebase/internal/RevivingScheduledExecutor.java
```diff
@@ -88,6 +90,7 @@ public RevivingScheduledExecutor(
       final long initialDelayMs,
       final long timeoutMs) {
     super(0);
+    checkNotNull(threadFactory, "thread factory must not be null");
```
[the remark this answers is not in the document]
hiranya911: Done

src/main/java/com/google/firebase/database/connection/NettyWebSocketClient.java
```diff
@@ -0,0 +1,202 @@
...
+      final SslContext sslContext = SslContextBuilder.forClient()
+          .trustManager(InsecureTrustManagerFactory.INSTANCE).build();
+      Bootstrap bootstrap = new Bootstrap();
+      bootstrap.group(group)
+          .channel(NioSocketChannel.class)
+          .handler(new ChannelInitializer<SocketChannel>() {
+            @Override
+            protected void initChannel(SocketChannel ch) {
+              ChannelPipeline p = ch.pipeline();
+              p.addLast(sslContext.newHandler(ch.alloc(), uri.getHost(), DEFAULT_WSS_PORT));
```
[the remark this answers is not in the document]
hiranya911: `URI.getPort()` returns -1 when the source URI string doesn't explicitly state the port.

src/main/java/com/google/firebase/database/connection/NettyWebSocketClient.java
```diff
@@ -0,0 +1,202 @@
...
+              p.addLast(
+                  new HttpClientCodec(),
+                  new HttpObjectAggregator(8192),
```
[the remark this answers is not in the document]
hiranya911: I believe this only applies to the original handshake request. Once the HTTP connection has been upgraded to a 2-way WebSocket connection, this no longer has any effect. Having said that I've bumped this to 32K and also added a comment.

src/main/java/com/google/firebase/database/connection/NettyWebSocketClient.java
```diff
@@ -0,0 +1,202 @@
...
+
+  @Override
+  public void close() {
+    checkState(channel != null, "channel not initialized");
+    try {
+      channel.close();
+    } finally {
+      group.shutdownGracefully();
+      executorService.shutdown();
+      // TODO(hkj): https://github.com/netty/netty/issues/7310
```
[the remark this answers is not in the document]
hiranya911: Done

src/main/java/com/google/firebase/database/connection/NettyWebSocketClient.java
hiranya911: Cleaned up as per the suggestions from @schmidt-sebastian

src/main/java/com/google/firebase/database/connection/NettyWebSocketClient.java
hiranya911: Done

src/main/java/com/google/firebase/database/connection/WebsocketConnection.java
hiranya911: Done. Used option 2.

src/main/java/com/google/firebase/database/connection/WebsocketConnection.java
hiranya911: Delegate is idempotent with this regard. Once the connection has been marked closed at that level, it will start ignoring `onDisconnect()` events. But I think having that check makes the logic easier to follow. Added.

src/main/java/com/google/firebase/database/connection/WebsocketConnection.java
hiranya911: That is what this method (`onClosed()`) does:

```
private void onClosed() {
    if (!isClosed) {
      if (logger.logsDebug()) {
        logger.debug("closing itself");
      }
      closeAndNotify();
    }
  }
```

src/main/java/com/google/firebase/database/connection/WebsocketConnection.java
hiranya911: Added comments.

src/main/java/com/google/firebase/database/connection/WebsocketConnection.java
hiranya911: Done

src/main/java/com/google/firebase/database/connection/WebsocketConnection.java
hiranya911: Done

src/main/java/com/google/firebase/database/connection/NettyWebSocketClient.java
hiranya911: Tested by injecting some exceptions. It fires the `NettyWebSocketClient.exceptionCaught()` event, which we hand off to `delegate.onError()`. The net result is, the existing connection gets closed gracefully, and the client moves on to the next connection attempt.

src/main/java/com/google/firebase/database/connection/NettyWebSocketClient.java
hiranya911: Done

src/main/java/com/google/firebase/database/core/RepoManager.java
```diff
@@ -54,6 +55,8 @@ public static void destroy(Context ctx) {
...
+      FastThreadLocal.removeAll();
```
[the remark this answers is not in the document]
hiranya911: Tested with a multi-app set up. Seems to work fine. After the first app has been deleted, the 2nd app continues to receive updates. I'll continue to test this scenario a bit deeply.

src/main/java/com/google/firebase/database/core/RepoManager.java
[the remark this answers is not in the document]
hiranya911: Done.

src/main/java/com/google/firebase/database/connection/NettyWebSocketClient.java
```diff
@@ -0,0 +1,202 @@
...
+        threadFactory, "firebase-websocket-worker", GaeThreadFactory.isAvailable());
+    this.group = new NioEventLoopGroup(1, this.executorService);
+  }
+
+  @Override
+  public void connect() {
+    checkState(channel == null, "channel already initialized");
+    try {
+      final SslContext sslContext = SslContextBuilder.forClient()
+          .trustManager(InsecureTrustManagerFactory.INSTANCE).build();
```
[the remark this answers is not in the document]
hiranya911: Used a proper trust manager. Thanks for pointing it out.

src/main/java/com/google/firebase/database/core/RepoManager.java
hiranya911: Updated

hiranya911:
Made the suggested changes. Over to you for the next round.

src/main/java/com/google/firebase/database/connection/NettyWebSocketClient.java
mikelehen: If this check fails, that means there's a  bug in this code and we should at minimum log it (and crashing the process wouldn't be out-of-the-question).  So if this just silently retries without logging anything, then I don't think that's sufficient.

src/main/java/com/google/firebase/database/connection/NettyWebSocketClient.java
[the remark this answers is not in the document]
mikelehen: I am fine with using DEFAULT_WSS_PORT in that case, but if you explicitly specify a port, this code should respect it.  It may not matter much in practice, since prod firebase uses 443, but if somebody on the RTDB team wanted to test against a localhost instance or something, they might need to use a different port.

src/main/java/com/google/firebase/database/connection/NettyWebSocketClient.java
hiranya911: `delegate.onError()` logs the exception (delegate is `WebsocketConnection.WSClientHandlerImpl` in this case):

```
      if (e.getCause() != null && e.getCause() instanceof EOFException) {
        logger.error("WebSocket reached EOF", e);
      } else {
        logger.error("WebSocket error", e);
      }
```

mikelehen: 👍

src/main/java/com/google/firebase/database/connection/NettyWebSocketClient.java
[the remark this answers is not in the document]
hiranya911: Fair enough. Changed to:

```
int port = uri.getPort() != -1 ? uri.getPort() : DEFAULT_WSS_PORT;
```

src/main/java/com/google/firebase/database/connection/WebsocketConnection.java
```diff
@@ -16,43 +16,51 @@
...
+ * higher layer when necessary. Higher layer signals this implementation when it needs to send a
+ * message out, or when a graceful connection tear down should be initiated.
+ */
 class WebsocketConnection {
 
   private static final long KEEP_ALIVE_TIMEOUT_MS = 45 * 1000; // 45 seconds
   private static final long CONNECT_TIMEOUT_MS = 30 * 1000; // 30 seconds
   private static final int MAX_FRAME_SIZE = 16384;
-  private static long connectionId = 0;
+  private static final AtomicLong CONN_ID = new AtomicLong(0);
```
schmidt-sebastian: @mikelehen's comment regarding the suggested thread safety applies here as well.

src/main/java/com/google/firebase/database/connection/NettyWebSocketClient.java
```diff
@@ -0,0 +1,197 @@
...
+            @Override
+            protected void initChannel(SocketChannel ch) {
+              ChannelPipeline p = ch.pipeline();
+              p.addLast(sslContext.newHandler(ch.alloc(), uri.getHost(), port));
+              p.addLast(
+                  new HttpClientCodec(),
+                  // Set the max size for the HTTP responses. This only applies to the WebSocket
+                  // handshake response from the server.
+                  new HttpObjectAggregator(32 * 1024),
+                  WebSocketClientCompressionHandler.INSTANCE,
```
rockwotj: We don't compress websocket frames and don't plan on it, so this probably isn't needed

src/main/java/com/google/firebase/database/connection/WebsocketConnection.java
mikelehen: I almost called this out too, but actually since this is static it can technically be accessed by multiple threads (each Repo has its own run loop with its own thread and so these global connection ids could be assigned simultaneously on different threads).  So I think AtomicLong is applicable here (although a race is unlikely in practice and wouldn't be particularly harmful, since the ID is just used for logging).

src/main/java/com/google/firebase/database/connection/NettyWebSocketClient.java
hiranya911: Done. Thanks @rockwotj

PR #92, opened 2017-10-31
Custom Claims and List Users Support

hiranya911:
Adding two new operations to the `FirebaseAuth` API:

* `listUsersAsync(String pageToken, int maxResults)`: Facilitates paging through all the user accounts of a Firebase project.
* `setCustomClaimsAsync(String uid, Map<String, Object> claims)`: Sets custom attributes on a user account.

go/firebase-admin-custom-claims


hiranya911:
@mikelehen This is a recently approved API (already released in Node.js). Please review when you get a chance. Not urgent as of now.

src/main/java/com/google/firebase/auth/ExportedUserRecord.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+package com.google.firebase.auth;
+
+import com.google.firebase.auth.internal.DownloadAccountResponse.User;
+import com.google.firebase.internal.Nullable;
+
+/**
+ * Contains metadata associated with a Firebase user account, along with password hash and salt.
+ * Instances of this class are immutable and thread safe.
```
schmidt-sebastian: s/thread safe/thread-safe.

src/main/java/com/google/firebase/auth/ExportedUserRecord.java
```diff
@@ -0,0 +1,46 @@
...
+  private final String passwordHash;
+  private final String passwordSalt;
+
+  ExportedUserRecord(User response) {
+    super(response);
+    this.passwordHash = response.getPasswordHash();
+    this.passwordSalt = response.getPasswordSalt();
+  }
+
+  @Nullable
```
schmidt-sebastian: Please add documentation here and below. Also explain when this might be null.

src/main/java/com/google/firebase/auth/FirebaseUserManager.java
```diff
@@ -187,6 +196,28 @@ void deleteUser(String uid) throws FirebaseAuthException {
...
+    }
+
+    DownloadAccountResponse response;
+    try {
+      response = post("downloadAccount", builder.build(), DownloadAccountResponse.class);
+    } catch (IOException e) {
+      throw new FirebaseAuthException(LIST_USERS_ERROR,
+          "IO error while downloading user accounts.", e);
+    }
+    if (response == null) {
```
schmidt-sebastian: Nit: this might be more readable if you move this into the try block.

src/main/java/com/google/firebase/auth/ListUsersPage.java
```diff
@@ -0,0 +1,260 @@
...
+
+  private final UserSource source;
+  private final int maxResults;
+
+  private final ListUsersResult currentBatch;
+
+  private ListUsersPage(@NonNull UserSource source, int maxResults, @Nullable String pageToken) {
+    this.source = source;
+    this.maxResults = maxResults;
+    this.currentBatch = source.fetch(maxResults, pageToken);
```
schmidt-sebastian: This is performing a synchronous network operation inside the constructor. Can you move this code to the accessors?

src/main/java/com/google/firebase/auth/ListUsersPage.java
```diff
@@ -0,0 +1,260 @@
...
+    DefaultUserSource(FirebaseUserManager userManager) {
+      this.userManager = checkNotNull(userManager, "user manager must not be null");
+    }
+
+    @Override
+    public ListUsersResult fetch(int maxResults, String pageToken) {
+      try {
+        DownloadAccountResponse response = userManager.listUsers(maxResults, pageToken);
+        ImmutableList.Builder<ExportedUserRecord> builder = ImmutableList.builder();
+        if (response.getUsers() != null) {
```
schmidt-sebastian: s/getUsers/hasUsers()/?

src/main/java/com/google/firebase/auth/ListUsersPage.java
```diff
@@ -0,0 +1,260 @@
...
+          for (DownloadAccountResponse.User user : response.getUsers()) {
+            builder.add(new ExportedUserRecord(user));
+          }
+        }
+        String nextPageToken = response.getPageToken() != null
+            ? response.getPageToken() : END_OF_LIST;
```
schmidt-sebastian: Have you though about using "null" to mark the final page token and just validating that the end user never passes it in?

src/main/java/com/google/firebase/auth/UserRecord.java
```diff
@@ -73,6 +79,17 @@
       }
     }
     this.userMetadata = new UserMetadata(response.getCreatedAt(), response.getLastLoginAt());
+    String customClaims = response.getCustomClaims();
+    if (!Strings.isNullOrEmpty(customClaims)) {
+      Map<String, Object> parsed = new JSONObject(customClaims).toMap();
+      if (parsed.isEmpty()) {
+        this.customClaims = null;
+      } else {
+        this.customClaims = ImmutableMap.copyOf(parsed);
```
schmidt-sebastian: Is it possible to delete a single custom claim?

src/main/java/com/google/firebase/auth/UserRecord.java
```diff
@@ -177,6 +194,16 @@ public UserMetadata getUserMetadata() {
   }
 
   /**
+   * Returns additional custom claims set on this user.
```
schmidt-sebastian: I don't understand what additional means in this context. I suppose it returns all the custom claims?

src/main/java/com/google/firebase/auth/UserRecord.java
```diff
@@ -177,6 +194,16 @@ public UserMetadata getUserMetadata() {
...
+   *
+   * @return a non-empty Map of custom claims or null.
+   */
+  @Nullable
+  public Map<String,Object> getCustomClaims() {
+    return customClaims;
```
schmidt-sebastian: I think we should return an immutable copy here.

Also - can we return an empty map if there are no custom claims?

src/main/java/com/google/firebase/auth/UserRecord.java
```diff
@@ -204,6 +231,32 @@ private static void checkPassword(String password) {
...
+    }
+    for (String key : customClaims.keySet()) {
+      checkArgument(!FirebaseUserManager.RESERVED_CLAIMS.contains(key),
+          "Claim '" + key + "' is reserved and cannot be set");
+    }
+  }
+
+  private static String serializeCustomClaims(Map customClaims, JsonFactory jsonFactory) {
+    checkNotNull(jsonFactory, "JsonFactory must not be null");
+    if (customClaims == null || customClaims.isEmpty()) {
```
schmidt-sebastian: If you initialize customClaims to an empty map, this would go away as well.

src/main/java/com/google/firebase/auth/ExportedUserRecord.java
hiranya911: Done

src/main/java/com/google/firebase/auth/FirebaseUserManager.java
hiranya911: Done

src/main/java/com/google/firebase/auth/ListUsersPage.java
hiranya911: Users should be able to pass `null` here, which stands for the first page.

src/main/java/com/google/firebase/auth/UserRecord.java
hiranya911: Removed `additional`

src/main/java/com/google/firebase/auth/UserRecord.java
hiranya911: Setting to empty map when there are no claims. This always get set to ImmutableMap in the constructor.

src/main/java/com/google/firebase/auth/UserRecord.java
hiranya911: This can be a value directly passed in by the user via `UpdateRequest.setCustomClaims()`. Therefore `null` values are allowed here.

src/main/java/com/google/firebase/auth/UserRecord.java
hiranya911: Something like:

```
Map<String, Object> copy = new HashMap(userRecord.getCustomClaims());
copy.remove("someClaim");
updateRequest.setCustomClaims(copy);
```

src/main/java/com/google/firebase/auth/ListUsersPage.java
hiranya911: Done

src/main/java/com/google/firebase/auth/ListUsersPage.java
hiranya911: This class is much easier to reason about when it's immutable. Therefore I'd rather fetch the data upon construction, and assign to a `final` attribute. Also lazy loading is not a requirement for this class. `FirebaseAuth.listUsers()` and `Page.getNextPage()` are expected to make RPC calls, and return fully populated pages as the result.

We can move the RPC call out of the constructor, but since we create pages in 2 places the constructor seems like a natural place to place that bit of code. Alternatively we can try a private `ensureCurrentBatch()` method to load the data on demand, but we are not really saving anything since all operations on `Page` will have to call it.

hiranya911:
@schmidt-sebastian Made most of the suggested changes. Provided explanations on the rest. On to the next round.

src/main/java/com/google/firebase/auth/FirebaseUserManager.java
```diff
@@ -187,6 +196,28 @@ void deleteUser(String uid) throws FirebaseAuthException {
     }
   }
 
+  DownloadAccountResponse listUsers(int maxResults, String pageToken) throws FirebaseAuthException {
+    ImmutableMap.Builder<String, Object> builder = ImmutableMap.<String, Object>builder()
+        .put("maxResults", maxResults);
+    if (pageToken != null) {
+      checkArgument(!pageToken.isEmpty(), "invalid end of list page token");
```
schmidt-sebastian: How about pageToken.equals(ListUsersPage.END_OF_LIST)? That way,  that implementation detail only exists in one place.

src/main/java/com/google/firebase/auth/ListUsersPage.java
schmidt-sebastian: This is generally frowned upon as it makes testing hard (SeeL http://misko.hevery.com/code-reviewers-guide/flaw-constructor-does-real-work/).

One easy way to mitigate would be to create a factory method and pass in `currentBatch`.

src/test/java/com/google/firebase/auth/FirebaseUserManagerTest.java
```diff
@@ -344,47 +454,85 @@ public void testInvalidPassword() {
...
+  @Test
+  public void testNullJsonFactory() {
+    UpdateRequest update = new UpdateRequest("test");
+    Map<String, Object> claims = ImmutableMap.<String, Object>of("admin", true, "package", "gold");
+    update.setCustomClaims(claims);
+    try {
+      update.getProperties(null);
+      fail("No error thrown for null JsonFactory");
+    } catch (Exception ignore) {
+      // expected
```
schmidt-sebastian: Are there any specifics you can verify here?

src/main/java/com/google/firebase/auth/FirebaseUserManager.java
hiranya911: Good call. Done.

src/test/java/com/google/firebase/auth/FirebaseUserManagerTest.java
hiranya911: Done. Checking for NPE now.

hiranya911:
Made the suggested changes. Moved the RPC calls in `ListUsersPage` constructor to the `PageFactory` helper class,

schmidt-sebastian:
Thanks so much! I am a big fan.

PR #115, opened 2017-12-20
Improved error handling for user management API

hiranya911:
User management API currently returns a set of custom error codes specific to Java Admin SDK. This PR changes the implementation to:
* parse the error messages/codes sent by the ID toolkit backend service, and
* map errors to [well-defined](https://firebase.google.com/docs/auth/admin/errors) Admin SDK error codes.

### DIFF
--- a/src/main/java/com/google/firebase/auth/FirebaseUserManager.java
+++ b/src/main/java/com/google/firebase/auth/FirebaseUserManager.java
@@ -23,6 +23,7 @@ import com.google.api.client.http.GenericUrl;
 import com.google.api.client.http.HttpRequest;
 import com.google.api.client.http.HttpRequestFactory;
 import com.google.api.client.http.HttpResponse;
+import com.google.api.client.http.HttpResponseException;
 import com.google.api.client.http.HttpResponseInterceptor;
 import com.google.api.client.http.HttpTransport;
 import com.google.api.client.http.json.JsonHttpContent;
@@ -40,6 +41,7 @@ import com.google.firebase.auth.UserRecord.UpdateRequest;
 import com.google.firebase.auth.internal.DownloadAccountResponse;
 import com.google.firebase.auth.internal.GetAccountInfoResponse;
 
+import com.google.firebase.auth.internal.HttpErrorResponse;
 import com.google.firebase.internal.SdkUtils;
 import java.io.IOException;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/firebase/auth/FirebaseUserManager.java
```diff
@@ -54,12 +56,27 @@
...
-  static final String USER_CREATE_ERROR = "USER_CREATE_ERROR";
-  static final String USER_UPDATE_ERROR = "USER_UPDATE_ERROR";
-  static final String USER_DELETE_ERROR = "USER_DELETE_ERROR";
-  static final String LIST_USERS_ERROR = "LIST_USERS_ERROR";
-  static final String INTERNAL_ERROR = "INTERNAL_ERROR";
+  static final String USER_NOT_FOUND_ERROR = "user-not-found";
+  static final String INTERNAL_ERROR = "internal-error";
+
+  // Map of server-side error codes to SDK error codes.
+  // SDK error codes defined at: https://firebase.google.com/docs/auth/admin/errors
```
avishalom: I might be missing something, shouldn't this map should be eyed on the error code values in that link e.g. "auth/claims-too-large", not the upper cased version. (later we get from the map based on getErrorCode())

hiranya911: The backend server sends the upper-cased error codes. We need to map them to the client-side error codes defined in the public documentation.

src/main/java/com/google/firebase/auth/FirebaseUserManager.java
```diff
@@ -149,50 +148,31 @@ UserRecord getUserByPhoneNumber(String phoneNumber) throws FirebaseAuthException
...
+    GenericJson response = post(
+        "signupNewUser", request.getProperties(), GenericJson.class);
     if (response != null) {
       String uid = (String) response.get("localId");
       if (!Strings.isNullOrEmpty(uid)) {
         return uid;
       }
     }
-    throw new FirebaseAuthException(USER_CREATE_ERROR, "Failed to create new user");
+    throw new FirebaseAuthException(INTERNAL_ERROR, "Failed to create new user");
```
schmidt-sebastian: If the UID is empty, is there indication of what is went wrong in the request body that we could show in this error?

src/main/java/com/google/firebase/auth/FirebaseUserManager.java
```diff
@@ -204,36 +184,61 @@ DownloadAccountResponse listUsers(int maxResults, String pageToken) throws Fireb
...
-      }
-      return response;
-    } catch (IOException e) {
-      throw new FirebaseAuthException(LIST_USERS_ERROR,
-          "IO error while downloading user accounts.", e);
+    DownloadAccountResponse response = post(
+        "downloadAccount", builder.build(), DownloadAccountResponse.class);
+    if (response == null) {
+      throw new FirebaseAuthException(INTERNAL_ERROR,
+          "Unexpected response from download user account API.");
```
schmidt-sebastian: Do you want to match you other error messages and say "Failed to retrieve..."?

hiranya911: Done

src/main/java/com/google/firebase/auth/FirebaseUserManager.java
hiranya911: I don't think that situation actually occurs in this REST API. If something went wrong, it should respond with a different HTTP status. The check here is to prevent the SDK from returning a null/empty UID in the rare case the server misbehaves.

PR #132, opened 2018-01-26
Logging any errors encountered while executing a DB event handler

### LINKED ISSUE #79: Exceptions swallowed in snapshot.getValue()
I was looking for hours why getValue doesn't return anything. The thread dies in the background silently. First as i step debug the code i've seen that some things must be present in the pojos like empty constructor and getters at least. I don't know why the exception is swallowed. I just can assume that google endpoints hide it from my eyes. I am using firebase with cloud endpoints together.

hiranya911:
Logging errors thrown by event handlers. Resolves #79. Without this errors thrown by event handlers are silently discarded.

### Example code:
```
ref.addListenerForSingleValueEvent(new ValueEventListener() {
      @Override
      public void onDataChange(DataSnapshot snapshot) {
        throw new RuntimeException("test error");
      }

      @Override
      public void onCancelled(DatabaseError error) {
      }
    });
```

### Log:
```
[FirebaseDatabaseEventTarget] ERROR com.google.firebase.database.core.ThreadPoolEventTarget - Event handler threw an exception
java.lang.RuntimeException: test error
	at com.google.firebase.database.integration.EventTestIT$17.onDataChange(EventTestIT.java:1160)
	at com.google.firebase.database.Query$1.onDataChange(Query.java:182)
	at com.google.firebase.database.core.ValueEventRegistration.fireEvent(ValueEventRegistration.java:77)
	at com.google.firebase.database.core.view.DataEvent.fire(DataEvent.java:65)
	at com.google.firebase.database.core.view.EventRaiser$1.run(EventRaiser.java:58)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

### DIFF
--- a/src/main/java/com/google/firebase/database/core/ThreadPoolEventTarget.java
+++ b/src/main/java/com/google/firebase/database/core/ThreadPoolEventTarget.java
@@ -24,9 +24,13 @@ import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /** ThreadPoolEventTarget is an event target using a configurable threadpool. */
 class ThreadPoolEventTarget implements EventTarget, UncaughtExceptionHandler {
+
+  private static final Logger logger = LoggerFactory.getLogger(ThreadPoolEventTarget.class);
 
   private final ThreadPoolExecutor executor;
   private UncaughtExceptionHandler exceptionHandler;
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/firebase/database/core/ThreadPoolEventTarget.java
```diff
@@ -92,8 +96,12 @@ public void uncaughtException(Thread t, Throwable e) {
     synchronized (this) {
       delegate = exceptionHandler;
     }
-    if (delegate != null) {
-      delegate.uncaughtException(t, e);
+    try {
```
schmidt-sebastian: Nit: I would wrap all of this in the try/finally to reduce the number of braces back to back.

PR #136, opened 2018-02-07
Accept prefixed topic names

hiranya911:
The new FCM API does not accept topic names prefixed with `/topics/`. But we can easily check for this in the SDK and remove it. This results in a better and consistent dev experience since the other methods in this API accepts prefixed topic names.

### DIFF
--- a/src/main/java/com/google/firebase/messaging/Message.java
+++ b/src/main/java/com/google/firebase/messaging/Message.java
@@ -68,19 +68,25 @@ public class Message {
     this.androidConfig = builder.androidConfig;
     this.webpushConfig = builder.webpushConfig;
     this.apnsConfig = builder.apnsConfig;
+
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/firebase/messaging/Message.java
```diff
@@ -68,19 +68,25 @@ private Message(Builder builder) {
     this.androidConfig = builder.androidConfig;
     this.webpushConfig = builder.webpushConfig;
     this.apnsConfig = builder.apnsConfig;
+
+    String unprefixedTopic = null;
```
schmidt-sebastian: Can you move this to a helper function and call this stripPrefix()? I think the name unprefixedTopic will never win my heart over :)

hiranya911: Done :+1:

PR #146, opened 2018-02-26
Introducing Synchronous Methods to Public APIs

hiranya911:
We are now approaching the final phase of the thread model refactoring work we started last year (#55). So far we have deprecated the `Task` API and all the methods that return a `Task`. This PR removes those deprecated public methods, and adds a set of synchronous methods in their place. For instance `FirebaseAuth` will now have:

```
String createCustomToken(String uid) // Added in this PR
ApiFuture<String> createCustomTokenAsync(String uid) // Already exists
```

This is a breaking change, and hence will not be merged/released anytime soon. Instead I'll use this PR as a way to set up the `v6` branch, and develop other breaking changes planned for the next major release on top it.

go/firebase-admin-java-sync


src/main/java/com/google/firebase/database/DatabaseReference.java
```diff
@@ -177,7 +175,7 @@ public DatabaseReference push() {
    * @return The ApiFuture for this operation.
    */
   public ApiFuture<Void> setValueAsync(Object value) {
-    return new TaskToApiFuture<>(setValue(value));
+    return setValueInternal(value, PriorityUtilities.parsePriority(this.path, null), null);
```
schmidt-sebastian: Are we planning on either renaming these back to setValue() or on adding blocking methods?

src/test/java/com/google/firebase/internal/CallableOperationTest.java
```diff
@@ -0,0 +1,113 @@
...
+  public void testCallResult() throws Exception {
+    FirebaseApp app = FirebaseApp.initializeApp(OPTIONS);
+    CallableOperation<String,Exception> operation = new CallableOperation<String, Exception>() {
+      @Override
+      protected String execute() throws Exception {
+        String threadName = Thread.currentThread().getName();
+        if (TEST_FIREBASE_THREAD.equals(threadName)) {
+          return threadName;
+        }
+        return SAME_THREAD;
```
schmidt-sebastian: Should this just return the thread name? Or a boolean/enum with the result of the thread check?

src/test/java/com/google/firebase/testing/GenericFunction.java
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.firebase.testing;
+
+/**
+ * A generic operation that can be called with arbitrary arguments.
+ */
+public interface GenericFunction<T> {
```
schmidt-sebastian: I don't see any code that uses more than one argument. If I remember correctly, this Admin SDK is also moving to Java 7. Can you then just use https://docs.oracle.com/javase/8/docs/api/java/util/function/Function.html?

src/main/java/com/google/firebase/database/DatabaseReference.java
hiranya911: There are no plans to introduce blocking methods to database, since the implementation is asynchronous all the way down to the core. Developers can wait on the futures if they need blocking-like semantics.

src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -187,7 +185,11 @@ public static FirebaseApp initializeApp() {
    * {@link #initializeApp()} method.
    */
   public static FirebaseApp initializeApp(String name) {
-    return initializeApp(getOptionsFromEnvironment(), name);
+    try {
+      return initializeApp(getOptionsFromEnvironment(), name);
+    } catch (IOException e) {
+      throw new IllegalArgumentException("Failed to load settings from the environment", e);
```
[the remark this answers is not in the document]
hiranya911: Done

src/main/java/com/google/firebase/auth/FirebaseAuth.java
```diff
@@ -114,305 +118,351 @@ public static synchronized FirebaseAuth getInstance(FirebaseApp app) {
...
    * @param uid The UID to store in the token. This identifies the user to other Firebase services
-   *     (Firebase Database, Firebase Auth, etc.)
-   * @return A {@link Task} which will complete successfully with the created Firebase Custom Token,
-   *     or unsuccessfully with the failure Exception.
-   * @deprecated Use {@link #createCustomTokenAsync(String)}
+   *     (Firebase Realtime Database, Firebase Auth, etc.).
+   * @return A Firebase custom token string.
+   * @throws IllegalArgumentException If the specified uid is null or empty, or if the app has not
+   *     been initialized with service account credentials.
+   * @throws FirebaseAuthException If an error occurs while minting the custom token.
```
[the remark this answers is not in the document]
hiranya911: Changed to "generating"

src/test/java/com/google/firebase/internal/CallableOperationTest.java
hiranya911: Done

src/test/java/com/google/firebase/testing/GenericFunction.java
hiranya911: There are some multi arg function calls in FCM tests. But you're right. Once we are on java8 we should be able to clean this up significantly with native `Function` interface and lambdas. Specifically, I plan to replace most anonymous `CallableOp` instances with lambdas.

hiranya911:
Made the suggested changes. As for calling async methods from sync ones, we are trying to provide a stronger guarantee here than merely providing blocking semantics. We are trying to ensure that blocking methods actually execute on the calling thread, and are guaranteed to not spawn any new threads under the hood. This means developers can easily plug the SDK into thread-constrained runtimes (GAE, other J2EE containers etc) and run everything on runtime-provisioned threads if they want to.

src/main/java/com/google/firebase/messaging/FirebaseMessaging.java
```diff
@@ -208,48 +231,42 @@ public TopicManagementResponse call() throws FirebaseMessagingException {
...
+        .callAsync(app);
   }
 
-  private Task<TopicManagementResponse> unsubscribeFromTopic(
-      final List<String> registrationTokens, final String topic) {
-    checkRegistrationTokens(registrationTokens);
-    checkTopic(topic);
-
-    return ImplFirebaseTrampolines.submitCallable(app, new Callable<TopicManagementResponse>() {
+  private CallableOperation<String, FirebaseMessagingException> makeSendRequest(
```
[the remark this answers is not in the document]
hiranya911: Good idea. Changed to `sendOp` and `manageTopicOp`

src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -187,7 +185,12 @@ public static FirebaseApp initializeApp() {
    * {@link #initializeApp()} method.
    */
   public static FirebaseApp initializeApp(String name) {
-    return initializeApp(getOptionsFromEnvironment(), name);
+    try {
+      return initializeApp(getOptionsFromEnvironment(), name);
+    } catch (IOException e) {
+      throw new IllegalArgumentException(
```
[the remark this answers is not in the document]
hiranya911: Done

hiranya911:
## DO NOT MERGE (BREAKING CHANGES)

src/main/java/com/google/firebase/FirebaseApp.java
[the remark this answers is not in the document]
avishalom: moving comment ...

dimipaun:
Hi @hiranya911, as I commented already, removing the Task based API precludes us from sharing code between Android and the Server. This is a non-trivial problem, and it completely breaks a lot of our testing harness that shares code from Android.

hiranya911:
@dimipaun This is not a use case we intend to support. FWIW, we did not intend to support it in the past either. It sort of worked that way due to the similarity in the API signatures. But we certainly don't want developers to rely on that behavior. We want the Admin SDK, and the Android SDK to evolve independently. Specifically, we want the Admin SDK to follow the conventions established by the other server-side gcloud libraries. Therefore supporting an adhoc `Task` API is both awkward and expensive for us.

To be honest I'm not sure how you even got the same code to work in these 2 environments. The `Task` interface in Android and Admin SDK belong to different packages. So at least their import statements should be different. They are really 2 different classes, that happened to have the same short name, and the same public API surface. That is not really code sharing. It is relying on the syntactic similarity between two different APIs, which is always risky.

You may have noticed that this change is now in effect in the 6.0.0 release. You can continue to use the 5.x.x versions of the SDK while you figure out a migration plan. I would highly encourage you to [migrate](https://medium.com/google-cloud/firebase-asynchronous-operations-with-admin-java-sdk-82ca9b4f6022) your server-side components to use the `ApiFuture` interface. If syntactic similarity with `Task` is critical, you can try implementing your own `Task` wrapper around the futures (you can even try to implement the actual [`Task` interface from Android gms](https://developers.google.com/android/reference/com/google/android/gms/tasks/Task), and attempt actual code sharing with compiler support).

PR #147, opened 2018-02-26
Improved Auth Unit Test Coverage

hiranya911:
Unit tests currently test the `FirebaseUserManager` internal class, but does not test the public API surface in `FirebaseAuth`. As a result the `com.google.firebase.auth` package has a test coverage of only 69%. This PR modifies the tests so that we test the public API surface, bringing the package test coverage to 91%. 

The overall test coverage goes from 81% to 82%.

### DIFF
--- a/src/test/java/com/google/firebase/auth/FirebaseUserManagerTest.java
+++ b/src/test/java/com/google/firebase/auth/FirebaseUserManagerTest.java
@@ -18,6 +18,7 @@ package com.google.firebase.auth;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
@@ -237,9 +238,7 @@ public class FirebaseUserManagerTest {
 
   @Test
   public void testSetCustomAttributes() throws Exception {
-    initializeAppForUserManagement(
-        TestUtils.loadResource("createUser.json"),
-        TestUtils.loadResource("getUser.json"));
+    initializeAppForUserManagement(TestUtils.loadResource("createUser.json"));
     FirebaseAuth auth = FirebaseAuth.getInstance();
     FirebaseUserManager userManager = auth.getUserManager();
     TestResponseInterceptor interceptor = new TestResponseInterceptor();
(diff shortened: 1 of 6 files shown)


src/test/java/com/google/firebase/auth/FirebaseUserManagerTest.java
```diff
@@ -27,270 +28,265 @@
...
+    initializeAppForUserManagement("{}");
+    FirebaseAuth auth = FirebaseAuth.getInstance();
+    FirebaseUserManager userManager = auth.getUserManager();
     TestResponseInterceptor interceptor = new TestResponseInterceptor();
     userManager.setInterceptor(interceptor);
-    DownloadAccountResponse download = userManager.listUsers(999, null);
-    assertNull(download.getUsers());
-    assertNull(download.getPageToken());
+    ListUsersPage page = auth.listUsersAsync(null).get();
+    assertEquals(0, Iterables.size(page.getValues()));
```
schmidt-sebastian: assertTrue(Iterables .isEmpty)

src/test/java/com/google/firebase/auth/FirebaseUserManagerTest.java
```diff
@@ -27,270 +28,265 @@
...
     interceptor.getResponse().getRequest().getContent().writeTo(out);
+    JsonFactory jsonFactory = Utils.getDefaultJsonFactory();
     GenericJson parsed = jsonFactory.fromString(new String(out.toByteArray()), GenericJson.class);
     assertEquals("testuser", parsed.get("localId"));
     assertEquals(jsonFactory.toString(claims), parsed.get("customAttributes"));
   }
 
+  @Test
+  public void testRevokeRefreshTokens() throws Exception {
+    initializeAppForUserManagement(TestUtils.loadResource("createUser.json"));
```
schmidt-sebastian: Could `initializeAppForUserManagement` be a setup() method that also encompasses the next 5 lines?

src/test/java/com/google/firebase/testing/MultiRequestMockHttpTransport.java
```diff
@@ -0,0 +1,46 @@
...
+import java.io.IOException;
+import java.util.List;
+
+/**
+ * A mock HttpTransport that can simulate multiple (sequential) HTTP interactions. This can be
+ * used when an SDK operation makes multiple backend calls.
+ */
+public class MultiRequestMockHttpTransport extends MockHttpTransport {
+
+  private final List<MockLowLevelHttpResponse> responses;
```
schmidt-sebastian: This could be a Queue: https://docs.oracle.com/javase/7/docs/api/java/util/Queue.html

You can then just .poll() the first element.

src/test/java/com/google/firebase/testing/TestResponseInterceptor.java
```diff
@@ -18,21 +18,31 @@
...
 
   public HttpResponse getResponse() {
-    return response;
+    if (responses.isEmpty()) {
+      return null;
+    }
+    return Iterables.getLast(responses);
+  }
+
+  public HttpResponse getResponse(int index) {
```
schmidt-sebastian: Is this method being used?

src/test/java/com/google/firebase/auth/FirebaseUserManagerTest.java
hiranya911: Done

src/test/java/com/google/firebase/auth/FirebaseUserManagerTest.java
hiranya911: The mock response is different in each test case (getUser.json, getUserError.json, listUsers.json etc.) Also half the tests in this file don't need it.

schmidt-sebastian: FWIW, you could still pass the *.json file to the setup() method.

src/test/java/com/google/firebase/testing/MultiRequestMockHttpTransport.java
hiranya911: Good idea. Although I prefer `Queue.remove()` in this case which throws an exception for empty queues.

src/test/java/com/google/firebase/testing/TestResponseInterceptor.java
hiranya911: Removed

src/test/java/com/google/firebase/auth/FirebaseUserManagerTest.java
hiranya911: Ok, I didn't add a setup method, but managed to significantly simplify this by moving rest of the user manager init code into the same helper method.

hiranya911:
Made the suggested changes.

src/test/java/com/google/firebase/testing/TestResponseInterceptor.java
```diff
@@ -41,8 +41,4 @@ public HttpResponse getResponse() {
     }
     return Iterables.getLast(responses);
```
schmidt-sebastian: `responses` can now just be a single HttpResponse.

hiranya911: Done

PR #153, opened 2018-03-12
Dropping Support for App Engine Java 7 Runtime

hiranya911:
In order to support Java 7 runtime in App Engine, we had implemented a number of abstractions and workarounds:

* `GaeThreadFactory` and `GaeExecutorService` for leveraging background threads support
* `GaePlatform` class for injecting the GAE threads into RTDB client code
* `RevivingScheduledExecutor` for periodically restarting threads, to work around GAE thread deadlines

With the new Java 8 runtime, none of the above is necessary. We can use regular threads instead of the expensive background threads, and the threads are no longer killed abruptly by the runtime (verified via long running tests of this PR).


src/main/java/com/google/firebase/internal/FirebaseScheduledExecutor.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+public class FirebaseScheduledExecutor extends ScheduledThreadPoolExecutor {
+
+  public FirebaseScheduledExecutor(ThreadFactory threadFactory, String name) {
+    this(threadFactory, name, null);
+  }
+
+  public FirebaseScheduledExecutor(
+      ThreadFactory threadFactory, String name, Thread.UncaughtExceptionHandler handler) {
+    super(1, decorateThreadFactory(threadFactory, name, handler));
```
schmidt-sebastian: I think it would be super duper awesome if we just used `ScheduledThreadPoolExecutor` in our code. It would be much easier to grasp that we are now using an off-the-shelf executor. Maybe you can find a different place for `decorateThreadFactory`?

src/main/java/com/google/firebase/internal/FirebaseScheduledExecutor.java
```diff
@@ -0,0 +1,58 @@
...
+    setRemoveOnCancelPolicy(true);
+  }
+
+  static ThreadFactory decorateThreadFactory(ThreadFactory threadFactory, String name) {
```
schmidt-sebastian: The name of this external method is somewhat generic and as such it is hard to tell what it does. Do you mind finding a better name?

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 - [fixed] Improved error handling in FCM by mapping more server-side
   errors to client-side error codes.
+- [added] `FirebaseAuth`, `'FirebaseMessaging` and `FirebaseInstanceId`
+  interfaces now expose a set of blocking methods. Each operation has
+  blocking an asynchronous versions.
+- [changed] Removed the deprecated `FirebaseCredential` interface.
+- [changed] Removed the deprecated `Task` interface along with the
+  `com.google.firebase.tasks` package.
+- [changed] Dropped support for App Engine Java 7 runtime.
```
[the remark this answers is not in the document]
hiranya911: Done

src/main/java/com/google/firebase/internal/FirebaseScheduledExecutor.java
hiranya911: All of them are access from `DatabaseConfig` and a couple of other places. I managed to make `firebaseApp` package-protected though.

src/main/java/com/google/firebase/database/core/Context.java
```diff
@@ -38,17 +38,18 @@
...
-  private PersistenceManager forcedPersistenceManager;
+
+  EventTarget eventTarget;
+  AuthTokenProvider authTokenProvider;
+  RunLoop runLoop;
+  String persistenceKey;
+  List<String> loggedComponents;
+  Logger.Level logLevel = Logger.Level.INFO;
+  boolean persistenceEnabled;
+  long cacheSize = DEFAULT_CACHE_SIZE;
```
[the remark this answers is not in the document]
hiranya911: All of them are accessed from `DatabaseConfig` and a couple of other places. I managed to make `firebaseApp` package-protected though.

src/main/java/com/google/firebase/internal/FirebaseScheduledExecutor.java
hiranya911: In addition to decorating the `ThreadFactory` this class also customizes the executor a little bit by calling `setRemoveOnCancelPolicy()`. I'd rather not to duplicate that in all the places we use this class. Also having all this in one place gives us an easy way to introduce more customizations in the future if required. WDYT?

src/main/java/com/google/firebase/internal/FirebaseScheduledExecutor.java
hiranya911: Changed to `getThreadFactoryWithName()`

src/main/java/com/google/firebase/internal/FirebaseScheduledExecutor.java
hiranya911: Another way to get around this would be to introduce a new util method somewhere:

```
public static ScheduledThreadPoolExecutor newScheduledExecutor(ThreadFactory factory, String name)
```

We can do all the decorating/customizing in the method. `DefaultRunLoop` will have to find another approach though, since it extends the executor. Thoughts?

hiranya911:
@schmidt-sebastian I've addressed the trivial comments. As for removing `FirebaseScheduledExecutor`, I've left some follow up questions to get you take on. Please take a look.

src/main/java/com/google/firebase/internal/FirebaseScheduledExecutor.java
schmidt-sebastian: I like your last approach better, but I can't think of a good way to replace the inheritance in DefaultRunLoop. We might want to ponder a little more about this, but I don't think we should hold up this PR on it. If anything, we could argue that this derisks this change since it keeps the existing pattern.

src/main/java/com/google/firebase/database/utilities/DefaultRunLoop.java
```diff
@@ -33,43 +33,38 @@
...
-
-          @Override
-          protected void beforeRestart() {
-            if (context != null) {
-              RepoManager.interrupt(context);
+  protected DefaultRunLoop(ThreadFactory threadFactory) {
+    executor = new FirebaseScheduledExecutor(threadFactory, "firebase-database-worker") {
+      @Override
+      protected void afterExecute(Runnable runnable, Throwable throwable) {
+        if (throwable == null && runnable instanceof Future<?>) {
```
schmidt-sebastian: Please call super.afterExecute().

hiranya911: Done

PR #156, opened 2018-03-22
Implementing support for APNS content-mutable field

### LINKED ISSUE #155: FR: Expose mutable_content property in iOS Notifications API
* Operating System version: Any
  * Firebase SDK version: 5.9.0
  * Library version: Latest
  * Firebase Product: notifications

Please provide an API for setting the `mutable_content` property for iOS in Firebase Notifications.
`com.google.firebase.messaging.Aps` class exposes some some properties like `content-available` but not `mutable_content`. 

`mutable_content` property is important for processing notifications in an `UNNotificationServiceExtension`. 

[Apple docs](https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/ModifyingNotifications.html)

hiranya911:
Implements new setter in the `Aps.Builder` class:

```
Builder setContentMutable(boolean)
Builder putCustomField(String, Object)
Builder putAllCustomFields(Map<String,Object>)
```

Resolves #155


src/main/java/com/google/firebase/messaging/Aps.java
```diff
@@ -137,6 +146,41 @@ public Builder setContentAvailable(boolean contentAvailable) {
...
+
+    /**
+     * Puts a custom key-value pair to the aps dictionary.
+     *
+     * @param key A non-null key.
+     * @param value A non-null, json-serializable value.
+     * @return This builder.
+     */
+    public Builder putCustomData(@NonNull String key, @NonNull Object value) {
+      this.customData.put(key, value);
```
schmidt-sebastian: Should we assert here that `mutableContent` is set to true? Please correct me if I am misunderstanding how this works.

src/main/java/com/google/firebase/messaging/Aps.java
```diff
@@ -137,6 +146,41 @@ public Builder setContentAvailable(boolean contentAvailable) {
...
+      return this;
+    }
+
+    /**
+     * Puts all the key-value pairs in the specified map to the aps dictionary.
+     *
+     * @param fields A non-null map. Map must not contain null keys or values.
+     * @return This builder.
+     */
+    public Builder putAllCustomData(@NonNull Map<String, Object> fields) {
```
schmidt-sebastian: Java 8 allows you to do `@NonNull Map<@NonNull String, @NonNull Object>`

src/main/java/com/google/firebase/messaging/Aps.java
```diff
@@ -159,6 +203,13 @@ public Builder setThreadId(String threadId) {
       return this;
     }
 
+    /**
+     * Builds a new {@link Aps} instance from the fields set on this builder.
+     *
+     * @return A non-null {@link Aps}.
+     * @throws IllegalArgumentException If the alert is specified both as an object and a string.
+     *     Or if the same field is set both using a setter method, and as a custom field.
```
schmidt-sebastian: Is this last part of this enforced somewhere?

src/main/java/com/google/firebase/messaging/Aps.java
hiranya911: We are still on Java 7. Only GAE support has been bumped to Java 8.

src/main/java/com/google/firebase/messaging/Aps.java
hiranya911: Well in that case it has to be set as `mutable-content: 1`. Since we are exposing a typed setter to configure this (`setMutabeContent()`), my preference is to not validate it here. And if a developer tries to set some nonsense value, they will get a backend error.

src/main/java/com/google/firebase/messaging/Aps.java
hiranya911: `ImmutableMap.Builder` checks for this internally. We have a test case to verify.

schmidt-sebastian: Got it. I saw the test cases but I was wondering if maybe it caught an unrelated `IllegalArgumentException`.

DJFliX:
I thought this was going to be in the 5.10.0 release but it doesn't seem to be there. Can you give me an estimate on the next release (that hopefully includes these changes 😀)

PR #157, opened 2018-03-23
Handling some WebSocket connection errors gracefully

hiranya911:
There's a small delay between `channelInactive()` low-level event, and us calling `channel.close()`. The former executes on a Netty worker thread, while the latter is on the runloop. An outgoing `send()` may sneak through this window to cause a hard assertion failure. 

The old tubesock implementation handled this type of race conditions more gracefully, by simply logging the error and closing the connection for good. This PR essentially implements the same semantics on top of Netty.

b/74831164

### DIFF
--- a/src/main/java/com/google/firebase/database/connection/WebsocketConnection.java
+++ b/src/main/java/com/google/firebase/database/connection/WebsocketConnection.java
@@ -324,8 +324,7 @@ class WebsocketConnection {
 
     @Override
     public void onError(final Throwable e) {
-      if (e instanceof EOFException
-          || (e.getCause() != null && e.getCause() instanceof EOFException)) {
+      if (e instanceof EOFException || e.getCause() instanceof EOFException) {
         logger.debug("WebSocket reached EOF", e);
       } else {
         logger.error("WebSocket error", e);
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/firebase/database/connection/WebsocketConnection.java
```diff
@@ -325,9 +325,9 @@ public void run() {
     @Override
     public void onError(final Throwable e) {
       if (e.getCause() != null && e.getCause() instanceof EOFException) {
-        logger.error("WebSocket reached EOF", e);
+        logger.debug("WebSocket reached EOF", e);
       } else {
-        logger.error("WebSocket error", e);
+        logger.debug("WebSocket error", e);
```
schmidt-sebastian: Do you think it might make sense to still log all other errors at 'error' level? You should really only suppress the IllegalStateException you created in NettyWebsocketClient.

hiranya911: Done

src/main/java/com/google/firebase/database/connection/WebsocketConnection.java
```diff
@@ -324,8 +324,9 @@ public void run() {
 
     @Override
     public void onError(final Throwable e) {
-      if (e.getCause() != null && e.getCause() instanceof EOFException) {
-        logger.error("WebSocket reached EOF", e);
+      if (e instanceof EOFException
+          || (e.getCause() != null && e.getCause() instanceof EOFException)) {
```
schmidt-sebastian: You don't need to perform a null check before calling instanceof.

hiranya911: Done

PR #159, opened 2018-03-29
Reading FCM error code from the details section

hiranya911:
Porting error handling improvement from https://github.com/firebase/firebase-admin-node/pull/246

Reading error code from the `details` section of the HTTP response when available.

### DIFF
--- a/src/test/java/com/google/firebase/messaging/FirebaseMessagingTest.java
+++ b/src/test/java/com/google/firebase/messaging/FirebaseMessagingTest.java
@@ -97,7 +97,7 @@ public class FirebaseMessagingTest {
   }
 
   @Test
-  public void testNullMessage() throws Exception {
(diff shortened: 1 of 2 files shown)


src/test/java/com/google/firebase/messaging/FirebaseMessagingTest.java
```diff
@@ -97,7 +97,7 @@ public void testNoProjectId() {
   }
 
   @Test
-  public void testNullMessage() throws Exception {
```
avishalom: did you remove the throws on purpose? (I'm good either way, but the next test has it)

hiranya911: Yes, the compiler flagged this as redundant. Other tests call methods that throw checked exceptions.

PR #167, opened 2018-04-20
Add handling of null response content

### LINKED ISSUE #166: Error Response with zero content throws NullPointerException
Hello, 

When a response comes back with an erred status code com.google.firebase.messaging.FirebaseMessaging#handleSendHttpError attempts to read the body, however if the content is null, the jsonFactory throws a NullPointerException, which is not converted into a FirebaseMessagingException. This ends up hiding the http response code.

In the tests, there is a test for testSendError, but there is no test that uses .setZeroContent() which will expose the problem.

To reproduce, send a mock response with an error status code and no content.

funkybookends:
If an error response comes back with no content, the json factory throws a
NullPointerException when it attempts to read the response, which hides the
response code from the client.

fixes #166

### DIFF
--- a/src/main/java/com/google/firebase/messaging/FirebaseMessaging.java
+++ b/src/main/java/com/google/firebase/messaging/FirebaseMessaging.java
@@ -253,13 +253,14 @@ public class FirebaseMessaging {
 
   private void handleSendHttpError(HttpResponseException e) throws FirebaseMessagingException {
     MessagingServiceErrorResponse response = new MessagingServiceErrorResponse();
-    try {
-      JsonParser parser = jsonFactory.createJsonParser(e.getContent());
-      parser.parseAndClose(response);
-    } catch (IOException | NullPointerException ignored) {
-      // ignored
+    if (e.getContent() != null) {
+      try {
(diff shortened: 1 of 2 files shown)


funkybookends:
I signed it!

Sorry for not following templates to the letter, but it's a pretty simple change.

src/main/java/com/google/firebase/messaging/FirebaseMessaging.java
```diff
@@ -256,7 +256,7 @@ private void handleSendHttpError(HttpResponseException e) throws FirebaseMessagi
     try {
       JsonParser parser = jsonFactory.createJsonParser(e.getContent());
       parser.parseAndClose(response);
-    } catch (IOException ignored) {
+    } catch (IOException | NullPointerException ignored) {
```
hiranya911: Handling NPEs is almost never a good idea. Can this be avoided by checking `Strings.isNullOrEmpty(e.getContent())`, before calling the parser? That makes the intent more clear as well.

funkybookends: I've addressed this now.

PR #172, opened 2018-05-07
Staged Release 6.0.0

hiranya911:


pom.xml
```diff
@@ -19,7 +19,7 @@
 
     <groupId>com.google.firebase</groupId>
     <artifactId>firebase-admin</artifactId>
-    <version>5.11.1-SNAPSHOT</version>
+    <version>6.0.1-SNAPSHOT</version>
```
schmidt-sebastian: 6..0.0?

hiranya911: 6.0.0 is the version that's going out. Our release tools build the release version, and then update the pom.xml to the next development version -- 6.0.1 in this case.

PR #181, opened 2018-05-29
Snippets for user import functionality

hiranya911:
Snippets to appear on https://firebase.google.com/docs/auth/admin/import-users

### DIFF
--- a/src/test/java/com/google/firebase/snippets/FirebaseAuthSnippets.java
+++ b/src/test/java/com/google/firebase/snippets/FirebaseAuthSnippets.java
@@ -413,6 +413,7 @@ public class FirebaseAuthSnippets {
 
   public void importUsers() {
     // [START build_user_list]
+    // Up to 1000 users can be imported at once.
     List<ImportUserRecord> users = new ArrayList<>();
     users.add(ImportUserRecord.builder()
         .setUid("uid1")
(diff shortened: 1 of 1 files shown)


src/test/java/com/google/firebase/snippets/FirebaseAuthSnippets.java
```diff
@@ -396,4 +410,191 @@ public Response clearSessionCookieAndRevoke(@CookieParam("session") Cookie cooki
...
+    // [END build_user_list]
+
+    // [START import_users]
+    UserImportOptions options = UserImportOptions.withHash(
+        HmacSha256.builder()
+            .setKey("secretKey".getBytes())
+            .build());
+    try {
+      UserImportResult result = FirebaseAuth.getInstance().importUsers(users, options);
+      // The number of successful imports is determined via: result.getSuccessCount().
```
schmidt-sebastian: I'm gonna suggest that you turn this comments into code. 

```
 UserImportResult result = FirebaseAuth.getInstance().importUsers(users, options);
 int successCount = result.getSuccessCount();
 int failureCount = result. getFailureCount();

System.out.println("...");

for (ErrorInfo indexedError : result.getErrors()) {
 int index = indexedError.getIndex()
 String reason = indexedError.getReason());
 System.out.println("...");
}
```

It probably only make sense to do this for this one snippet. I think you should leave the others as is.

hiranya911: Good idea. Done.

PR #189, opened 2018-06-26
Reading project ID from GOOGLE_CLOUD_PROJECT

hiranya911:
`GCLOUD_PROJECT` is no longer supported in some managed Google runtimes. `GOOGLE_CLOUD_PROJECT` is the new recommended environment variable.


hiranya911:
Done. Added the suggested test case.

PR #193, opened 2018-07-16
Added WebpushNotification.Builder API with additional Webpush parameters

hiranya911:
FCM Webpush API now supports additional parameters as defined in https://developer.mozilla.org/en-US/docs/Web/API/notification/Notification

Related to https://github.com/firebase/firebase-admin-node/pull/305


src/main/java/com/google/firebase/messaging/WebpushNotification.java
```diff
@@ -54,8 +59,347 @@ public WebpushNotification(String title, String body) {
...
+     */
+    public Builder setBody(String body) {
+      this.body = body;
+      return this;
+    }
+
+    /**
+     * Sets any arbitrary data that should be associated with the notification.
+     *
+     * @param data A json-serializable object.
```
schmidt-sebastian: s/json/JSON/ :)

src/main/java/com/google/firebase/messaging/WebpushConfig.java
```diff
@@ -35,12 +35,12 @@
   private final Map<String, String> data;
 
   @Key("notification")
-  private final WebpushNotification notification;
+  private final Map<String, Object> notification;
```
schmidt-sebastian: Sorry, I don't quite follow where this private fields are ever read. Can you elaborate?

I am asking this because I liked using `WebpushNotification` here better but I can see reason why the Map would be more appropriate.

hiranya911: All these get serialized into JSON. We now need a Map because the notification now allows arbitrary key-value pairs (via `putCustomData()`).

src/main/java/com/google/firebase/messaging/WebpushNotification.java
hiranya911: Done

Issue #204, opened 2018-09-11
FR: Expose the FirestoreOptions#setTimestampsInSnapshotsEnabled(boolean) method


RohanTalip:
firebase-admin-java does (6.4.0) not yet expose the FirestoreOptions#setTimestampsInSnapshotsEnabled(boolean) method :
https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOptions.java#L203

I would like to keep our POJOs similar between our Android code base and our Java code that uses Firebase Admin (for Firestore).

I have a pull request that is almost ready which will solve this.  I'm just filing an issue as per the directions on the pull request template.

PR #203, opened 2018-08-30
Upgrading Google Cloud Dependencies

hiranya911:
Test case changes are due to https://github.com/google/google-api-java-client/issues/1151

PR #205, opened 2018-09-11
Upgraded google-cloud-firestore to the latest version (0.61.0-beta) …

RohanTalip:
…and added a FirebaseOptions#setFireStoreTimestampsInSnapshotsEnabled(boolean) method.

The FirestoreOptions#setTimestampsInSnapshotsEnabled(boolean) method is then called in the FirestoreClient constructor with the value passed to the FirebaseOptions#setFireStoreTimestampsInSnapshotsEnabled(boolean) method.

This allows for firebase-admin-java users to adapt to the Timestamps being returned in DocumentSnapshots as added in https://github.com/GoogleCloudPlatform/google-cloud-java/pull/3317

com.google.api:gax was added as a dependency to the pom.xml file as there were "java.lang.NoClassDefFoundError: com/google/api/gax/rpc/ClientStream" errors when running "mvn test" without this dependency.

A couple of tests were also added to the FirestoreClientTest class.

See also issue #204.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -399,6 +399,11 @@
             <version>1.2.0</version>
         </dependency>
         <dependency>
+            <groupId>com.google.api</groupId>
+            <artifactId>gax</artifactId>
+            <version>1.30.0</version>
+        </dependency>
+        <dependency>
             <groupId>com.google.auth</groupId>
             <artifactId>google-auth-library-oauth2-http</artifactId>
             <version>0.8.0</version>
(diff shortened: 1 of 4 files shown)


hiranya911:
See #203

RohanTalip:
> There is already some work to support this. Plan is to support passing an entire `FirestoreOptions` object into `FirebaseOptions`.

Great!  I see your recent commits on #203 .  I wasn't sure if exposing the whole `FirestoreOptions` object was going to be accepted in the code review, so I tried just exposing only what I needed.

Thanks for working on this @hiranya911 !

PR #211, opened 2018-10-29
Add App-scoped API For Firebase Project management

xiangkangjw:
App-scoped actions for Firebase apps within Firebase projects. Uses Firebase Management REST APIs (https://firebase.google.com/docs/projects/api/reference/rest/)


schmidt-sebastian:
This PR only contains previously reviewed changes.

weixifan:
I fixed some iOS vs Android mistakes in the Javadoc for various methods in `FirebaseProjectManagement`. I will assign a reviewer in case there are comments about that change.

weixifan:
I also just adjusted the error codes list. Both Android certificates and Apps are handled by the same error codes map, so I changed the term "App" to simply "resource." Please take a look.

schmidt-sebastian:
Still LGTM

weixifan:
Unfortunately I discovered more things to fix. :(

src/test/java/com/google/firebase/projectmanagement/FirebaseProjectManagementServiceImplTest.java
```diff
@@ -0,0 +1,911 @@
...
+    GET,
+    POST,
+    DELETE
+  }
+
+  /**
+   * Can be used to intercept multiple HTTP requests and responses made by the SDK during tests.
+   */
+  private class MultiRequestTestResponseInterceptor implements HttpResponseInterceptor {
+    private List<HttpResponse> responsesList = new ArrayList<HttpResponse>();
```
hiranya911: private final

hiranya911: Use diamond syntax (we are on Java 7 where it's allowed):

```
new ArrayList<>();
```

src/test/java/com/google/firebase/projectmanagement/FirebaseProjectManagementServiceImplTest.java
```diff
@@ -0,0 +1,911 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
```
hiranya911: There's a ton of code duplication in this file. Can we do something about it? One simple solution is to combine some tests into a single method as follows (you might have to change how mock responses are set up):

```
  @Test
  public void getIosApp() throws Exception {
    String expectedUrl = String.format(
        "%s/v1beta1/projects/-/iosApps/%s", FIREBASE_PROJECT_MANAGEMENT_URL, IOS_APP_ID);
    firstRpcResponse.setContent(String.format(GET_IOS_RESPONSE, DISPLAY_NAME));
    serviceImpl = initServiceImpl(firstRpcResponse, interceptor);
    IosAppMetadata iosAppMetadata = serviceImpl.getIosApp(IOS_APP_ID);
     checkRequestHeader(expectedUrl, HttpMethod.GET);
     assertEquals(iosAppMetadata, IOS_APP_METADATA);
  
     iosAppMetadata = serviceImpl.getIosAppAsync(IOS_APP_ID).get();
     checkRequestHeader(expectedUrl, HttpMethod.GET);
     assertEquals(iosAppMetadata, IOS_APP_METADATA);
  }
```

src/main/java/com/google/firebase/projectmanagement/FirebaseProjectManagementServiceImpl.java
```diff
@@ -0,0 +1,909 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+class FirebaseProjectManagementServiceImpl implements AndroidAppService, IosAppService {
+  @VisibleForTesting static final String FIREBASE_PROJECT_MANAGEMENT_URL =
```
hiranya911: Add newline

src/main/java/com/google/firebase/projectmanagement/FirebaseProjectManagementServiceImpl.java
```diff
@@ -0,0 +1,909 @@
...
+      new CreateAndroidAppFromAppIdFunction();
+  private final CreateIosAppFromAppIdFunction createIosAppFromAppIdFunction =
+      new CreateIosAppFromAppIdFunction();
+
+  private HttpResponseInterceptor interceptor;
+
+  FirebaseProjectManagementServiceImpl(FirebaseApp app) {
+    this.app = app;
+
+    JsonFactory jsonFactory = app.getOptions().getJsonFactory();
```
hiranya911: Assign directly to `this.jsonFactory`

src/main/java/com/google/firebase/projectmanagement/FirebaseProjectManagementServiceImpl.java
```diff
@@ -0,0 +1,909 @@
...
+    this.jsonFactory = jsonFactory;
+
+    HttpTransport httpTransport = app.getOptions().getHttpTransport();
+    this.requestFactory =
+        httpTransport.createRequestFactory(new FirebaseRequestInitializer(app));
+
+    this.scheduledExecutor = new FirebaseScheduledExecutor(
```
hiranya911: Please remove this. No service should initialize its own thread pools in the SDK.

src/main/java/com/google/firebase/projectmanagement/FirebaseProjectManagementServiceImpl.java
```diff
@@ -0,0 +1,909 @@
...
+  private CallableOperation<AndroidAppMetadata, FirebaseProjectManagementException> getAndroidAppOp(
+      final String appId) {
+    return new CallableOperation<AndroidAppMetadata, FirebaseProjectManagementException>() {
+      @Override
+      protected AndroidAppMetadata execute() throws FirebaseProjectManagementException {
+        String url = String.format(
+            "%s/v1beta1/projects/-/androidApps/%s", FIREBASE_PROJECT_MANAGEMENT_URL, appId);
+        AndroidAppResponse parsedResponse = new AndroidAppResponse();
+        makeGetRequest(url, parsedResponse, appId, "App ID");
+        return new AndroidAppMetadata(
```
hiranya911: I think AndroidAppMetadata should just provide a constructor that accepts an AndroidAppResponse as the only argument.

src/main/java/com/google/firebase/projectmanagement/FirebaseProjectManagementServiceImpl.java
```diff
@@ -0,0 +1,909 @@
...
+  private CallableOperation<IosAppMetadata, FirebaseProjectManagementException> getIosAppOp(
+      final String appId) {
+    return new CallableOperation<IosAppMetadata, FirebaseProjectManagementException>() {
+      @Override
+      protected IosAppMetadata execute() throws FirebaseProjectManagementException {
+        String url = String.format(
+            "%s/v1beta1/projects/-/iosApps/%s", FIREBASE_PROJECT_MANAGEMENT_URL, appId);
+        IosAppResponse parsedResponse = new IosAppResponse();
+        makeGetRequest(url, parsedResponse, appId, "App ID");
+        return new IosAppMetadata(
```
hiranya911: Same here. Provide a constructor that takes IosAppResponse as the only argument.

src/main/java/com/google/firebase/projectmanagement/FirebaseProjectManagementServiceImpl.java
```diff
@@ -0,0 +1,909 @@
...
+        String.format(
+            "%s \"%s\": %s", requestIdentifierDescription, requestIdentifier, messageBody),
+        cause);
+  }
+
+  /* Helper types. */
+
+  private interface CreateAppFromAppIdFunction<T> extends ApiFunction<String, T> {}
+
+  private class CreateAndroidAppFromAppIdFunction
```
hiranya911: Make these static classes. Pass the service impl instance as a constructor arg.

src/main/java/com/google/firebase/projectmanagement/FirebaseProjectManagementServiceImpl.java
```diff
@@ -0,0 +1,909 @@
...
+      @Nullable Exception cause) {
+    return new FirebaseProjectManagementException(
+        String.format(
+            "%s \"%s\": %s", requestIdentifierDescription, requestIdentifier, messageBody),
+        cause);
+  }
+
+  /* Helper types. */
+
+  private interface CreateAppFromAppIdFunction<T> extends ApiFunction<String, T> {}
```
hiranya911: Lets remove this. Implementing types can directly implement ApiFunction interface.

src/main/java/com/google/firebase/projectmanagement/FirebaseProjectManagementServiceImpl.java
```diff
@@ -0,0 +1,909 @@
...
+    }
+
+    /**
+     * Returns an {@link ApiFuture} that will eventually contain the App ID of the new created App,
+     * or an exception if an error occurred during polling.
+     */
+    @Override
+    public ApiFuture<String> apply(String operationName) throws FirebaseProjectManagementException {
+      SettableApiFuture<String> settableFuture = SettableApiFuture.<String>create();
+      scheduledExecutor.schedule(
```
hiranya911: Expose the FirebaseApp.schedule() method via a trampoline, and use it here.

src/main/java/com/google/firebase/projectmanagement/FirebaseProjectManagementServiceImpl.java
```diff
@@ -0,0 +1,909 @@
...
+      this.operationName = operationName;
+      this.projectId = projectId;
+      this.settableFuture = settableFuture;
+    }
+
+    @Override
+    public void run() {
+      String url = String.format("%s/v1/%s", FIREBASE_PROJECT_MANAGEMENT_URL, operationName);
+      OperationResponse operationResponseInstance;
+      operationResponseInstance = new OperationResponse();
```
hiranya911: Combine with the previous line.

src/main/java/com/google/firebase/projectmanagement/FirebaseProjectManagementServiceImpl.java
```diff
@@ -0,0 +1,909 @@
...
+        "Bundle ID",
+        "Unable to create iOS App");
+  }
+
+  @Override
+  public ApiFuture<IosApp> createIosAppAsync(
+      String projectId, String bundleId, String displayName) {
+    checkArgument(!Strings.isNullOrEmpty(bundleId), "bundle ID must not be null or empty");
+    return
+        transform(
```
hiranya911: Please qualify these as `ApiFutures.transform()`. In general all static references should be qualified. Guava preconditions are the only acceptable exception.

src/main/java/com/google/firebase/projectmanagement/FirebaseProjectManagementServiceImpl.java
```diff
@@ -0,0 +1,909 @@
...
+    };
+  }
+
+  /**
+   * An {@link ApiAsyncFunction} that transforms a Long Running Operation name to an {@link IosApp}
+   * or an {@link AndroidApp} instance by repeatedly polling the server (with exponential backoff)
+   * until the App is created successfully, or until the number of poll attempts exceeds the maximum
+   * allowed.
+   */
+  private class WaitOperationFunction implements ApiAsyncFunction<String, String> {
```
hiranya911: static

src/main/java/com/google/firebase/projectmanagement/FirebaseProjectManagementServiceImpl.java
```diff
@@ -0,0 +1,909 @@
...
+                  projectId,
+                  settableFuture),
+              delay,
+              TimeUnit.MILLISECONDS);
+        }
+        return;
+      }
+      // The Long Running Operation API guarantees that when done == true, exactly one of 'response'
+      // or 'error' is set.
+      if (operationResponseInstance.response == null
```
hiranya911: Null check is redundant. You've already dereferenced this once earlier.

src/main/java/com/google/firebase/projectmanagement/FirebaseProjectManagementServiceImpl.java
```diff
@@ -0,0 +1,909 @@
...
+    @Key("shaHash")
+    private String shaHash;
+
+    @Key("certType")
+    private String certType;
+  }
+
+  /* Helper methods. */
+
+  private <T> void makeGetRequest(
```
hiranya911: Consider introducing a new helper class. May be `HttpHelper.java`?

src/main/java/com/google/firebase/projectmanagement/FirebaseProjectManagementServiceImpl.java
```diff
@@ -0,0 +1,909 @@
...
+    @Key("nextPageToken")
+    private String nextPageToken;
+  }
+
+  /* createAndroidApp, createIosApp */
+
+  @Override
+  public AndroidApp createAndroidApp(String projectId, String packageName, String displayName)
+      throws FirebaseProjectManagementException {
+    return blockOnFutureAndHandleExceptions(
```
hiranya911: Ideally these blocking methods should not submit work to other threads. Can we get it to run everything on the calling thread?

```
createAndroidAppOp(projectId, packageName, displayName).call();
while (true) {
    if (isAppCreated()) {
        return result;
    }
    Thread.sleep(exponentiallyIncreasingDelay);
}
```

src/main/java/com/google/firebase/projectmanagement/FirebaseProjectManagementServiceImpl.java
```diff
@@ -0,0 +1,909 @@
...
+      }
+      if (!operationResponseInstance.done) {
+        if (numberOfPreviousPolls + 1 >= MAXIMUM_POLLING_ATTEMPTS) {
+          settableFuture.setException(createFirebaseProjectManagementException(
+              projectId,
+              "Project ID",
+              "Unable to create App: deadline exceeded.",
+              /* cause= */ null));
+        } else {
+          long delay = (long) (
```
hiranya911: The exponential growth of the delay should be capped at some maximum. I think the Node.js implementation of this API uses a max of 10 seconds.

src/main/java/com/google/firebase/projectmanagement/AndroidApp.java
```diff
@@ -0,0 +1,191 @@
...
+ * An instance of this class is a reference to an Android App within a Firebase Project; it can be
+ * used to query detailed information about the App, modify the display name of the App, or download
+ * the configuration file for the App.
+ */
+public class AndroidApp {
+
+  private final AndroidAppService androidAppService;
+  private final String appId;
+
+  private AndroidApp(String appId, AndroidAppService androidAppService) {
```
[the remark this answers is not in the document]
weixifan: An example of a thing that can be done in a factory method that you cannot do in a constructor is use a different method name for two different factory methods. The reason this is needed is we plan to have another constructor in our upcoming proposal that is also based on a String (the fully qualified resource name of the app, rather than the App ID). This is impossible with a constructor since you cannot overload the constructor with exactly the same signature.

src/main/java/com/google/firebase/projectmanagement/ShaCertificate.java
```diff
@@ -0,0 +1,123 @@
...
+    } else if (SHA256_PATTERN.matcher(shaHash).matches()) {
+      return ShaCertificateType.SHA_256;
+    }
+    throw new IllegalArgumentException("Invalid SHA hash, it is neither SHA-1 nor SHA-256.");
+  }
+
+  /**
+   * Returns the fully qualified resource name of this SHA certificate.
+   */
+  String getName() {
```
weixifan: These should be public. Please see the API proposal document for the precise public API.

src/main/java/com/google/firebase/projectmanagement/FirebaseProjectManagementServiceImpl.java
hiranya911: No, not a project-wide helper class. Just a helper for this package/class. Basically get all the http cruft out of this class.

src/main/java/com/google/firebase/projectmanagement/FirebaseProjectManagementServiceImpl.java
hiranya911: Spawning threads is expensive (and even restricted) in some environments. The old GAE runtime is a good example, where there are strict caps on how many threads can be spawned by an app. Therefore we have designed our blocking methods to never spawn additional threads.

src/main/java/com/google/firebase/projectmanagement/ShaCertificateType.java
```diff
@@ -19,6 +19,8 @@
  * Enum denoting types of SHA certificates currently supported by Firebase.
  */
 public enum ShaCertificateType {
+  /** Certificate generated by SHA-1 hashing algorithm. */
   SHA_1,
```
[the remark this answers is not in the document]
weixifan: According to our own official Public API documentation, we are using the underscore in our constants. https://firebase.google.com/docs/projects/api/reference/rest/v1beta1/projects.androidApps.sha#ShaCertificateType

weixifan:
Thanks for the review. Aside from the asynchronous polling question that remains, we have another small issue: I noticed that for the `AndroidApp.deleteShaCertificate()` method, the caller must pass in a `ShaCertificate`, constructed with the `name` field set. This means we either force the caller to only be able to pass in a `ShaCertificate` returned by the `getShaCertificates()` method, or we should expose a public method allowing them to create a `ShaCertificate` with a caller-supplied `name` field. I think we should allow the user to delete Android SHA certificates without having to call `getShaCertificates` first, if they happen to know the full resource name of the certificate they want to delete prior to calling it, but this isn't that critical. Please let me know what you think.

src/main/java/com/google/firebase/projectmanagement/ShaCertificateType.java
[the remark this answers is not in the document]
hiranya911: In plain English it's normally written as SHA1 or as SHA-1 (see https://tools.ietf.org/html/rfc3174). I kind of think `SHA1` would have been better for our API, but it's not a big deal (especially since this is an enum value and not a class name).

src/main/java/com/google/firebase/ImplFirebaseTrampolines.java
```diff
@@ -67,6 +68,10 @@ public static ThreadFactory getThreadFactory(@NonNull FirebaseApp app) {
     return app.getThreadFactory();
   }
 
+  public static ScheduledExecutorService getScheduledExecutorService(@NonNull FirebaseApp app) {
```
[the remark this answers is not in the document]
hiranya911: I cannot seem to find any comment related to this. Did you forget to submit a comment?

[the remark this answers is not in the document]
weixifan: Sorry, I was referring to the comment in line 259 of `FirebaseProjectManagementServiceImpl.java`.

src/main/java/com/google/firebase/projectmanagement/FirebaseProjectManagementServiceImpl.java
hiranya911: Individual services should not spawn new threads. None of the other services (auth, fcm, iid) do, and I don't see any reason why this service should be different. Also methods like `createAndroidApp` are unlikely to be called at high volumes for performance to ever become a problem. Therefore please implement the simple/naive solution with blocking threads for now. If it ever comes to that, we can make an exception for this service and allow it to spawn threads.

weixifan: Thanks, good point. In fact, after thinking about this some more, I think more generally the only surface through which `ThreadManager` should affect threading is the asynchronous API, rather than the blocking API, so that clients that wish to have programmatic threading control should exclusively use our asynchronous API. It is indeed unnatural to force the client to choose a particular `ThreadManager` just so that their blocking calls do not spawn threads.

For this blocking version of `createAndroidApp`, I will use your original suggestion of blocking using just `Thread.sleep(...)` (will push shortly). As for my proposed asynchronous `createAndroidAppAsync` implementation, please let me know if you have any further concerns.

hiranya911: My recommendation would be to keep things absolutely simple. Once you have a blocking implementation based on `Thread.sleep()`, the async version of the method can just run the blocking implementation on a thread provisioned by the `ThreadManager` -- i.e. the same way the other async methods work. We can implement something more sophisticated later, if the need ever arises.

src/main/java/com/google/firebase/projectmanagement/FirebaseProjectManagementServiceImpl.java
weixifan: This was moved to `HttpHelper`. Resolving.

src/main/java/com/google/firebase/projectmanagement/FirebaseProjectManagementServiceImpl.java
weixifan: As we discussed, I have re-implemented the blocking API with `Thread.sleep()`, and instead of exposing the entire `ScheduledExecutorService` from the `FirebaseApp`, I have instead exposed three separate methods `transform`, `transformAsync`, and `schedule`.

src/main/java/com/google/firebase/projectmanagement/FirebaseProjectManagementServiceImpl.java
weixifan: Done.

src/main/java/com/google/firebase/ImplFirebaseTrampolines.java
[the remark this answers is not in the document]
weixifan: This is resolved as per our discussion in the above comments.

src/main/java/com/google/firebase/projectmanagement/FirebaseProjectManagementServiceImpl.java
weixifan: I will consolidate this thread with the one above. Resolving this one.

weixifan:
Thank you for the reviews. I have added the client version header as we discussed.

PR #227, opened 2018-11-29
Closing Firestore instance on FirebaseApp.delete()

hiranya911:
Ensures proper clean up of Firestore instance. Helps avoid warnings such as the following during tests:

```
Nov 28, 2018 4:46:23 PM io.grpc.internal.ManagedChannelOrphanWrapper$ManagedChannelReference cleanQueue
SEVERE: *~*~*~ Channel ManagedChannelImpl{logId=1, target=firestore.googleapis.com:443} was not shutdown properly!!! ~*~*~*
    Make sure to call shutdown()/shutdownNow() and wait until awaitTermination() returns true.
java.lang.RuntimeException: ManagedChannel allocation site
	at io.grpc.internal.ManagedChannelOrphanWrapper$ManagedChannelReference.<init>(ManagedChannelOrphanWrapper.java:103)
```

PR #228, opened 2018-11-30
Close firestore at end of relevant tests

rsgowman:
Avoids the grpc ManagedChannel complaining about not being shutdown.

http://b/120096974


hiranya911:
This is fixed by #227

PR #233, opened 2018-12-17
Implemented CriticalSound API for FCM

hiranya911:
go/apns-critical-sound

Implemented the `CriticalSound` API and the corresponding `Aps.Builder.setSound()` method.

### DIFF
--- a/src/main/java/com/google/firebase/messaging/Aps.java
+++ b/src/main/java/com/google/firebase/messaging/Aps.java
@@ -35,6 +35,8 @@ public class Aps {
   private Aps(Builder builder) {
     checkArgument(Strings.isNullOrEmpty(builder.alertString) || (builder.alert == null),
         "Multiple alert specifications (string and ApsAlert) found.");
+    checkArgument(Strings.isNullOrEmpty(builder.sound) || (builder.criticalSound == null),
+        "Multiple sound specifications (sound and CriticalSound) found.");
     ImmutableMap.Builder<String, Object> fields = ImmutableMap.builder();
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/firebase/messaging/Aps.java
```diff
@@ -135,6 +141,17 @@ public Builder setSound(String sound) {
       return this;
     }
 
+    /**
+     * Sets the critical alert sound to be played with the message.
+     *
+     * @param sound Sound file name or {@code "default"}.
```
bklimt: Update this.

src/main/java/com/google/firebase/messaging/CriticalSound.java
```diff
@@ -0,0 +1,115 @@
...
+    public Builder setName(String name) {
+      this.name = name;
+      return this;
+    }
+
+    /**
+     * The volume for the critical alert's sound. Must be a value between 0.0 (silent) and 1.0
+     * (full volume).
+     *
+     * @param volume A volume between 0.0 and 1.0.
```
bklimt: inclusive

src/main/java/com/google/firebase/messaging/Aps.java
hiranya911: Done

src/main/java/com/google/firebase/messaging/CriticalSound.java
hiranya911: Done

Issue #113, opened 2017-12-09
Remove Dependency on org.json:json Artifact


hiranya911:
Several users have reported issues with our use of this library (#112 being the latest). A number of  frameworks seem to be using various conflicting versions of this library, which causes errors when trying to use Admin SDK in those frameworks. To make matters worse, there's also another [different artifact](https://github.com/tdunning/open-json) which has the exact same package structure! It is possible to overcome this type of errors by carefully analyzing the dependency tree, and removing any conflicting artifacts. But it still results in a poor first-time user experience. Therefore we should consider using another, less ambiguous JSON library like gson in Admin SDK.

This won't be easy, since a good part of our RTDB JSON support is tied to this library at the moment. Perhaps we can start migrating non-RTDB code first.

PR #242, opened 2019-01-09
Remove Dependency on org.json:json Artifact (#113), Java 11 build problems (#243)

matejsp:
fixing #113

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
-- [added] java 11 build support
+- [changed] Removed org.json dependency and replaced with com.google.code.gson.
+- [changed] Upgraded Mockito dependency, and fixed the build on Java 11.
 
 # v6.7.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
--
+- [added] java 11 build support
 
 # v6.7.0
 
--- a/src/main/java/com/google/firebase/database/util/JsonMapper.java
+++ b/src/main/java/com/google/firebase/database/util/JsonMapper.java
@@ -16,19 +16,17 @@
 
 package com.google.firebase.database.util;
 
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.google.gson.JsonSyntaxException;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonToken;
 import java.io.IOException;
+import java.io.StringReader;
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-
-import org.json.JSONArray;
-import org.json.JSONException;
-import org.json.JSONObject;
-import org.json.JSONStringer;
-import org.json.JSONTokener;
 
 /**
  * Helper class to convert from/to JSON strings. TODO: This class should ideally not live in
@@ -37,102 +35,82 @@ import org.json.JSONTokener;
  */
 public class JsonMapper {
 
-  public static String serializeJson(Map<String, Object> object) throws IOException {
-    return serializeJsonValue(object);
-  }
+  private static final Gson GSON = new GsonBuilder().serializeNulls().create();
 
-  @SuppressWarnings("unchecked")
-  public static String serializeJsonValue(Object object) throws IOException {
-    if (object == null) {
-      return "null";
-    } else if (object instanceof String) {
-      return JSONObject.quote((String) object);
-    } else if (object instanceof Number) {
-      try {
-        return JSONObject.numberToString((Number) object);
-      } catch (JSONException e) {
-        throw new IOException("Could not serialize number", e);
-      }
-    } else if (object instanceof Boolean) {
-      return ((Boolean) object) ? "true" : "false";
-    } else {
-      try {
-        JSONStringer stringer = new JSONStringer();
-        serializeJsonValue(object, stringer);
-        return stringer.toString();
-      } catch (JSONException e) {
-        throw new IOException("Failed to serialize JSON", e);
-      }
-    }
-  }
-
-  private static void serializeJsonValue(Object object, JSONStringer stringer)
-      throws IOException, JSONException {
-    if (object instanceof Map) {
-      stringer.object();
-      @SuppressWarnings("unchecked")
-      Map<String, Object> map = (Map<String, Object>) object;
-      for (Map.Entry<String, Object> entry : map.entrySet()) {
-        stringer.key(entry.getKey());
-        serializeJsonValue(entry.getValue(), stringer);
-      }
-      stringer.endObject();
-    } else if (object instanceof Collection) {
-      Collection<?> collection = (Collection<?>) object;
-      stringer.array();
-      for (Object entry : collection) {
-        serializeJsonValue(entry, stringer);
-      }
-      stringer.endArray();
-    } else {
-      stringer.value(object);
+  public static String serializeJson(Object object) throws IOException {
+    try {
+      return GSON.toJson(object);
+    } catch (JsonSyntaxException e) {
+      throw new IOException(e);
     }
   }
 
   public static Map<String, Object> parseJson(String json) throws IOException {
     try {
-      return unwrapJsonObject(new JSONObject(json));
-    } catch (JSONException e) {
+      JsonReader jsonReader = new JsonReader(new StringReader(json));
+      return unwrapJsonObject(jsonReader);
+    } catch (IllegalStateException | JsonSyntaxException e) {
       throw new IOException(e);
     }
   }
 
   public static Object parseJsonValue(String json) throws IOException {
     try {
-      return unwrapJson(new JSONTokener(json).nextValue());
-    } catch (JSONException e) {
+      JsonReader jsonReader = new JsonReader(new StringReader(json));
+      jsonReader.setLenient(true);
+      return unwrapJson(jsonReader);
+    } catch (IllegalStateException | JsonSyntaxException e) {
       throw new IOException(e);
     }
   }
 
-  @SuppressWarnings("unchecked")
-  private static Map<String, Object> unwrapJsonObject(JSONObject jsonObject) throws JSONException {
-    Map<String, Object> map = new HashMap<>(jsonObject.length());
-    Iterator<String> keys = jsonObject.keys();
-    while (keys.hasNext()) {
-      String key = keys.next();
-      map.put(key, unwrapJson(jsonObject.get(key)));
+  private static Map<String, Object> unwrapJsonObject(JsonReader jsonReader) throws IOException {
+    Map<String, Object> map = new HashMap<>();
+    jsonReader.beginObject();
+    while (jsonReader.peek() != JsonToken.END_OBJECT) {
+      String key = jsonReader.nextName();
+      map.put(key, unwrapJson(jsonReader));
     }
+    jsonReader.endObject();
     return map;
   }
 
-  private static List<Object> unwrapJsonArray(JSONArray jsonArray) throws JSONException {
-    List<Object> list = new ArrayList<>(jsonArray.length());
-    for (int i = 0; i < jsonArray.length(); i++) {
-      list.add(unwrapJson(jsonArray.get(i)));
+  private static List<Object> unwrapJsonArray(JsonReader jsonReader) throws IOException {
+    List<Object> list = new ArrayList<>();
+    jsonReader.beginArray();
+    while (jsonReader.peek() != JsonToken.END_ARRAY) {
+      list.add(unwrapJson(jsonReader));
     }
+    jsonReader.endArray();
     return list;
   }
 
-  private static Object unwrapJson(Object o) throws JSONException {
-    if (o instanceof JSONObject) {
-      return unwrapJsonObject((JSONObject) o);
-    } else if (o instanceof JSONArray) {
-      return unwrapJsonArray((JSONArray) o);
-    } else if (o.equals(JSONObject.NULL)) {
-      return null;
-    } else {
-      return o;
+  private static Object unwrapJson(JsonReader jsonReader) throws IOException {
+    switch (jsonReader.peek()) {
+      case BEGIN_ARRAY:
+        return unwrapJsonArray(jsonReader);
+      case BEGIN_OBJECT:
+        return unwrapJsonObject(jsonReader);
+      case STRING:
+        return jsonReader.nextString();
+      case NUMBER:
+        String value = jsonReader.nextString();
+        if (value.matches("-?\\d+")) {
+          long longValue = Long.parseLong(value);
+          if (longValue <= Integer.MAX_VALUE && longValue >= Integer.MIN_VALUE) {
+            return (int) longValue;
+          }
+          return Long.valueOf(value);
+        }
+        return Double.parseDouble(value);
+      case BOOLEAN:
+        return jsonReader.nextBoolean();
+      case NULL:
+        jsonReader.nextNull();
+        return null;
+      default:
+        throw new IllegalStateException("unknown type " + jsonReader.peek());
     }
   }
+
 }
--- a/src/test/java/com/google/firebase/database/connection/WebsocketConnectionTest.java
+++ b/src/test/java/com/google/firebase/database/connection/WebsocketConnectionTest.java
@@ -112,9 +112,7 @@ public class WebsocketConnectionTest {
   public void testIncomingMessageIOError() {
     MockClientFactory clientFactory = new MockClientFactory();
     ImmutableMap<String, Object> data = ImmutableMap.<String, Object>of("key", "value");
-    Mockito.doThrow(ClassCastException.class).when(clientFactory.delegate).onMessage(data);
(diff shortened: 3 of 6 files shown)


src/main/java/com/google/firebase/database/util/JsonMapper.java
```diff
@@ -46,93 +45,90 @@ public static String serializeJsonValue(Object object) throws IOException {
...
     }
   }
 
   public static Map<String, Object> parseJson(String json) throws IOException {
     try {
-      return unwrapJsonObject(new JSONObject(json));
-    } catch (JSONException e) {
+      JsonReader jsonReader = new JsonReader(new StringReader(json));
+      jsonReader.setLenient(true);
+      return (Map)unwrapJson(jsonReader);
```
hiranya911: Add space. `(Map) unwrapJson(jsonReader);`

src/main/java/com/google/firebase/database/util/JsonMapper.java
```diff
@@ -16,27 +16,26 @@
...
-import org.json.JSONTokener;
-
 /**
  * Helper class to convert from/to JSON strings. TODO: This class should ideally not live in
  * firebase-database-connection, but it's required by both firebase-database and
  * firebase-database-connection, so leave it here for now.
  */
 public class JsonMapper {
 
+  private static Gson GSON = new Gson();
```
hiranya911: final

src/test/java/com/google/firebase/database/connection/WebsocketConnectionTest.java
```diff
@@ -112,7 +112,7 @@ public void testReceiveMultipleFrames() {
   public void testIncomingMessageIOError() {
     MockClientFactory clientFactory = new MockClientFactory();
     ImmutableMap<String, Object> data = ImmutableMap.<String, Object>of("key", "value");
-    Mockito.doThrow(IOException.class).when(clientFactory.delegate).onMessage(data);
+    Mockito.doThrow(ClassCastException.class).when(clientFactory.delegate).onMessage(data);
```
hiranya911: Lets be generic and make it a `RuntimeException`

src/main/java/com/google/firebase/database/util/JsonMapper.java
matejsp: ok

src/main/java/com/google/firebase/database/util/JsonMapper.java
matejsp: ok

src/test/java/com/google/firebase/database/connection/WebsocketConnectionTest.java
matejsp: this won't be ok ... the reason why I added ClassCastException was the way onMessage is implemented.

```
    // Decode JSON
    String combined = buffer.combine();
    try {
      Map<String, Object> decoded = JsonMapper.parseJson(combined);
      logger.debug("{} Parsed complete frame: {}", label, decoded);
      delegate.onMessage(decoded);
    } catch (IOException e) {
      logger.error("{} Error parsing frame: {}", label, combined, e);
      closeAndNotify();
    } catch (ClassCastException e) {
      logger.error("{} Error parsing frame (cast error): {}", label, combined, e);
      closeAndNotify();
    }
```
Point of unit test is to get into closeAndNotify() method.

matejsp:
I have updated changelog, and resolved two issues that you pointed out.
And commented on last issue.

hiranya911:
@matejsp this is failing an integration test:

```
Failed tests: 
  QueryTestIT.testEndAtWithTwoArgumentsAndLimit:1828 Values different.
Expected: {b=b, c=c, d=d, e=e, f=f}
Actual: {d=d, e=e, f=f}
```

hiranya911:
RTDB is responding differently when different json libraries are used.

## Existing behavior (org.json):
```
[firebase-database-worker] DEBUG com.google.firebase.database.connection.Connection - [conn_0] Sending data: {t=d, d={a=s, r=0, b={c={sdk.admin_java.6-7-1-SNAPSHOT=1}}}}
[firebase-database-worker] DEBUG com.google.firebase.database.connection.Connection - [conn_0] Sending data (contents hidden)
[firebase-database-worker] DEBUG com.google.firebase.database.connection.Connection - [conn_0] Sending data: {t=d, d={a=p, r=2, b={p=-LVyO5qTnbP6ihEsnzCr, d={a=a, b=b, c=c, d=d, e=e, f=f, g=g, h=h}}}}
[firebase-database-worker] DEBUG com.google.firebase.database.connection.Connection - [conn_0] Sending data: {t=d, d={a=q, r=3, b={p=-LVyO5qTnbP6ihEsnzCr, q={vf=r, en=f, ep=null, l=5}, t=1, h=}}}
[firebase-database-worker] DEBUG com.google.firebase.database.connection.Connection - [conn_0] Sending data: {t=d, d={a=n, r=4, b={p=-LVyO5qTnbP6ihEsnzCr, q={vf=r, en=f, ep=null, l=5}, t=1}}}

[firebase-websocket-worker] DEBUG com.google.firebase.database.connection.WebsocketConnection - [ws_0] WS message: {"t":"c","d":{"t":"h","d":{"ts":1547234996349,"v":"5","h":"s-usc1c-nss-235.firebaseio.com","s":"lBzQdRCSRFKgGwGChqvtxvxWEQkUv8cO"}}}
[firebase-websocket-worker] DEBUG com.google.firebase.database.connection.WebsocketConnection - [ws_0] WS message: {"t":"d","d":{"r":0,"b":{"s":"ok","d":""}}}
[firebase-websocket-worker] DEBUG com.google.firebase.database.connection.WebsocketConnection - [ws_0] WS message: {"t":"d","d":{"r":1,"b":{"s":"ok","d":{"auth":null,"expires":1547238596}}}}
[firebase-websocket-worker] DEBUG com.google.firebase.database.connection.WebsocketConnection - [ws_0] WS message: {"t":"d","d":{"r":2,"b":{"s":"ok","d":""}}}
[firebase-websocket-worker] DEBUG com.google.firebase.database.connection.WebsocketConnection - [ws_0] WS message: {"t":"d","d":{"b":{"p":"-LVyO5qTnbP6ihEsnzCr","d":{"b":"b","c":"c","d":"d","e":"e","f":"f"},"t":1},"a":"d"}}
```

## Modified behavior (gson)
```
[firebase-database-worker] DEBUG com.google.firebase.database.connection.Connection - [conn_0] Sending data: {t=d, d={a=s, r=0, b={c={sdk.admin_java.6-7-1-SNAPSHOT=1}}}}
[firebase-database-worker] DEBUG com.google.firebase.database.connection.Connection - [conn_0] Sending data (contents hidden)
[firebase-database-worker] DEBUG com.google.firebase.database.connection.Connection - [conn_0] Sending data: {t=d, d={a=p, r=2, b={p=-LVyOZEzspYgAQKjyxxW, d={a=a, b=b, c=c, d=d, e=e, f=f, g=g, h=h}}}}
[firebase-database-worker] DEBUG com.google.firebase.database.connection.Connection - [conn_0] Sending data: {t=d, d={a=q, r=3, b={p=-LVyOZEzspYgAQKjyxxW, q={vf=r, en=f, ep=null, l=5}, t=1, h=}}}
[firebase-database-worker] DEBUG com.google.firebase.database.connection.Connection - [conn_0] Sending data: {t=d, d={a=n, r=4, b={p=-LVyOZEzspYgAQKjyxxW, q={vf=r, en=f, ep=null, l=5}, t=1}}}

[firebase-websocket-worker] DEBUG com.google.firebase.database.connection.WebsocketConnection - [ws_0] WS message: {"t":"c","d":{"t":"h","d":{"ts":1547235117146,"v":"5","h":"s-usc1c-nss-235.firebaseio.com","s":"xgg5G1TWJrs6fUaQwcNttDD8IMV6lhBZ"}}}
[firebase-websocket-worker] DEBUG com.google.firebase.database.connection.WebsocketConnection - [ws_0] WS message: {"t":"d","d":{"r":0,"b":{"s":"ok","d":""}}}
[firebase-websocket-worker] DEBUG com.google.firebase.database.connection.WebsocketConnection - [ws_0] WS message: {"t":"d","d":{"r":1,"b":{"s":"ok","d":{"auth":null,"expires":1547238717}}}}
[firebase-websocket-worker] DEBUG com.google.firebase.database.connection.WebsocketConnection - [ws_0] WS message: {"t":"d","d":{"r":2,"b":{"s":"ok","d":""}}}
[firebase-websocket-worker] DEBUG com.google.firebase.database.connection.WebsocketConnection - [ws_0] WS message: {"t":"d","d":{"b":{"p":"-LVyOZEzspYgAQKjyxxW","d":{"d":"d","e":"e","f":"f","g":"g","h":"h"},"t":1},"a":"d"}}
```

Note that the result of the query is different, which causes the test failure. @rockwotj @mikelehen any ideas?

hiranya911:
I'll try to find the two nodes in the console. This is the failing test:

https://github.com/firebase/firebase-admin-java/blob/6b46deda57c18c534f2bc261b0927a52d4b89ba2/src/test/java/com/google/firebase/database/integration/QueryTestIT.java#L1813-L1829

hiranya911:
@rockwotj The 2 nodes are rendered identically in RTDB afaict. If I run the gson-based implementation against the node created by org.json, the test still fails.

rockwotj:
I would then expect the difference is reading from the network? That looks consistent with the test.

hiranya911:
So these are the two responses we get.

Success case:
```
{"t":"d","d":{"b":{"p":"-LVyO5qTnbP6ihEsnzCr","d":{"b":"b","c":"c","d":"d","e":"e","f":"f"},"t":1},"a":"d"}}
```

Failure case:
```
{"t":"d","d":{"b":{"p":"-LVyOZEzspYgAQKjyxxW","d":{"d":"d","e":"e","f":"f","g":"g","h":"h"},"t":1},"a":"d"}}
```

It seems RTDB returns 5 things in each case, but one of them doesn't respect the `endAt` constraint.

mikelehen:
@schmidt-sebastian is OOO. I'll take a look at this when I get a few spare cycles.  For the test failure, can you see what the SDK is sending to the backend?  If it's getting wrong query results, it's probably sending the wrong query somehow (thanks for digging in!).

rockwotj:
@mikelehen the data looks the same in both cases from the client - if I knew how to run these locally I'd look at what the server is seeing (tcpdump?).

mikelehen:
@rockwotj Gah, sorry. I see you guys covered that above. My bad.

hiranya911:
@rockwotj you can do this to run the test locally.

1. Clone this repo, and import into Idea as a Java project.
2. Get a service account json file from a Firebase project, rename it to `integration_cert.json`, and add to the root of the project.
3. Now you can open `QueryTestIT` class from Idea, pick the test case in question, and just run it.

matejsp:
There was a problem with "null" handling. It seems that firebase code does not handle "null" values very well. I have changed Gson to emit null values (instead of skipping them as is default Gson behaviour).

This test now works again.

rockwotj:
Ah strange those logs must have been wrong then?  Our server does explicitly expect `null` and not missing properties. Good catch!

hiranya911:
Very interesting. I think logs get generated pre-serialization. So it logs the Map in the memory as opposed to the actual outgoing json string.

hiranya911:
Adding a bit of extra logging exposes the problem.

Success case:
```
[ws_0] Raw outgoing message: {"t":"d","d":{"a":"q","r":3,"b":{"p":"-LWClm_BS9_eIPsUPBKw","q":{"vf":"r","en":"f","ep":null,"l":5},"t":1,"h":""}}}
```

Failure case:
```
[ws_0] Raw outgoing message: {"t":"d","d":{"a":"q","r":3,"b":{"p":"-LWClKjGYfTPnSarEl7S","q":{"vf":"r","en":"f","l":5},"t":1,"h":""}}}
```

mikelehen:
I have left a few comments. If we resolve those, then I think I'm happy with this.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
--
+- [changed] json dependency removed
+- [added] java 11 build support
```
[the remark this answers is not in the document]
matejsp: Normal format? it seems that this is formatted md file and not generic CHANGELOG file (without extension). It's more like "github markup changelog format

Check:
https://github.com/firebase/firebase-admin-java/blob/master/CHANGELOG.md

I would hate to break formatting in the middle of the changelog. Perhaps another issue/PR 
should be raised for this change.

src/test/java/com/google/firebase/database/util/JsonMapperTest.java
```diff
@@ -62,13 +62,15 @@ public void testList() throws IOException {
   }
 
   @Test
+  @Ignore
+  // TODO object is not really invalid object from Gson perspective
```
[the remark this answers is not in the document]
matejsp: ok, removing

src/main/java/com/google/firebase/database/util/JsonMapper.java
```diff
@@ -46,93 +45,91 @@ public static String serializeJsonValue(Object object) throws IOException {
...
     }
   }
 
+  @SuppressWarnings("unchecked")
   public static Map<String, Object> parseJson(String json) throws IOException {
     try {
-      return unwrapJsonObject(new JSONObject(json));
-    } catch (JSONException e) {
+      JsonReader jsonReader = new JsonReader(new StringReader(json));
+      jsonReader.setLenient(true);
```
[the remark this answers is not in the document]
matejsp: Lenient parsing here is not really needed (for Map). I have just removed it.

The other location is when parsing parseJsonValue. This is self explanatory because json simple value (like number, string, ...) is not a valid json and we need to be lenient.

matejsp:
commented, fixed, rebased, squashed :)

CHANGELOG.md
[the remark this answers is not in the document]
hiranya911: I think @mikelehen is mostly referring to the content here. Something like:

[changed] Removed `org.json` dependency and replaced with `com.google.code.gson`.
[changed] Upgraded Mockito dependency, and fixed the build on Java 11.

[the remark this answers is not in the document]
matejsp: Oh ok, I have changed the msg as you suggested. Pushed

matejsp:
fixed

src/main/java/com/google/firebase/database/util/JsonMapper.java
```diff
@@ -46,93 +46,90 @@ public static String serializeJsonValue(Object object) throws IOException {
...
+      return unwrapJson(jsonReader);
+    } catch (JsonSyntaxException e) {
       throw new IOException(e);
     }
   }
 
   @SuppressWarnings("unchecked")
```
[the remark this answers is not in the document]
matejsp: You are correct regarding toJson() call ... It seems that the way I was starting converting to json was to convert the code first and didn't check if it can be further optimised.
I have just removed the whole serializeJsonValue method.

src/main/java/com/google/firebase/database/util/JsonMapper.java
```diff
@@ -46,93 +46,90 @@ public static String serializeJsonValue(Object object) throws IOException {
...
     }
   }
 
+  @SuppressWarnings("unchecked")
   public static Map<String, Object> parseJson(String json) throws IOException {
     try {
-      return unwrapJsonObject(new JSONObject(json));
-    } catch (JSONException e) {
+      JsonReader jsonReader = new JsonReader(new StringReader(json));
+      return (Map) unwrapJson(jsonReader);
```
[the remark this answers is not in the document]
matejsp: I have reworked exception handling a bit. Instead of casting I just call unwrapJsonObject. But I have also add IllegalStateException exception because beginArray, beginObject (in gson code) internally throws in case there is unexpected token.

src/test/java/com/google/firebase/database/connection/WebsocketConnectionTest.java
mikelehen: Hrm. I expected this diff to go away. Am I missing something?

matejsp: Mockito can no longer throw IOException for method that does not have throws. It goes again Java language. Mockito used to allow this, but latest version does not (rightfully).

ERROR:
org.mockito.exceptions.base.MockitoException:  Checked exception is invalid for this method! Invalid: java.io.IOException

  interface Delegate {
    void onMessage(Map<String, Object> message);
    void onDisconnect(boolean wasEverConnected);
  }

onMessage does not throw IOException:
Mockito.doThrow(IOException.class).when(clientFactory.delegate).onMessage(data);

I can revert, but then unit test is gonna fail. This fix is related to Mockito version upgrade (for java 11) and does NOT have to do with JsonMapper changes.

a) leave as it is
b) add throws IOException to onMessage for the purpose of unit test (and catch it in all the places it affects)
c) remove the test because it mocks the wrong part
d) adding powermock dep and mock static JsonMapper method

matejsp: I have take the option e) ... 
this code was broken anyway so I have changed it to not even have a mock.

mikelehen:
(and since we always do "squash merge", it's okay to have tons of commits in a PR, so there's no need to force-push)

PR #244, opened 2019-01-24
Fix for the uploadAccountResponse for correct error message/reason

amohabir:
Fix in the UploadAccountResponse object for mapping the error message.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - [changed] Removed org.json dependency and replaced with com.google.code.gson.
 - [changed] Upgraded Mockito dependency, and fixed the build on Java 11.
+- [fixed] UploadAccountResponse now has the correct error message/reason.
 
 # v6.7.0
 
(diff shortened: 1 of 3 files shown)


hiranya911:
@amohabir did you actually see the server responding with `errorMessage` instead of `message`? The existing implementation is consistent with how the responses are parsed in the Node.js SDK: https://github.com/firebase/firebase-admin-node/blob/master/test/unit/auth/user-import-builder.spec.ts#L570

@bojeil-google any thoughts?

amohabir:
@hiranya911 Yep, I actually verified this by debugging up to the HTTP call in the Java code.

hiranya911:
Thank you @amohabir. I can also repro the problem by slightly modifying one of the integration tests. I'll make that part of the test suite, so this gets verified in the future (you're also welcome to do so in this PR). Can you add an entry to the `CHANGELOG` file? Then I can merge this.

@bojeil-google we will need to fix this in Node.js too.

amohabir:
I updated the CHANGELOG.

hiranya911:
This has been fixed in the Auth backend. Reverting this change.

PR #255, opened 2019-02-25
HTTP Retry Support

hiranya911:
We are in the process of implementing HTTP retry support across all Admin SDKs. Specifically, we want to retry an HTTP request in the following situations:

* All I/O errors (i.e. network errors, socket errors etc)
* HTTP 500 and 503 errors

This PR lays the foundation for supporting HTTP retries in the Java Admin SDK. However, we are not enabling retries for any of the APIs just yet. This will need some more refactoring in each of the modules, and therefore we will do it later in an incremental fashion.


src/main/java/com/google/firebase/internal/FirebaseRequestInitializer.java
```diff
@@ -19,32 +19,57 @@
...
-    httpRequest.setReadTimeout(readTimeout);
+  public void initialize(HttpRequest request) throws IOException {
+    for (HttpRequestInitializer initializer : initializers) {
+      initializer.initialize(request);
+    }
+  }
+
+  private static class TimeoutInitializer implements HttpRequestInitializer {
+
+    private final int connectTimeout;
```
schmidt-sebastian: Do you mind including the unit of measurement in the names? Something like `connectTimeoutMs` would be helpful.

src/main/java/com/google/firebase/internal/RetryConfig.java
```diff
@@ -0,0 +1,160 @@
...
+import com.google.api.client.util.Sleeper;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Configures when and how HTTP requests should be retried.
+ */
+public final class RetryConfig {
```
schmidt-sebastian: Did you take a look at just using https://github.com/googleapis/gax-java/blob/master/gax/src/main/java/com/google/api/gax/retrying/RetrySettings.java ?

I think this would be a huge win (if feasible) and we are already using it in Firestore.

src/main/java/com/google/firebase/internal/RetryConfig.java
```diff
@@ -0,0 +1,160 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.firebase.internal;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.api.client.util.BackOff;
```
schmidt-sebastian: FWIW, Firestore is using https://github.com/googleapis/gax-java/blob/master/gax/src/main/java/com/google/api/gax/retrying/ExponentialRetryAlgorithm.java

src/main/java/com/google/firebase/internal/RetryConfig.java
```diff
@@ -0,0 +1,160 @@
...
+    } else {
+      this.retryStatusCodes = ImmutableList.of();
+    }
+
+    checkArgument(builder.maxRetries >= 0, "maxRetries must not be negative");
+    this.maxRetries = builder.maxRetries;
+    this.sleeper = checkNotNull(builder.sleeper);
+    this.backOffBuilder = new ExponentialBackOff.Builder()
+        .setInitialIntervalMillis(INITIAL_INTERVAL_MILLIS)
+        .setMaxIntervalMillis(builder.maxIntervalMillis)
```
schmidt-sebastian: It's not immediately obvious to me why some of these are configureable (max interval, multiplier), while others are not. Have you considered making them all configurable? If that is a direction that we could move to, then we might be able to use `ExponentialBackOff` directly.

src/main/java/com/google/firebase/internal/RetryConfig.java
```diff
@@ -0,0 +1,160 @@
...
+    private int maxIntervalMillis = (int) TimeUnit.MINUTES.toMillis(2);
+    private double backOffMultiplier = 2.0;
+    private Sleeper sleeper = Sleeper.DEFAULT;
+
+    private Builder() { }
+
+    /**
+     * Sets a list of HTTP status codes that should be retried. If null or empty, HTTP requests
+     * will not be retried as long as they result in some HTTP response message. I/O errors
+     * will still be retried.
```
schmidt-sebastian: We have had some long discussion with the Firestore backend team about retrying requests that were dropped and never responded to. Unless we no for certain that no byte made it across the wire, it was decided that we cannot retry. Since you are essentially building a retry framework for a wide variety of consumers, I would advise that we don't retry I/O errors by default.

src/main/java/com/google/firebase/internal/RetryConfig.java
```diff
@@ -0,0 +1,160 @@
...
+     *
+     * @param retryStatusCodes A list of status codes.
+     * @return This builder.
+     */
+    public Builder setRetryStatusCodes(List<Integer> retryStatusCodes) {
```
schmidt-sebastian: Do you foresee that these status codes will be different for each service?

src/main/java/com/google/firebase/internal/RetryInitializer.java
```diff
@@ -0,0 +1,106 @@
...
+      try {
+        boolean retry = preRetryHandler.handleResponse(request, response, supportsRetry);
+        if (!retry) {
+          retry = retryHandler.handleResponse(request, response, supportsRetry);
+        }
+        return retry;
+      } finally {
+        // Pre-retry handler may have reset the unsuccessful response handler on the
+        // request. This changes it back.
+        request.setUnsuccessfulResponseHandler(this);
```
schmidt-sebastian: Do we need to update `preRetryHandler` and replace it with the current retry handler?

src/main/java/com/google/firebase/internal/RetryUnsuccessfulResponseHandler.java
```diff
@@ -0,0 +1,106 @@
...
+import com.google.api.client.util.BackOffUtils;
+import com.google.api.client.util.Clock;
+import com.google.api.client.util.Sleeper;
+import com.google.common.base.Strings;
+import java.io.IOException;
+import java.util.Date;
+import org.apache.http.client.utils.DateUtils;
+
+/**
+ * An {@code HttpUnsuccessfulResponseHandler} that retries failing requests after an interval. The
```
schmidt-sebastian: Word missing at end of sentence.

src/main/java/com/google/firebase/internal/RetryUnsuccessfulResponseHandler.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+    return false;
+  }
+
+  private boolean waitAndRetry(HttpResponse response) throws IOException, InterruptedException {
+    String retryAfterHeader = response.getHeaders().getRetryAfter();
+    if (!Strings.isNullOrEmpty(retryAfterHeader)) {
+      long intervalMillis = parseRetryAfterHeaderIntoMillis(retryAfterHeader.trim());
+      if (intervalMillis > retryConfig.getMaxIntervalMillis()) {
+        return false;
```
schmidt-sebastian: Is there a reason we don't just wait `getMaxIntervalMillis`? If so, can you add a comment?

src/main/java/com/google/firebase/internal/RetryConfig.java
hiranya911: By default we are going to retry on 503 and 500 errors (as recommended at https://cloud.google.com/apis/design/errors#error_retries). But after talking to a few teams about this, we have also decided to make these customizable for individual services. For example, if Auth doesn't want to retry on 500 errors, they should be able to.

src/main/java/com/google/firebase/internal/RetryConfig.java
hiranya911: As far as I can understand gax is tied to GRPC, whereas our HTTP support is provided by the Google API client (everywhere except Firestore). Therefore we have to implement retries using the interfaces provided by the Google API client. We might be able to implement those interfaces using gax algorithm implementations, but I'd rather do this using just one dependency instead of two.

As for `RetrySettings`, it certainly does have all the parameters we'd like to expose. But I'd rather shield the individual classes in the Admin SDK from the dependencies (both gax and API client) by implementing our own configuration class. The way I've implemented this now, individual classes like `FirebaseAuth` and `FirebaseMessaging` don't have to know how the retries are implemented, and they don't have to take a hard dependency on API client. They just pass a `RetryConfig` (which is an interface we control), and the `internal` package provides retries in a single place.

In the future, if we decide to change how retries are implemented (say we manage to migrate to gax), we just need to change the `internal` package.

src/main/java/com/google/firebase/internal/RetryConfig.java
hiranya911: We have rolled out similar retry mechanisms in Python and Go. I was trying to be consistent with those implementations. The default policy we want to support is:

1. Retry up to N times on selected HTTP status codes
2. Delay subsequent retries with exponential backoff
3. Support the `Retry-After` header when available, as long as it is below a specified maximum

I designed `RetryConfig` to have the minimum set of controls required to configure the above policy. We can add more controls in the future, if necessary. In either case I don't think we want to expose Google API client types to other classes in the SDK. Less exposure to the third-party code, the better.

src/main/java/com/google/firebase/internal/RetryInitializer.java
hiranya911: No it should be initialized only once per request.

src/main/java/com/google/firebase/internal/RetryUnsuccessfulResponseHandler.java
hiranya911: Reads correct to me. Do you have a suggestion to improve?

src/main/java/com/google/firebase/internal/RetryUnsuccessfulResponseHandler.java
hiranya911: Added a comment to explain

src/main/java/com/google/firebase/internal/RetryConfig.java
hiranya911: In Python and Go we have enabled this, but I can see why this can be a bad idea in some cases. Added an option to make this configurable via `RetryConfig`.

src/main/java/com/google/firebase/internal/FirebaseRequestInitializer.java
hiranya911: Done

hiranya911:
Thanks @schmidt-sebastian. I've made some changes and responded to your comments. In general, I don't see how to reuse gax retry mechanism in our code, since we don't use gax. But I've managed to make most of the other improvements suggested.

src/test/java/com/google/firebase/internal/RetryUnsuccessfulResponseHandlerTest.java
```diff
@@ -0,0 +1,281 @@
...
+
+    assertEquals(MAX_RETRIES, sleeper.getCount());
+    assertArrayEquals(new long[]{500, 1000, 2000, 4000}, sleeper.getDelays());
+    assertEquals(MAX_RETRIES + 1, failingRequest.getCount());
+  }
+
+
+  @Test
+  public void testRetryAfterIsAbsent() throws IOException {
+    MultipleCallSleeper sleeper = new MultipleCallSleeper();
```
schmidt-sebastian: Cough cough :) This test is basically the same as the one above.

src/test/java/com/google/firebase/internal/RetryUnsuccessfulResponseHandlerTest.java
```diff
@@ -0,0 +1,281 @@
...
+    try {
+      request.execute();
+      fail("No exception thrown for HTTP error");
+    } catch (HttpResponseException e) {
+      assertEquals(503, e.getStatusCode());
+    }
+
+    assertEquals(10, sleeper.getCount());
+    assertArrayEquals(
+        new long[]{500, 1000, 2000, 4000, 8000, 16000, 32000, 64000, 120000, 120000},
```
schmidt-sebastian: There are quite a few tests in this PR that test the same feature at different levels (see testExponentialBackOff). You can certainly leave this as is, but it might be worth considering to reduce duplication in future PRs.

src/main/java/com/google/firebase/internal/RetryConfig.java
```diff
@@ -0,0 +1,178 @@
...
+
+  private static final int INITIAL_INTERVAL_MILLIS = 500;
+
+  private final List<Integer> retryStatusCodes;
+  private final boolean retryOnIOExceptions;
+  private final int maxRetries;
+  private final Sleeper sleeper;
+  private final ExponentialBackOff.Builder backOffBuilder;
+
+  private RetryConfig(Builder builder) {
```
schmidt-sebastian: Feel free to ignore, but I think you could use `@AutoValue` here since this is internal and doesn't leak to our customers. You are already getting it for free as part of the Firestore dependencies.

src/test/java/com/google/firebase/internal/RetryUnsuccessfulResponseHandlerTest.java
hiranya911: Removed :)

src/test/java/com/google/firebase/internal/RetryInitializerTest.java
```diff
@@ -0,0 +1,236 @@
...
+      fail("No exception thrown for HTTP error");
+    } catch (Exception e) {
+      assertEquals("test", e.getMessage());
+    }
+
+    assertEquals(1, otherErrorHandlerCalls.get());
+    assertSame(retryHandler, request.getUnsuccessfulResponseHandler());
+  }
+
+  private RetryConfig testRetryConfig(Sleeper sleeper) {
```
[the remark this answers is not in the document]
hiranya911: Renamed the helper to `retryOnIOAndServiceUnavailableErrors()`

src/test/java/com/google/firebase/internal/RetryInitializerTest.java
```diff
@@ -0,0 +1,236 @@
...
+        .setMaxRetries(MAX_RETRIES)
+        .setRetryOnIOExceptions(false)
+        .setRetryStatusCodes(ImmutableList.of(503))
+        .build());
+    HttpRequest request = TestUtils.createRequest();
+
+    initializer.initialize(request);
+
+    assertEquals(MAX_RETRIES, request.getNumberOfRetries());
+    assertNotNull(request.getUnsuccessfulResponseHandler());
```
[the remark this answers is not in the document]
hiranya911: Done

src/test/java/com/google/firebase/internal/RetryConfigTest.java
```diff
@@ -0,0 +1,124 @@
...
+  public void testBuilderWithAllSettings() {
+    ImmutableList<Integer> statusCodes = ImmutableList.of(500, 503);
+    Sleeper sleeper = new MockSleeper();
+    RetryConfig config = RetryConfig.builder()
+        .setMaxRetries(4)
+        .setRetryStatusCodes(statusCodes)
+        .setMaxIntervalMillis(5 * 60 * 1000)
+        .setBackOffMultiplier(1.5)
+        .setSleeper(sleeper)
+        .build();
```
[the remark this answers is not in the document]
hiranya911: Done

src/main/java/com/google/firebase/internal/RetryInitializer.java
```diff
@@ -0,0 +1,108 @@
...
+import com.google.api.client.http.HttpIOExceptionHandler;
+import com.google.api.client.http.HttpRequest;
+import com.google.api.client.http.HttpRequestInitializer;
+import com.google.api.client.http.HttpResponse;
+import com.google.api.client.http.HttpUnsuccessfulResponseHandler;
+import java.io.IOException;
+
+/**
+ * Configures HTTP requests to be retried. Requests that encounter I/O errors are always retried
+ * with exponential back off. Requests failing with unsuccessful HTTP responses are first referred
```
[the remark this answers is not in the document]
hiranya911: Done

src/main/java/com/google/firebase/internal/RetryConfig.java
hiranya911: We don't use it anywhere in the codebase at the moment. I'd like to not introduce another dependency at this point.

PR #259, opened 2019-03-12
Removing the hard dependency on Apache HTTP Client

hiranya911:
In #255 we used an Apache HTTP client API (`DateUtils`) to parse HTTP date headers. We have this library in our classpath at the moment as a transitive dependency of Google API client. But the Google API client is undergoing a refactor, and this dependency will be removed in a future release. In preparation for this, I'm creating our own `DateUtils` class by copying the related bit of code from the Apache HTTP client.

The original implementation in Apache HTTP client uses thread locals to cache the `SimpleDateFormat` instances, and use them in a thread safe manner. I used explicit locking instead, in order to avoid potential memory leaks that may arise from unreleased thread locals (other suggestions welcome).

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -1,0 +1,5 @@
+Firebase Admin Java SDK
+Copyright 2019 Google Inc.
+
+This product includes software developed at
+The Apache Software Foundation (http://www.apache.org/).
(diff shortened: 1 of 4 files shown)


hiranya911:
Added `NOTICE.txt` file, and also removed the `Copyright Google Inc.` statement from the `DateUtils` class.

matejsp:
How about migrating to DateTimeFormatter? However it is only supported in Java 8 and forward.A lot of extra code for obsolete and EOL Java version.

hiranya911:
We cannot use Java 8 at the moment. Current GCP policy requires us to be Java 7 compatible.

PR #262, opened 2019-03-16
Reject non-standard Number classes during POJO serialization (RTDB)

schmidt-sebastian:
This ports the changes from https://github.com/firebase/firebase-android-sdk/pull/272 and updates the error messages in the RTDB client when a user attempts to serialize Number types that we cannot serialize to JSON.

Addresses: https://github.com/firebase/firebase-admin-java/issues/75

### DIFF
--- a/src/main/java/com/google/firebase/database/utilities/encoding/CustomClassMapper.java
+++ b/src/main/java/com/google/firebase/database/utilities/encoding/CustomClassMapper.java
@@ -123,20 +123,20 @@ public class CustomClassMapper {
           return ((Number) obj).longValue();
         }
         return doubleValue;
-      } else if (obj instanceof Short) {
-        throw new DatabaseException("Shorts are not supported, please use int or long");
-      } else if (obj instanceof Byte) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/firebase/database/utilities/encoding/CustomClassMapper.java
```diff
@@ -498,7 +493,7 @@ public BeanMapper(Class<T> clazz) {
             if (existingPropertyName != null) {
               if (!existingPropertyName.equals(propertyName)) {
                 throw new DatabaseException(
-                    "Found setter with invalid case-sensitive name: " + method.getName());
+                    "Found setter with invalid " + "case-sensitive name: " + method.getName());
```
hiranya911: Seems unnecessary.

schmidt-sebastian: I removed this instance and the remaining string concatenations in this file that follow a similar pattern.

hiranya911: Thanks 👍

PR #281, opened 2019-06-19
Change addAllTokens argument data type from List to Collection (fireb…

NunuM:
Modify the argument data type to a more permissive one. This change permits the caller to pass as an argument a data structure that better reflects their requirements.

### DIFF
--- a/src/main/java/com/google/firebase/messaging/MulticastMessage.java
+++ b/src/main/java/com/google/firebase/messaging/MulticastMessage.java
@@ -115,7 +115,7 @@ public class MulticastMessage {
     }
 
     /**
-     * Adds a list of tokens to which the message should be sent. Up to 100 tokens can be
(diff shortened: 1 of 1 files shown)


hiranya911:
Thanks @NunuM. I'm trying to get this API change approved. Sit tight.

src/main/java/com/google/firebase/messaging/MulticastMessage.java
```diff
@@ -116,10 +118,10 @@ public Builder addToken(@NonNull String token) {
      * Adds a list of tokens to which the message should be sent. Up to 100 tokens can be
```
hiranya911: Please update to `Adds a collection tokens to which...`.

Also please specify that the tokens will be added in the iteration order of the collection.

NunuM: @hiranya911 Thanks to pointing out. Already make the change. Thank you.

PR #285, opened 2019-07-02
Add support for analytics_label to be specified

the-real-mrcs:


src/main/java/com/google/firebase/internal/FcmOptionsUtil.java
```diff
@@ -0,0 +1,15 @@
+package com.google.firebase.internal;
```
hiranya911: Add the license header.

hiranya911: Let's move this into the `messaging` package and drop `public` from the class definition. There's really no need for this class to be visible outside the `messaging` package.

src/main/java/com/google/firebase/internal/FcmOptionsUtil.java
```diff
@@ -0,0 +1,15 @@
...
+
+import java.util.regex.Pattern;
+
+public final class FcmOptionsUtil {
+
+  private static final Pattern ANALYTICS_LABEL_REGEX = Pattern.compile("^[a-zA-Z0-9-_.~%]{0,50}$");
```
hiranya911: Nit: Perhaps add a comment describing what the accepted pattern is.

src/main/java/com/google/firebase/messaging/AndroidFcmOptions.java
```diff
@@ -0,0 +1,67 @@
+package com.google.firebase.messaging;
```
hiranya911: Add license header

src/main/java/com/google/firebase/messaging/AndroidFcmOptions.java
```diff
@@ -0,0 +1,67 @@
...
+import static com.google.common.base.Preconditions.checkArgument;
+
+import com.google.api.client.util.Key;
+import com.google.firebase.internal.FcmOptionsUtil;
+
+/**
+ * Represents the Android-specific FCM options that can be included in an {@link AndroidConfig}.
+ * Instances of this class are thread-safe and immutable.
+ */
+public class AndroidFcmOptions {
```
hiranya911: Lets also make this `final`. Same goes for all other public classes introduced in this PR.

src/main/java/com/google/firebase/messaging/AndroidFcmOptions.java
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  /**
+   * Creates a new {@link AndroidFcmOptions} with the specified analytics label.
+   *
+   * @param analyticsLabel An analytics label
+   * @return An AndroidFcmOptions with the analytics label set to the supplied value.
+   */
+  public static AndroidFcmOptions withAnalyticsLabel(String analyticsLabel) {
+    return new AndroidFcmOptions(new Builder().setAnalyticsLabel(analyticsLabel));
```
hiranya911: Nit: Use the public API here. We should minimize direct invocations of the private APIs.

```
return builder().setAnalyticsLabel(analyticsLabel).build();
```

src/main/java/com/google/firebase/messaging/AndroidFcmOptions.java
```diff
@@ -0,0 +1,67 @@
...
+ * Represents the Android-specific FCM options that can be included in an {@link AndroidConfig}.
+ * Instances of this class are thread-safe and immutable.
+ */
+public class AndroidFcmOptions {
+
+  @Key("analytics_label")
+  private final String analyticsLabel;
+
+  private AndroidFcmOptions(Builder builder) {
+    checkArgument(FcmOptionsUtil.validAnalyticsLabelFormat(builder.analyticsLabel),
```
hiranya911: This essentially makes the analyticsLabel required. Was that the intention? And what happens if analyticsLabel is null? You might want to check for that in your util.

src/main/java/com/google/firebase/messaging/ApnsFcmOptions.java
```diff
@@ -0,0 +1,67 @@
+package com.google.firebase.messaging;
```
hiranya911: Add license header and make the class final.

src/main/java/com/google/firebase/messaging/ApnsFcmOptions.java
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  /**
+   * Creates a new {@link ApnsFcmOptions} with the specified analytics label.
+   *
+   * @param analyticsLabel An analytics label
+   * @return An ApnsFcmOptions with the analytics label set to the supplied value.
+   */
+  public static ApnsFcmOptions withAnalyticsLabel(String analyticsLabel) {
+    return new ApnsFcmOptions(new Builder().setAnalyticsLabel(analyticsLabel));
```
hiranya911: Use public API.

src/main/java/com/google/firebase/messaging/FcmOptions.java
```diff
@@ -0,0 +1,68 @@
...
+ * Instances of this class are thread-safe and immutable.
+ */
+public class FcmOptions {
+
+  @Key("analytics_label")
+  private final String analyticsLabel;
+
+  private FcmOptions(Builder builder) {
+    checkArgument(FcmOptionsUtil.validAnalyticsLabelFormat(builder.analyticsLabel),
+        "Analytics label must have format matching'^[a-zA-Z0-9-_.~%]{1,50}$'");
```
hiranya911: I see this error message is repeated in several places. You can avoid this duplication by slightly modifying your util method:

```
public static void checkAnalyticsLabel(String label) {
  checkArgument(isValid(label), "Analytics label must have format matching'^[a-zA-Z0-9-_.~%]{1,50}$");
}

// Then you can call this as...
FcmOptionsUtil.checkAnalyticsLabel(label);
```

src/test/java/com/google/firebase/messaging/MessageTest.java
```diff
@@ -583,6 +583,72 @@ public void testWrapForTransport() {
...
+      FcmOptions.builder()
+          .setAnalyticsLabel("THIS_IS_LONGER_THAN_50_CHARACTERS_WHICH_IS_NOT_ALLOWED")
+          .build();
+      fail("No error thrown when using bad analytics label format.");
+    } catch (IllegalArgumentException expected) {
+      //expected
+    }
+
+    try {
+      FcmOptions.builder().setAnalyticsLabel("   ").build();
```
hiranya911: Also check for the `label = null` case.

src/main/java/com/google/firebase/messaging/AndroidFcmOptions.java
the-real-mrcs: done

src/main/java/com/google/firebase/messaging/ApnsFcmOptions.java
the-real-mrcs: done

src/test/java/com/google/firebase/messaging/MessageTest.java
the-real-mrcs: Added it to the FcmOptionsUtil.checkAnalyticsLabel

src/main/java/com/google/firebase/internal/FcmOptionsUtil.java
the-real-mrcs: done

the-real-mrcs: done

src/main/java/com/google/firebase/messaging/AndroidFcmOptions.java
the-real-mrcs: done

src/main/java/com/google/firebase/messaging/AndroidFcmOptions.java
the-real-mrcs: Wops, nice catch! Added test for this as well.

src/main/java/com/google/firebase/messaging/ApnsFcmOptions.java
the-real-mrcs: done

src/main/java/com/google/firebase/messaging/FcmOptions.java
the-real-mrcs: done.

src/main/java/com/google/firebase/internal/FcmOptionsUtil.java
the-real-mrcs: done.

src/main/java/com/google/firebase/messaging/AndroidFcmOptions.java
the-real-mrcs: done

src/main/java/com/google/firebase/messaging/ApnsFcmOptions.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  /**
+   * Creates a new {@link ApnsFcmOptions} with the specified analytics label.
+   *
+   * @param analyticsLabel An analytics label
+   * @return An ApnsFcmOptions with the analytics label set to the supplied value.
+   */
+  public static ApnsFcmOptions withAnalyticsLabel(String analyticsLabel) {
+    return new ApnsFcmOptions(builder().setAnalyticsLabel(analyticsLabel));
```
hiranya911: Nit: This could be written just using the public APIs

```
return builder().setAnalyticsLabel(analyticsLabel).build();
```

Same in other similar places.

the-real-mrcs: done

the-real-mrcs:
Thanks for reviewing!

PR #298, opened 2019-08-05
Add image in notification support.

chong-shao:
Add image in notification support.

### Testing

  * Added unit tests and modified one existing integration test to reflect this change.

### API Changes

  * In Firebase Cloud Messaging, added "image" field in the general Notification class, the AndroidNotification class, and the ApnsFcmOptions class.

### DIFF
--- a/src/main/java/com/google/firebase/messaging/Notification.java
+++ b/src/main/java/com/google/firebase/messaging/Notification.java
@@ -29,6 +29,9 @@ public class Notification {
 
   @Key("body")
   private final String body;
+  
+  @Key("image")
+  private final String image;
 
   /**
    * Creates a new {@code Notification} using the given title and body.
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/firebase/messaging/Notification.java
```diff
@@ -39,6 +42,20 @@
   public Notification(String title, String body) {
     this.title = title;
     this.body = body;
+    this.image = null;
```
hiranya911: `this(title, body, null);`

src/test/java/com/google/firebase/messaging/FirebaseMessagingIT.java
```diff
@@ -44,7 +44,7 @@ public static void setUpClass() {
   public void testSend() throws Exception {
     FirebaseMessaging messaging = FirebaseMessaging.getInstance();
     Message message = Message.builder()
-        .setNotification(new Notification("Title", "Body"))
+        .setNotification(new Notification("Title", "Body", "Image.png"))
```
hiranya911: Use something like `https://example.com/image.png` as the test value.

src/test/java/com/google/firebase/messaging/MessageTest.java
```diff
@@ -678,6 +678,76 @@ public void testIncorrectAnalyticsLabelFormat() {
     }
   }
 
+  @Test
+  public void testImageInNotification() throws IOException {
+    Message message = Message.builder()
+        .setNotification(new Notification("title", "body", "image.png"))
```
hiranya911: Use a URL test value. Define a constant at the top of the file and reuse it in the new test cases.

PR #299, opened 2019-08-12
feat(rtdb): Connect to RTDB emulator when valid emulator URL is passed OR env vars are set correctly

sarajmunjal:
This PR makes it possible for the Java admin SDK to talk to the RTDB emulator instead of the actual database instance.
This can happen in the following fashion:-
1. If the user passes in a valid emulator URL (http scheme only, host can't contain `firebaseio.com`, has the `ns` query parameter)
2. Else, if the environment variable `FIREBASE_RTDB_EMULATOR_HOST` is set to a value of the form `host:port`.

If either 1. or 2. is true, we decide to talk to the RTDB emulator. In such cases, we 
- provide a valid emulator URL (either by replacing the supplied URL or making a new one)
- replace existing credentials with `EmulatorCredentials`, which use `owner` as the access token.

URL replacement happens in three places:- 
1. while constructing `FirebaseOptions` instances in `FirebaseApp`
2. when a specific `databaseUrl` is passed in `FirebaseDatabase.getInstance(String url)`
3. when a specific URL with path is passed in `FirebaseDatabase.getReferenceFromUrl(String url)`

Tests have been added with different combinations of passed-in URLs and environment variable values for classes `FirebaseApp`, `FirebaseDatabase`, and also for  `EmulatorHelper`: the helper class that actually does URL replacement if necessary.

**Notes for reviewers**
I have made changes to existing behavior to `Utilities.parseUrl(String url)`: these changes were necessary to recognize the `ns` query parameter, and also to suppost valid `<net_loc>` URL semantics where a URL like `http://localhost:8080/` is deemed valid.
These changes are nearly identical to changes to the same file in [this Android SDK PR](https://github.com/firebase/firebase-android-sdk/pull/680/files).

### DIFF
--- a/src/main/java/com/google/firebase/FirebaseApp.java
+++ b/src/main/java/com/google/firebase/FirebaseApp.java
@@ -33,9 +33,9 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Joiner;
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Strings;
+import com.google.common.base.Supplier;
+import com.google.common.base.Suppliers;
 import com.google.common.collect.ImmutableList;
-import com.google.firebase.FirebaseOptions.DeferredApplicationCredentials;
-import com.google.firebase.FirebaseOptions.DeferredApplicationCredentials.CredentialsGenerator;
 import com.google.firebase.internal.FirebaseAppStore;
 import com.google.firebase.internal.FirebaseScheduledExecutor;
 import com.google.firebase.internal.FirebaseService;
@@ -584,7 +584,7 @@ public class FirebaseApp {
     String defaultConfig = System.getenv(FIREBASE_CONFIG_ENV_VAR);
     if (Strings.isNullOrEmpty(defaultConfig)) {
       return new FirebaseOptions.Builder()
-          .setDeferredCredentials(DEFAULT_OR_EMPTY_CREDS)
+          .setCredentials(defaultOrNullCredentials())
           .build();
     }
     JsonFactory jsonFactory = Utils.getDefaultJsonFactory();
@@ -598,22 +598,17 @@ public class FirebaseApp {
       parser = jsonFactory.createJsonParser(reader);
     }
     parser.parseAndClose(builder);
-    builder.setDeferredCredentials(DEFAULT_OR_EMPTY_CREDS);
+    builder.setCredentials(defaultOrNullCredentials());
     return builder.build();
   }
 
-  private static final DeferredApplicationCredentials DEFAULT_OR_EMPTY_CREDS =
-      new DeferredApplicationCredentials(new CredentialsGenerator() {
-        @Override
-        public GoogleCredentials generate() {
-          try {
-            return GoogleCredentials.getApplicationDefault();
-          } catch (IOException e) {
-            logger
-                .error("Failed to fetch default application credentials, proceeding with null", e);
-            return null;
-          }
-        }
-      }
-      );
+  private static Supplier<GoogleCredentials> defaultOrNullCredentials() {
+    try {
+      return Suppliers.ofInstance(GoogleCredentials.getApplicationDefault());
+    } catch (IOException e) {
+      logger
+          .error("Failed to fetch default application credentials, proceeding with null", e);
+      return Suppliers.ofInstance(null);
+    }
+  }
 }
--- a/src/main/java/com/google/firebase/FirebaseApp.java
+++ b/src/main/java/com/google/firebase/FirebaseApp.java
@@ -34,7 +34,8 @@ import com.google.common.base.Joiner;
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
-import com.google.firebase.database.util.EmulatorHelper;
+import com.google.firebase.FirebaseOptions.DeferredApplicationCredentials;
+import com.google.firebase.FirebaseOptions.DeferredApplicationCredentials.CredentialsGenerator;
 import com.google.firebase.internal.FirebaseAppStore;
 import com.google.firebase.internal.FirebaseScheduledExecutor;
 import com.google.firebase.internal.FirebaseService;
@@ -46,7 +47,6 @@ import java.io.FileReader;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -583,7 +583,8 @@ public class FirebaseApp {
   private static FirebaseOptions getOptionsFromEnvironment() throws IOException {
     String defaultConfig = System.getenv(FIREBASE_CONFIG_ENV_VAR);
     if (Strings.isNullOrEmpty(defaultConfig)) {
-      return new FirebaseOptions.Builder().setCredentials(getApplicationDefaultOrEmptyCredentials())
+      return new FirebaseOptions.Builder()
+          .setDeferredCredentials(DEFAULT_OR_EMPTY_CREDS)
           .build();
     }
     JsonFactory jsonFactory = Utils.getDefaultJsonFactory();
@@ -597,18 +598,22 @@ public class FirebaseApp {
       parser = jsonFactory.createJsonParser(reader);
     }
     parser.parseAndClose(builder);
-    builder.setCredentials(getApplicationDefaultOrEmptyCredentials());
+    builder.setDeferredCredentials(DEFAULT_OR_EMPTY_CREDS);
     return builder.build();
   }
 
-  private static GoogleCredentials getApplicationDefaultOrEmptyCredentials() {
-    try {
-      return GoogleCredentials.getApplicationDefault();
-    } catch (IOException e) {
-      logger.error(
-          "Failed to fetch default application credentials. Proceeding with blank credentials.",
-          e);
-      return GoogleCredentials.newBuilder().build();
-    }
-  }
+  private static final DeferredApplicationCredentials DEFAULT_OR_EMPTY_CREDS =
+      new DeferredApplicationCredentials(new CredentialsGenerator() {
+        @Override
+        public GoogleCredentials generate() {
+          try {
+            return GoogleCredentials.getApplicationDefault();
+          } catch (IOException e) {
+            logger
+                .error("Failed to fetch default application credentials, proceeding with null", e);
+            return null;
+          }
+        }
+      }
+      );
 }
--- a/src/main/java/com/google/firebase/FirebaseApp.java
+++ b/src/main/java/com/google/firebase/FirebaseApp.java
@@ -92,6 +92,17 @@ public class FirebaseApp {
    * accesses to instances map should be protected by this lock.
    */
   private static final Object appsLock = new Object();
+  private static final Supplier<GoogleCredentials> DEFAULT_CREDS_CALLABLE =
+      new Supplier<GoogleCredentials>() {
+        @Override
+        public GoogleCredentials get() {
+          try {
+            return GoogleCredentials.getApplicationDefault();
+          } catch (IOException e) {
+            throw new IllegalStateException(e);
+          }
+        }
+      };
 
   private final String name;
   private final FirebaseOptions options;
@@ -584,7 +595,7 @@ public class FirebaseApp {
     String defaultConfig = System.getenv(FIREBASE_CONFIG_ENV_VAR);
     if (Strings.isNullOrEmpty(defaultConfig)) {
       return new FirebaseOptions.Builder()
-          .setCredentials(defaultOrNullCredentials())
+          .setCredentials(DEFAULT_CREDS_CALLABLE)
           .build();
     }
     JsonFactory jsonFactory = Utils.getDefaultJsonFactory();
@@ -598,17 +609,7 @@ public class FirebaseApp {
       parser = jsonFactory.createJsonParser(reader);
     }
     parser.parseAndClose(builder);
-    builder.setCredentials(defaultOrNullCredentials());
+    builder.setCredentials(DEFAULT_CREDS_CALLABLE);
     return builder.build();
-  }
-
-  private static Supplier<GoogleCredentials> defaultOrNullCredentials() {
-    try {
-      return Suppliers.ofInstance(GoogleCredentials.getApplicationDefault());
-    } catch (IOException e) {
-      logger
-          .error("Failed to fetch default application credentials, proceeding with null", e);
-      return Suppliers.ofInstance(null);
-    }
   }
 }
--- a/src/main/java/com/google/firebase/FirebaseApp.java
+++ b/src/main/java/com/google/firebase/FirebaseApp.java
@@ -19,6 +19,7 @@ package com.google.firebase;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
+import static com.google.firebase.FirebaseOptions.APPLICATION_DEFAULT_CREDENTIALS;
 
 import com.google.api.client.googleapis.util.Utils;
 import com.google.api.client.json.JsonFactory;
@@ -34,7 +35,6 @@ import com.google.common.base.Joiner;
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Strings;
 import com.google.common.base.Supplier;
-import com.google.common.base.Suppliers;
 import com.google.common.collect.ImmutableList;
 import com.google.firebase.internal.FirebaseAppStore;
 import com.google.firebase.internal.FirebaseScheduledExecutor;
@@ -92,17 +92,6 @@ public class FirebaseApp {
    * accesses to instances map should be protected by this lock.
    */
   private static final Object appsLock = new Object();
-  private static final Supplier<GoogleCredentials> DEFAULT_CREDS_CALLABLE =
-      new Supplier<GoogleCredentials>() {
-        @Override
-        public GoogleCredentials get() {
-          try {
-            return GoogleCredentials.getApplicationDefault();
-          } catch (IOException e) {
-            throw new IllegalStateException(e);
-          }
-        }
-      };
 
   private final String name;
   private final FirebaseOptions options;
@@ -595,7 +584,7 @@ public class FirebaseApp {
     String defaultConfig = System.getenv(FIREBASE_CONFIG_ENV_VAR);
     if (Strings.isNullOrEmpty(defaultConfig)) {
       return new FirebaseOptions.Builder()
-          .setCredentials(DEFAULT_CREDS_CALLABLE)
+          .setCredentials(APPLICATION_DEFAULT_CREDENTIALS)
           .build();
     }
     JsonFactory jsonFactory = Utils.getDefaultJsonFactory();
@@ -609,7 +598,7 @@ public class FirebaseApp {
       parser = jsonFactory.createJsonParser(reader);
     }
     parser.parseAndClose(builder);
-    builder.setCredentials(DEFAULT_CREDS_CALLABLE);
+    builder.setCredentials(APPLICATION_DEFAULT_CREDENTIALS);
     return builder.build();
   }
 }
--- a/src/main/java/com/google/firebase/FirebaseApp.java
+++ b/src/main/java/com/google/firebase/FirebaseApp.java
@@ -34,6 +34,7 @@ import com.google.common.base.Joiner;
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
+import com.google.firebase.database.util.EmulatorHelper;
 import com.google.firebase.internal.FirebaseAppStore;
 import com.google.firebase.internal.FirebaseScheduledExecutor;
 import com.google.firebase.internal.FirebaseService;
@@ -45,6 +46,7 @@ import java.io.FileReader;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -581,11 +583,9 @@ public class FirebaseApp {
   private static FirebaseOptions getOptionsFromEnvironment() throws IOException {
     String defaultConfig = System.getenv(FIREBASE_CONFIG_ENV_VAR);
     if (Strings.isNullOrEmpty(defaultConfig)) {
-      return new FirebaseOptions.Builder()
-        .setCredentials(GoogleCredentials.getApplicationDefault())
-        .build();
+      return new FirebaseOptions.Builder().setCredentials(getApplicationDefaultOrEmptyCredentials())
+          .build();
     }
-
     JsonFactory jsonFactory = Utils.getDefaultJsonFactory();
     FirebaseOptions.Builder builder = new FirebaseOptions.Builder();
     JsonParser parser;
@@ -597,7 +597,18 @@ public class FirebaseApp {
       parser = jsonFactory.createJsonParser(reader);
     }
     parser.parseAndClose(builder);
-    builder.setCredentials(GoogleCredentials.getApplicationDefault());
+    builder.setCredentials(getApplicationDefaultOrEmptyCredentials());
     return builder.build();
+  }
+
+  private static GoogleCredentials getApplicationDefaultOrEmptyCredentials() {
+    try {
+      return GoogleCredentials.getApplicationDefault();
+    } catch (IOException e) {
+      logger.error(
+          "Failed to fetch default application credentials. Proceeding with blank credentials.",
+          e);
+      return GoogleCredentials.newBuilder().build();
+    }
   }
 }
--- a/src/main/java/com/google/firebase/FirebaseOptions.java
+++ b/src/main/java/com/google/firebase/FirebaseOptions.java
@@ -33,6 +33,7 @@ import com.google.firebase.internal.FirebaseThreadManagers;
 import com.google.firebase.internal.NonNull;
 import com.google.firebase.internal.Nullable;
 
+import java.io.IOException;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -57,6 +58,18 @@ public final class FirebaseOptions {
           // Enables access to Google Cloud Firestore
           "https://www.googleapis.com/auth/cloud-platform",
           "https://www.googleapis.com/auth/datastore");
+
+  static final Supplier<GoogleCredentials> APPLICATION_DEFAULT_CREDENTIALS =
+      new Supplier<GoogleCredentials>() {
+        @Override
+        public GoogleCredentials get() {
+          try {
+            return GoogleCredentials.getApplicationDefault();
+          } catch (IOException e) {
+            throw new IllegalStateException(e);
+          }
+        }
+      };
 
   private final String databaseUrl;
   private final String storageBucket;
--- a/src/main/java/com/google/firebase/FirebaseOptions.java
+++ b/src/main/java/com/google/firebase/FirebaseOptions.java
@@ -23,17 +23,21 @@ import com.google.api.client.googleapis.util.Utils;
 import com.google.api.client.http.HttpTransport;
 import com.google.api.client.json.JsonFactory;
 import com.google.api.client.util.Key;
+import com.google.auth.oauth2.AccessToken;
 import com.google.auth.oauth2.GoogleCredentials;
 import com.google.cloud.firestore.FirestoreOptions;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
+import com.google.firebase.database.util.EmulatorHelper;
 import com.google.firebase.internal.FirebaseThreadManagers;
 import com.google.firebase.internal.NonNull;
 import com.google.firebase.internal.Nullable;
 
+import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 
 /** Configurable Firebase options. */
 public final class FirebaseOptions {
@@ -70,10 +74,16 @@ public final class FirebaseOptions {
   private final FirestoreOptions firestoreOptions;
 
   private FirebaseOptions(@NonNull FirebaseOptions.Builder builder) {
-    this.credentials = checkNotNull(builder.credentials,
-        "FirebaseOptions must be initialized with setCredentials().")
-        .createScoped(FIREBASE_SCOPES);
-    this.databaseUrl = builder.databaseUrl;
+    String emulatorUrl = EmulatorHelper.getEmulatorUrl(builder.databaseUrl);
+    if (!Strings.isNullOrEmpty(emulatorUrl)) {
+      this.databaseUrl = emulatorUrl;
+      this.credentials = new EmulatorCredentials();
+    } else {
+      this.databaseUrl = builder.databaseUrl;
+      this.credentials = checkNotNull(builder.credentials,
+          "FirebaseOptions must be initialized with setCredentials().")
+          .createScoped(FIREBASE_SCOPES);
+    }
     this.databaseAuthVariableOverride = builder.databaseAuthVariableOverride;
     this.projectId = builder.projectId;
     if (!Strings.isNullOrEmpty(builder.storageBucket)) {
@@ -453,6 +463,23 @@ public final class FirebaseOptions {
      */
     public FirebaseOptions build() {
       return new FirebaseOptions(this);
+    }
+  }
+
+  private static class EmulatorCredentials extends GoogleCredentials {
+
+    private static AccessToken newToken() {
+      return new AccessToken("owner",
+          new Date(System.currentTimeMillis() + TimeUnit.HOURS.toMillis(1)));
+    }
+
+    EmulatorCredentials() {
+      super(newToken());
+    }
+
+    @Override
+    public AccessToken refreshAccessToken() {
+      return newToken();
     }
   }
 }
--- a/src/main/java/com/google/firebase/FirebaseOptions.java
+++ b/src/main/java/com/google/firebase/FirebaseOptions.java
@@ -26,8 +26,9 @@ import com.google.api.client.util.Key;
 import com.google.auth.oauth2.GoogleCredentials;
 import com.google.cloud.firestore.FirestoreOptions;
 import com.google.common.base.Strings;
+import com.google.common.base.Supplier;
+import com.google.common.base.Suppliers;
 import com.google.common.collect.ImmutableList;
-import com.google.firebase.FirebaseOptions.DeferredApplicationCredentials.CredentialsGenerator;
 import com.google.firebase.internal.FirebaseThreadManagers;
 import com.google.firebase.internal.NonNull;
 import com.google.firebase.internal.Nullable;
@@ -59,7 +60,7 @@ public final class FirebaseOptions {
 
   private final String databaseUrl;
   private final String storageBucket;
-  private final DeferredApplicationCredentials deferredCredentials;
+  private final Supplier<GoogleCredentials> credentialsSupplier;
   private final Map<String, Object> databaseAuthVariableOverride;
   private final String projectId;
   private final String serviceAccountId;
@@ -72,17 +73,8 @@ public final class FirebaseOptions {
 
   private FirebaseOptions(@NonNull final FirebaseOptions.Builder builder) {
     this.databaseUrl = builder.databaseUrl;
-    if (builder.deferredCredentials != null) {
-      this.deferredCredentials = builder.deferredCredentials;
-    } else {
-      this.deferredCredentials = new DeferredApplicationCredentials(new CredentialsGenerator() {
-        @Override
-        public GoogleCredentials generate() {
-          return checkNotNull(builder.credentials,
-              "FirebaseOptions must be initialized with setCredentials().");
-        }
-      });
-    }
+    this.credentialsSupplier = checkNotNull(
+        builder.credentialsSupplier, "FirebaseOptions must be initialized with setCredentials().");
     this.databaseAuthVariableOverride = builder.databaseAuthVariableOverride;
     this.projectId = builder.projectId;
     if (!Strings.isNullOrEmpty(builder.storageBucket)) {
@@ -127,7 +119,7 @@ public final class FirebaseOptions {
   }
 
   GoogleCredentials getCredentials() {
-    return deferredCredentials.getCredentials();
+    return credentialsSupplier.get();
   }
 
   /**
@@ -236,8 +228,7 @@ public final class FirebaseOptions {
 
     @Key("serviceAccountId")
     private String serviceAccountId;
-    private GoogleCredentials credentials;
-    private DeferredApplicationCredentials deferredCredentials;
+    private Supplier<GoogleCredentials> credentialsSupplier;
     private FirestoreOptions firestoreOptions;
     private HttpTransport httpTransport = Utils.getDefaultTransport();
     private JsonFactory jsonFactory = Utils.getDefaultJsonFactory();
@@ -257,7 +248,7 @@ public final class FirebaseOptions {
     public Builder(FirebaseOptions options) {
       databaseUrl = options.databaseUrl;
       storageBucket = options.storageBucket;
-      deferredCredentials = options.deferredCredentials;
+      credentialsSupplier = options.credentialsSupplier;
       databaseAuthVariableOverride = options.databaseAuthVariableOverride;
       projectId = options.projectId;
       httpTransport = options.httpTransport;
@@ -317,19 +308,19 @@ public final class FirebaseOptions {
      * @return This <code>Builder</code> instance is returned so subsequent calls can be chained.
      */
     public Builder setCredentials(GoogleCredentials credentials) {
-      this.credentials = checkNotNull(credentials);
+      this.credentialsSupplier = Suppliers
+          .ofInstance(checkNotNull(credentials).createScoped(FIREBASE_SCOPES));
       return this;
     }
 
     /**
-     * Sets the <code>DeferredApplicationCredentials</code> to use to authenticate the SDK. This is
-     * NOT intended for public use outside the SDK.
+     * Sets the <code>Supplier</code> of <code>GoogleCredentials</code> to use to authenticate the SDK. This is NOT intended for public use outside the SDK.
      *
-     * @param credentials DeferredApplicationCredentials instance that wraps GoogleCredentials.
+     * @param credentialsSupplier Supplier instance that wraps GoogleCredentials.
      * @return This <code>Builder</code> instance is returned so subsequent calls can be chained.
      */
-    Builder setDeferredCredentials(DeferredApplicationCredentials credentials) {
-      this.deferredCredentials = checkNotNull(credentials);
+    Builder setCredentials(Supplier<GoogleCredentials> credentialsSupplier) {
+      this.credentialsSupplier = checkNotNull(credentialsSupplier);
       return this;
     }
     /**
@@ -473,29 +464,6 @@ public final class FirebaseOptions {
      */
     public FirebaseOptions build() {
       return new FirebaseOptions(this);
-    }
-  }
-
-  static class DeferredApplicationCredentials {
-
-    private final CredentialsGenerator credentialsGenerator;
-    private GoogleCredentials googleCredentials = null;
-
-    DeferredApplicationCredentials(
-        CredentialsGenerator credentialsGenerator) {
-      this.credentialsGenerator = credentialsGenerator;
-    }
-
-    public GoogleCredentials getCredentials() {
-      if (googleCredentials == null) {
-        googleCredentials = credentialsGenerator.generate().createScoped(FIREBASE_SCOPES);
-      }
-      return googleCredentials;
-    }
-
-    public interface CredentialsGenerator {
-
-      GoogleCredentials generate();
     }
   }
 }
--- a/src/main/java/com/google/firebase/FirebaseOptions.java
+++ b/src/main/java/com/google/firebase/FirebaseOptions.java
@@ -64,7 +64,7 @@ public final class FirebaseOptions {
         @Override
         public GoogleCredentials get() {
           try {
-            return GoogleCredentials.getApplicationDefault();
+            return GoogleCredentials.getApplicationDefault().createScoped(FIREBASE_SCOPES);
           } catch (IOException e) {
             throw new IllegalStateException(e);
           }
--- a/src/main/java/com/google/firebase/FirebaseOptions.java
+++ b/src/main/java/com/google/firebase/FirebaseOptions.java
@@ -314,7 +314,8 @@ public final class FirebaseOptions {
     }
 
     /**
-     * Sets the <code>Supplier</code> of <code>GoogleCredentials</code> to use to authenticate the SDK. This is NOT intended for public use outside the SDK.
+     * Sets the <code>Supplier</code> of <code>GoogleCredentials</code> to use to authenticate the
+     * SDK. This is NOT intended for public use outside the SDK.
      *
      * @param credentialsSupplier Supplier instance that wraps GoogleCredentials.
      * @return This <code>Builder</code> instance is returned so subsequent calls can be chained.
@@ -323,6 +324,7 @@ public final class FirebaseOptions {
       this.credentialsSupplier = checkNotNull(credentialsSupplier);
       return this;
     }
+
     /**
      * Sets the <code>auth</code> variable to be used by the Realtime Database rules.
      *
--- a/src/main/java/com/google/firebase/FirebaseOptions.java
+++ b/src/main/java/com/google/firebase/FirebaseOptions.java
@@ -27,6 +27,7 @@ import com.google.auth.oauth2.GoogleCredentials;
 import com.google.cloud.firestore.FirestoreOptions;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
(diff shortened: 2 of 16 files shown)


src/main/java/com/google/firebase/database/FirebaseDatabase.java
```diff
@@ -45,6 +46,8 @@
  */
 public class FirebaseDatabase {
 
+  static final String FIREBASE_RTDB_EMULATOR_HOST_ENV_VAR = "FIREBASE_RTDB_EMULATOR_HOST";
+  static final String RTDB_EMULATOR_HOST = "localhost:9000";
```
hiranya911: I think the env variable should contain the emulator host URL -- not just a `true` value.

@schmidt-sebastian can you also take a look?

yuchenshi: @sarajmunjal is working on it. Please consider this PR WIP until we remove the WIP tag and re-request review.

sarajmunjal: Done

src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -597,7 +597,18 @@ private static FirebaseOptions getOptionsFromEnvironment() throws IOException {
       parser = jsonFactory.createJsonParser(reader);
     }
     parser.parseAndClose(builder);
-    builder.setCredentials(GoogleCredentials.getApplicationDefault());
+    builder.setCredentials(getApplicationDefaultOrEmptyCredentials());
     return builder.build();
   }
+
```
sarajmunjal: This is done to enforce that we don't error out, but proceed with empty credentials if we were unable to fetch default credentials.

src/main/java/com/google/firebase/database/connection/NettyWebSocketClient.java
```diff
@@ -57,38 +60,51 @@
...
   private final ExecutorService executorService;
   private final EventLoopGroup group;
+  private final boolean isSecure;
 
   private Channel channel;
 
   NettyWebSocketClient(
       URI uri, String userAgent, ThreadFactory threadFactory,
-      WebsocketConnection.WSClientEventHandler eventHandler) {
+      WebsocketConnection.WSClientEventHandler eventHandler, boolean isSecure) {
```
sarajmunjal: Had to make this change to pass in `isSecure` attach `SSLContext` to the `WebSocketConnection` only if it is true. Earlier, all WS connections were assumed to be secure.

src/test/java/com/google/firebase/database/FirebaseDatabaseTest.java
```diff
@@ -44,13 +48,13 @@
           .setDatabaseUrl("https://firebase-db-test.firebaseio.com")
           .build();
 
-  @BeforeClass
-  public static void setupClass() {
+  @Before
```
sarajmunjal: Made this change as many test methods in this class require a default FirebaseApp instance that is instantiated only once: this conflicted with my tests where I need to create a default instance and then clean up. So I changed this up to be on a per-method basis which is cleaner and prevents methods from conflicting.

src/test/java/com/google/firebase/testing/TestUtils.java
```diff
@@ -85,6 +86,30 @@ public static void setEnvironmentVariables(Map<String, String> vars) {
     }
   }
 
+  public static void unsetEnvironmentVariables(Set<String> vars) {
```
sarajmunjal: This is the same way that we set env variables in tests. We should use this in a lot of other places too, else environment variables can get carried over between test methods.
Currently I use this only for my new tests.

src/main/java/com/google/firebase/database/FirebaseDatabase.java
```diff
@@ -101,7 +103,7 @@ public static FirebaseDatabase getInstance(FirebaseApp app) {
 
   /**
    * Gets a FirebaseDatabase instance for the specified URL, using the specified FirebaseApp.
-   *
+   getInstance*
```
schmidt-sebastian: Accident?

src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -597,7 +597,18 @@ private static FirebaseOptions getOptionsFromEnvironment() throws IOException {
...
+  private static GoogleCredentials getApplicationDefaultOrEmptyCredentials() {
+    try {
+      return GoogleCredentials.getApplicationDefault();
+    } catch (IOException e) {
+      logger.error(
+          "Failed to fetch default application credentials. Proceeding with blank credentials.",
```
schmidt-sebastian: Nit: Prefer "empty" over "blank". Your call.

src/main/java/com/google/firebase/FirebaseOptions.java
```diff
@@ -70,10 +74,16 @@
   private final FirestoreOptions firestoreOptions;
 
   private FirebaseOptions(@NonNull FirebaseOptions.Builder builder) {
-    this.credentials = checkNotNull(builder.credentials,
-        "FirebaseOptions must be initialized with setCredentials().")
-        .createScoped(FIREBASE_SCOPES);
-    this.databaseUrl = builder.databaseUrl;
+    String emulatorUrl = EmulatorHelper.getEmulatorUrl(builder.databaseUrl);
```
schmidt-sebastian: Do we need to make this check twice (once here and once in FirebaseDatabase)? It seems like FirebaseDatabase could make this check and overwrite any existing settings.

src/main/java/com/google/firebase/database/connection/NettyWebSocketClient.java
schmidt-sebastian: Nit: Please make this the second argument. It makes for much more readable code if a consumer passes an inline lamdba into `NettyWebSocketClient`.

src/main/java/com/google/firebase/database/connection/NettyWebSocketClient.java
```diff
@@ -57,38 +60,51 @@
...
+    if (this.isSecure) {
+      TrustManagerFactory trustFactory = null;
+      try {
+        trustFactory = TrustManagerFactory.getInstance(
+            TrustManagerFactory.getDefaultAlgorithm());
+        trustFactory.init((KeyStore) null);
+        sslContext = SslContextBuilder.forClient()
+            .trustManager(trustFactory).build();
+      } catch (NoSuchAlgorithmException | KeyStoreException | SSLException e) {
+        eventHandler.onError(e.getCause());
```
schmidt-sebastian: Can you change this back to the original structure (one big try/catch block surrounding everything)? Right now, if the SSL initialization fails, the code below still executes.

src/main/java/com/google/firebase/database/connection/NettyWebSocketClient.java
```diff
@@ -57,38 +60,51 @@
...
+    }
     try {
-      TrustManagerFactory trustFactory = TrustManagerFactory.getInstance(
-          TrustManagerFactory.getDefaultAlgorithm());
-      trustFactory.init((KeyStore) null);
-      final SslContext sslContext = SslContextBuilder.forClient()
-          .trustManager(trustFactory).build();
       Bootstrap bootstrap = new Bootstrap();
       final int port = uri.getPort() != -1 ? uri.getPort() : DEFAULT_WSS_PORT;
+      final SslContext finalSslContext = sslContext;
```
schmidt-sebastian: While super ugly, it is generally more comment to use a one element array for this:

```
  final SslContext[] sslContext = new SslContext[] {null};
  sslContext[0] = new ...
```

src/main/java/com/google/firebase/database/util/EmulatorHelper.java
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  @VisibleForTesting
+  public static final String FIREBASE_RTDB_EMULATOR_HOST_ENV_VAR = "FIREBASE_RTDB_EMULATOR_HOST";
+
+  private static String getEmulatorHostFromEnv() {
+    return System.getenv(FIREBASE_RTDB_EMULATOR_HOST_ENV_VAR);
+  }
+
+  public static ParsedUrl parsedUrlForEmulator(String dbName, String emulatorUrl) {
```
schmidt-sebastian: This is the `emulatorHost` and `namespace`, correct? A database URL can also contain a path, so we should not use that terminology.

src/main/java/com/google/firebase/database/util/EmulatorHelper.java
```diff
@@ -0,0 +1,71 @@
...
+    repoInfo.namespace = dbName;
+    repoInfo.secure = false;
+    ParsedUrl parsedUrl = new ParsedUrl();
+    parsedUrl.repoInfo = repoInfo;
+    parsedUrl.path = new Path("");
+    return parsedUrl;
+  }
+
+  @VisibleForTesting
+  static String extractEmulatorUrlFromDbUrl(String suppliedDatabaseUrl) {
```
schmidt-sebastian: Mark function `@Nullable`.

src/main/java/com/google/firebase/database/util/EmulatorHelper.java
```diff
@@ -0,0 +1,71 @@
...
+    if (Strings.isNullOrEmpty(suppliedDatabaseUrl)) {
+      return null;
+    }
+    ParsedUrl parsedUrl = Utilities.parseUrl(suppliedDatabaseUrl);
+    RepoInfo repoInfo = parsedUrl.repoInfo;
+    if (repoInfo.isSecure() || repoInfo.host.endsWith(".firebaseio.com") || !suppliedDatabaseUrl
+        .contains("ns=")) {
```
schmidt-sebastian: I don't quite follow why we disallow namespaces here. If you are separating parsing of the environment variable from a manually passed database URL, can you make this more clear (with a comment or a rename of the argument or method?). Thanks!

src/main/java/com/google/firebase/database/util/EmulatorHelper.java
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  @VisibleForTesting
+  static String extractEmulatorUrlFromDbUrl(String suppliedDatabaseUrl) {
+    if (Strings.isNullOrEmpty(suppliedDatabaseUrl)) {
+      return null;
+    }
+    ParsedUrl parsedUrl = Utilities.parseUrl(suppliedDatabaseUrl);
+    RepoInfo repoInfo = parsedUrl.repoInfo;
+    if (repoInfo.isSecure() || repoInfo.host.endsWith(".firebaseio.com") || !suppliedDatabaseUrl
```
schmidt-sebastian: Is there a reason we disallow SSL? I would expect that a user could set it up to work properly.

src/main/java/com/google/firebase/database/util/EmulatorHelper.java
```diff
@@ -0,0 +1,71 @@
...
+        .contains("ns=")) {
+      return null;
+    }
+    String pathString = parsedUrl.path.isEmpty() ? "/" : parsedUrl.path.toString() + "/";
+    return String.format("http://%s%s?ns=%s", repoInfo.host, pathString, repoInfo.namespace);
+  }
+
+  public static String getEmulatorUrl(String suppliedDatabaseUrl) {
```
schmidt-sebastian: A better name might be `getEmulatorHost`.

src/main/java/com/google/firebase/database/util/EmulatorHelper.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+    String namespaceName = "default";
+    String path = "/";
+    if (!Strings.isNullOrEmpty(suppliedDatabaseUrl)) {
+      ParsedUrl parsedUrl = Utilities.parseUrl(suppliedDatabaseUrl);
+      namespaceName = parsedUrl.repoInfo.namespace;
+      path = parsedUrl.path.isEmpty() ? "/" : parsedUrl.path.toString() + "/";
+    }
+    // Must format correctly
+    return String.format("http://%s%s?ns=%s", emulatorHostFromEnv, path, namespaceName);
```
schmidt-sebastian: Combining some part of the argument with an environment variable makes for a very surprising API contract.  I think this would look cleaner if we also passed the contents of the environment variable as an argument. We could also change the name and include `overwrite` or `replace` somewhere (just spitballing: overwriteDatabaseUrlWithEmulatorHost)

src/main/java/com/google/firebase/database/utilities/Utilities.java
```diff
@@ -18,78 +18,117 @@
...
+    }
+  }
 
-    } catch (URISyntaxException e) {
-      throw new DatabaseException("Invalid Firebase Database url specified", e);
-    } catch (UnsupportedEncodingException e) {
-      throw new DatabaseException("Failed to URLEncode the path", e);
+  /**
+   * Extracts the path string from the original URL without changing the encoding (unlike
+   * Uri.getPath()).
```
schmidt-sebastian: I used Android's URI parses in the original implementation. Did you have a chance to verify that this is also needed with `java.net.URI`?

src/main/java/com/google/firebase/database/utilities/Utilities.java
```diff
@@ -18,78 +18,117 @@
...
+    String[] paramPairs = queryString.split("&");
+    for (String paramPair : paramPairs) {
+      String[] pairParts = paramPair.split("=");
+      String value = paramsMap.get(pairParts[0]);
+      if (Strings.isNullOrEmpty(value)) {
+        value = pairParts[1];
+      } else {
+        value += "," + pairParts[1];
+      }
+      paramsMap.put(pairParts[0], value);
```
schmidt-sebastian: Do we need to handle any escaping here, or does URI.getQuery() handle this for us?

src/test/java/com/google/firebase/database/util/EmulatorHelperTest.java
```diff
@@ -0,0 +1,144 @@
...
+        "http://localhost:8080?ns=test-ns",
+        "localhost"
+    );
+    for (String invalidEnvVar : invalidEnvVars) {
+      TestUtils.setEnvironmentVariables(
+          ImmutableMap.of(EmulatorHelper.FIREBASE_RTDB_EMULATOR_HOST_ENV_VAR,
+              Strings.nullToEmpty(invalidEnvVar)));
+      thrown.expect(IllegalArgumentException.class);
+      EmulatorHelper.getEmulatorUrl("https://valid-namespace.firebaseio.com");
+      TestUtils.unsetEnvironmentVariables(
```
schmidt-sebastian: Wrap this in a try/finally so that `unsetEnvironmentVariables` is called.

src/main/java/com/google/firebase/FirebaseOptions.java
hiranya911: +1 for doing this in `FirebaseDatabase`

src/main/java/com/google/firebase/FirebaseOptions.java
```diff
@@ -455,4 +465,21 @@ public FirebaseOptions build() {
       return new FirebaseOptions(this);
     }
   }
+
+  private static class EmulatorCredentials extends GoogleCredentials {
```
hiranya911: Lets move this into the `database` package as well.

src/main/java/com/google/firebase/database/util/EmulatorHelper.java
```diff
@@ -0,0 +1,71 @@
+package com.google.firebase.database.util;
```
hiranya911: License header

src/main/java/com/google/firebase/FirebaseApp.java
hiranya911: Most of the time it is an error to deploy the SDK where ADC is not available. I think emulator is probably going to be the only exception. Can we also check if the emulator is set up (perhaps by checking the env variable), and if not retain the existing behavior?

src/main/java/com/google/firebase/database/util/EmulatorHelper.java
```diff
@@ -0,0 +1,71 @@
...
+
+public final class EmulatorHelper {
+
+  private EmulatorHelper() {
+  }
+
+  @VisibleForTesting
+  public static final String FIREBASE_RTDB_EMULATOR_HOST_ENV_VAR = "FIREBASE_RTDB_EMULATOR_HOST";
+
+  private static String getEmulatorHostFromEnv() {
```
hiranya911: Nit: Move to the end of file, after all public methods.

src/main/java/com/google/firebase/FirebaseApp.java
yuchenshi: The behavior in this PR is consistent with the Python Admin SDK and with the approved internal design doc. To reiterate, the reason behind this is 1) Developers may use RTDB config (databaseUrl) instead of environment variables. It is impossible to detect that here, and 2) poking into product-specific config / env vars in FirebaseApp introduces circular dependencies and in general make the code more messy.

However, we should make sure it is still an error to run the SDK when ADC is not available. It does not have to throw an exception here -- a delayed exception when we actually try to connect to RTDB/Firestore/Storage/etc. is also acceptable per the design doc. @sarajmunjal , is this already the case? If so, please add comments / tests to clarify. If not, please make it so.

src/main/java/com/google/firebase/database/util/EmulatorHelper.java
yuchenshi: Is this even used? I don't see it anywhere in this PR.

src/main/java/com/google/firebase/database/util/EmulatorHelper.java
yuchenshi: In theory a user can set up a reverse proxy that talks to the RTDB emulator. In practice, I cannot imagine many users doing this since we explicitly say that the emulator is not for production use, etc. I don't feel strongly about this as long as it is consistent either the Android SDK or the Python Admin SDK.

src/main/java/com/google/firebase/database/util/EmulatorHelper.java
yuchenshi: +1 also easier to test

src/main/java/com/google/firebase/database/util/EmulatorHelper.java
yuchenshi: But it returns a URL as seen below. `http://`...

src/main/java/com/google/firebase/database/utilities/Utilities.java
yuchenshi: I'm pretty sure we need to unescape in case the URL contains %-encoded characters. `ns` is URL safe so it doesn't break for our use case, but I still want it to be correct as a general utility.

src/test/java/com/google/firebase/database/FirebaseDatabaseTest.java
yuchenshi: Thanks for doing this!

src/test/java/com/google/firebase/database/utilities/UtilitiesTest.java
```diff
@@ -185,4 +186,11 @@ public void onComplete(DatabaseError error, DatabaseReference ref) {
     assertSame(listener, result.getSecond());
   }
 
+  @Test
+  public void testExtractParamsFromUrl() {
+    Map<String, String> params = Utilities.getQueryParamsMap("abc=213&qpf=2312&xyz=true&qpf=hi");
```
yuchenshi: Add tests for %-encoded values, as mentioned above.

src/test/java/com/google/firebase/testing/TestUtils.java
yuchenshi: This really sucks. Would you consider modifying the code to take env vars as a map in constructors, methods, etc. instead?

src/main/java/com/google/firebase/FirebaseApp.java
hiranya911: The behavior as implemented here is not consistent with the Python implementation. We don't have the deferred credential look up feature here. May be implement a new `DeferredApplicationDefaultCredentials` implementation which provides that deferred behavior.

src/main/java/com/google/firebase/database/utilities/Utilities.java
sarajmunjal: Good point @yuchenshi. But it turns out that URI.getQuery() does handle this for us, as demonstrated in a test case that I've added to `TestUtils.parseUrlValid`

So I've added `@VisibleForTesting` to this method, which should restrict its scope of usage as only a companion to `Utilities.parseUrl`

src/test/java/com/google/firebase/database/util/EmulatorHelperTest.java
sarajmunjal: Removed this code, so resolving the comment.

src/main/java/com/google/firebase/FirebaseApp.java
sarajmunjal: Done!

src/test/java/com/google/firebase/testing/TestUtils.java
sarajmunjal: Done!

src/main/java/com/google/firebase/FirebaseOptions.java
sarajmunjal: Done, seems cleaner.

src/main/java/com/google/firebase/database/connection/NettyWebSocketClient.java
sarajmunjal: Done

src/main/java/com/google/firebase/database/util/EmulatorHelper.java
sarajmunjal: Good catch again. Removed.

src/main/java/com/google/firebase/database/util/EmulatorHelper.java
sarajmunjal: Changed the method name to `tryToExtractEmulatorUrlFromDbUrl`. Suggestions welcome for other names.

src/main/java/com/google/firebase/database/util/EmulatorHelper.java
sarajmunjal: Do we have more discussions here?

src/main/java/com/google/firebase/database/util/EmulatorHelper.java
sarajmunjal: marking as resolved

src/main/java/com/google/firebase/database/utilities/Utilities.java
sarajmunjal: Great catch. As it turns out, we don't need this. `URI.getPath` auto decodes spaces and other chars.

src/main/java/com/google/firebase/FirebaseApp.java
sarajmunjal: Changed this line as we return null now, wrapped in `DeferredApplicationCredentials`. Feel free to suggest another change if required, but I feel the NPE will be self explanatory.

src/main/java/com/google/firebase/database/FirebaseDatabase.java
sarajmunjal: Yus :man_facepalming:

src/main/java/com/google/firebase/FirebaseOptions.java
```diff
@@ -69,11 +70,19 @@
   private final ThreadManager threadManager;
   private final FirestoreOptions firestoreOptions;
```
sarajmunjal: I've tried not to break the public API for `FirebaseOptions.Builder`: so it supports both `setDeferredCredentials` (which is package private, and can't be used publicly) and the old `setCredentials` method. This constructor accepts `defaultCredentials` if not null, else falls back to creating a custom `DefaultApplicationCredentials` wrapping the regular `credentials`.

Suggestions welcome for doing this in a better way.

src/main/java/com/google/firebase/database/util/EmulatorHelper.java
schmidt-sebastian: The Android SDK supports talking to `https://localhost:8080`. I don't think we need to go out of our way to support SSL, but it seems non-ideal to write code that prevents SSL usage.

src/main/java/com/google/firebase/database/utilities/Utilities.java
schmidt-sebastian: Please note that on Android, if you call `FirebaseDatabase.getReference("http://localhost/foo%20bar")` we will actually try to get a reference at `foo%20bar` and not `foo bar` (on iOS, we use a space). This is certainly not the ideal behavior, but we decided to maintain backwards compatibility.

src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -597,7 +598,22 @@ private static FirebaseOptions getOptionsFromEnvironment() throws IOException {
...
     parser.parseAndClose(builder);
-    builder.setCredentials(GoogleCredentials.getApplicationDefault());
+    builder.setDeferredCredentials(DEFAULT_OR_EMPTY_CREDS);
     return builder.build();
   }
+
+  private static final DeferredApplicationCredentials DEFAULT_OR_EMPTY_CREDS =
+      new DeferredApplicationCredentials(new CredentialsGenerator() {
+        @Override
+        public GoogleCredentials generate() {
```
schmidt-sebastian: I think you could get rid of this interface and make `setDeferredCredentials` take a Supplier instead. When you do that, it would also make sense to me to rename it back to setCredentials (as an overload).

src/main/java/com/google/firebase/FirebaseOptions.java
schmidt-sebastian: As eluded to above, you essentially re-implemented the Supplier  pattern (available via com.google.common.base.Supplier in Java 7). I think this could just be:

```
Supplier<Credentials> credentials;

public void setCredentials(Credentials credentials) { this.credentials = Suppliers.ofInstance(credentials); }

void setCredentials(Supplier<Credentials> credentials) { this.credentials = credentials; }

```

src/main/java/com/google/firebase/database/FirebaseDatabase.java
```diff
@@ -112,14 +121,18 @@ public static synchronized FirebaseDatabase getInstance(FirebaseApp app, String
...
-
+    boolean connectingToEmulator = false;
+    String possibleEmulatorUrl = EmulatorHelper
+        .overwriteDatabaseUrlWithEmulatorHost(url, EmulatorHelper.getEmulatorHostFromEnv());
+    if (!Strings.isNullOrEmpty(possibleEmulatorUrl)) {
+      url = possibleEmulatorUrl;
+      connectingToEmulator = true;
+    }
     ParsedUrl parsedUrl = Utilities.parseUrl(url);
     if (!parsedUrl.path.isEmpty()) {
```
schmidt-sebastian: In `getInstance()`, we don't allow any custom paths. I think with that in mind we might not need to merge `url` and the environment variable? We would only need to return `repoInfo` based on the emulator host.

src/main/java/com/google/firebase/database/core/DatabaseConfig.java
```diff
@@ -16,8 +16,11 @@
 
 package com.google.firebase.database.core;
 
+import com.google.auth.oauth2.GoogleCredentials;
```
schmidt-sebastian: Revert changes in this file.

src/main/java/com/google/firebase/FirebaseOptions.java
```diff
@@ -455,4 +475,27 @@ public FirebaseOptions build() {
       return new FirebaseOptions(this);
     }
   }
+
+  static class DeferredApplicationCredentials {
```
hiranya911: I would expect this to directly extend the `GoogleCredentials` class. Something like:

```
// Simplified for illustration
class DeferredGoogleCredentials extends GoogleCredentials {

   private GoogleCredentials delegate;

   public AccessToken refreshAccessToken() {
       if (delegate == null) {
           delegate = GoogleCredentials.applicatioDefault();
       }
       return delegate.refreshAccessToken();
   }
}
```

Then in `FirebaseDatabase` just ignore the `GoogleCredentials` set in `FirebaseOptions` whenever connecting to the emulator.

hiranya911: Alternatively, you can use that `Supplier` pattern idea proposed by @schmidt-sebastian.  `FirebaseOptions` can store credentials as a supplier:

```
private Supplier<GoogleCredentials> credentials;

// Existing public API (signature unchanged)
public Builder setGoogleCredentials(GoogleCredentials credentials) {
   this.credentials = Suppliers.ofInstance(credentials);
}

// New internal API we will use when setting up ADC
Builder setGoogleCredentials(Supplier<GoogleCredentials> supplier) {
   this.credentials = supplier;
}
```

src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -597,7 +598,22 @@ private static FirebaseOptions getOptionsFromEnvironment() throws IOException {
...
+          try {
+            return GoogleCredentials.getApplicationDefault();
```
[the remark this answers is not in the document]
sarajmunjal: We do catch the only checked exception that it throws: `IOException`. In this case, we return null, and our supplier will throw a NPE when `getCredentials()` is called, not while creating `FirebaseOptions`.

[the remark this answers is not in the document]
yuchenshi: IMO this should just throw like before. Let's get rid of the try-catch since an IOException with details is much, much better than a NPE.

src/main/java/com/google/firebase/FirebaseOptions.java
yuchenshi: I like the supplier idea better. `GoogleCredentials` has way too many methods and overwriting all of them with delegate would be a lot of noise. I don't even know if it can be done cleanly since GoogleCredentials is a concrete and serializable class.

src/main/java/com/google/firebase/database/FirebaseDatabase.java
yuchenshi: FYI Python Admin SDK will strip out the path part from the URL.

src/main/java/com/google/firebase/database/connection/NettyWebSocketClient.java
```diff
@@ -75,20 +80,27 @@
...
-          TrustManagerFactory.getDefaultAlgorithm());
-      trustFactory.init((KeyStore) null);
-      final SslContext sslContext = SslContextBuilder.forClient()
-          .trustManager(trustFactory).build();
       Bootstrap bootstrap = new Bootstrap();
+      SslContext sslContext = null;
+      if (this.isSecure) {
+        TrustManagerFactory trustFactory = null;
+        trustFactory = TrustManagerFactory.getInstance(
+            TrustManagerFactory.getDefaultAlgorithm());
```
yuchenshi: Huh? I think we can just say ` TrustManagerFactory trustFactory = TrustManagerFactory.getInstance(...)` instead of `null` here. WDYT?

src/main/java/com/google/firebase/database/connection/NettyWebSocketClient.java
```diff
@@ -75,20 +80,27 @@
...
+        trustFactory.init((KeyStore) null);
+        sslContext = SslContextBuilder.forClient()
+            .trustManager(trustFactory).build();
+      }
       final int port = uri.getPort() != -1 ? uri.getPort() : DEFAULT_WSS_PORT;
+      final SslContext[] sslContexts = new SslContext[]{sslContext};
```
yuchenshi: Why are we doing this? Isn't a null-able variable sufficient?

src/main/java/com/google/firebase/database/utilities/Utilities.java
```diff
@@ -18,78 +18,124 @@
...
-    } catch (URISyntaxException e) {
-      throw new DatabaseException("Invalid Firebase Database url specified", e);
-    } catch (UnsupportedEncodingException e) {
-      throw new DatabaseException("Failed to URLEncode the path", e);
+  /**
+   * Extracts the path string from the original URL without changing the encoding (unlike
+   * Uri.getPath()).
+   */
+  private static String extractPathString(String originalUrl) {
+    int schemeOffset = originalUrl.indexOf("//");
```
yuchenshi: Again, where is this even used?

src/test/java/com/google/firebase/database/utilities/UtilitiesTest.java
```diff
@@ -185,4 +192,20 @@ public void onComplete(DatabaseError error, DatabaseReference ref) {
...
+    assertEquals("213", params.get("abc"));
+    assertEquals("2312,hi", params.get("qpf"));
+    assertEquals("true", params.get("xyz"));
+  }
+
+
+  @Test
+  public void testExtractParamsFromEncodedUrl() {
+    Map<String, String> params = Utilities.getQueryParamsMap("abc=213&qpf=2312&xyz=true&qpf=hi");
```
yuchenshi: Hey, isn't this the same as the one above?

yuchenshi: Also good stuff to test: `a=%3F%3F%3F&b=%3D%26%3D` should give you `???` for a and `=&=` for b. Also test this with the URI class and see if it does the right thing.

yuchenshi: As a extreme case, you should be able to put a full URL in a param if you encode it correctly :)

src/test/java/com/google/firebase/testing/TestUtils.java
yuchenshi: Not done? I still see we're messing with env vars in tests.

src/main/java/com/google/firebase/FirebaseOptions.java
hiranya911: `refreshAccessToken()` is the only method you have to override if you decide to go down that path. But, yeah, Supplier is probably a cleaner approach.

src/main/java/com/google/firebase/database/util/EmulatorHelper.java
sarajmunjal: Allowed this now.

src/main/java/com/google/firebase/database/utilities/Utilities.java
sarajmunjal: Changed it to `foo%20bar` to maintain backward compatibility.

src/main/java/com/google/firebase/database/FirebaseDatabase.java
sarajmunjal: Let's maintain backwards compatibility here and keep throwing the exception that was being thrown earlier.

src/main/java/com/google/firebase/database/core/DatabaseConfig.java
sarajmunjal: Done

src/main/java/com/google/firebase/FirebaseOptions.java
sarajmunjal: I went with the supplier approach

src/main/java/com/google/firebase/database/utilities/Utilities.java
sarajmunjal: Removed this unused method, it is not required.

src/test/java/com/google/firebase/database/utilities/UtilitiesTest.java
sarajmunjal: Thanks for this @yuchenshi. I switched up the behavior: `Utilities.getQueryParamsMap` now accepts an encoded query string and decodes each key-value pair separately.

src/main/java/com/google/firebase/database/connection/NettyWebSocketClient.java
sarajmunjal: I agree. Changing this back to `final SslContext finalSslContext = sslContext` (final-copy) pattern instead of the array. I think this way is cleaner.

src/main/java/com/google/firebase/FirebaseApp.java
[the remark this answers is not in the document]
sarajmunjal: We can't throw checked exceptions from suppliers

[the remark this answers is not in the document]
sarajmunjal: We just threw an unchecked exception from the custom supplier for this.

src/main/java/com/google/firebase/FirebaseApp.java
yuchenshi: I think with the updated code the behavior is now inline with the Python admin SDK. (And sorry I misunderstood you earlier -- your comment makes a lot of sense.)

src/test/java/com/google/firebase/testing/TestUtils.java
sarajmunjal: Yes we're still doing it in some tests which are sort of "integration" tests for FirebaseDatabase: here we must set the environment variable to test effectively :(

src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -90,6 +92,17 @@
    * accesses to instances map should be protected by this lock.
    */
   private static final Object appsLock = new Object();
+  private static final Supplier<GoogleCredentials> DEFAULT_CREDS_CALLABLE =
```
schmidt-sebastian: Nit: s/callable/supplier

src/main/java/com/google/firebase/database/FirebaseDatabase.java
```diff
@@ -140,14 +156,16 @@ public static synchronized FirebaseDatabase getInstance(FirebaseApp app, String
...
+        config.setCustomCredentials(new EmulatorCredentials(), true);
+      }
       database = new FirebaseDatabase(app, parsedUrl.repoInfo, config);
       dbInstances.put(parsedUrl.repoInfo, database);
     }
-
     return database;
   }
 
+
```
schmidt-sebastian: Nit: Remove empty line.

src/main/java/com/google/firebase/database/FirebaseDatabase.java
```diff
@@ -380,4 +403,31 @@ public void destroy() {
...
   }
+
+  private static class EmulatorCredentials extends GoogleCredentials {
+
+    private static AccessToken newToken() {
+      return new AccessToken("owner",
+          new Date(System.currentTimeMillis() + TimeUnit.HOURS.toMillis(1)));
+    }
+
+    EmulatorCredentials() {
```
schmidt-sebastian: Nit: Move constructor to the top.

src/main/java/com/google/firebase/database/FirebaseDatabase.java
```diff
@@ -380,4 +403,31 @@ public void destroy() {
...
+      return newToken();
+    }
+
+    @Override
+    public Map<String, List<String>> getRequestMetadata() throws IOException {
+      return ImmutableMap.of();
+    }
+
+    @Override
+    public void refresh() throws IOException {
```
schmidt-sebastian: This is not needed.

src/main/java/com/google/firebase/database/core/JvmAuthTokenProvider.java
```diff
@@ -40,8 +40,14 @@
     this(firebaseApp, executor, true);
   }
 
+
```
schmidt-sebastian: Nit: Remove empty line.

src/main/java/com/google/firebase/database/util/EmulatorHelper.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+    ParsedUrl parsedUrl = Utilities.parseUrl(suppliedDatabaseUrl);
+    RepoInfo repoInfo = parsedUrl.repoInfo;
+    if (repoInfo.host.endsWith(".firebaseio.com") || !suppliedDatabaseUrl
+        .contains("ns=")) {
+      return null;
+    }
+    String pathString = parsedUrl.path.isEmpty() ? "/" : parsedUrl.path.toString() + "/";
+    String scheme = repoInfo.isSecure() ? "https" : "http";
+    return String.format("%s://%s%s?ns=%s", scheme, repoInfo.host, pathString, repoInfo.namespace);
```
schmidt-sebastian: Could this method return a ParsedURL instead of constructing a new URL that we need to parse again?

src/main/java/com/google/firebase/database/util/EmulatorHelper.java
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  public static String overwriteDatabaseUrlWithEmulatorHost(String suppliedDatabaseUrl,
```
schmidt-sebastian: `@Nullable`

src/main/java/com/google/firebase/database/util/EmulatorHelper.java
```diff
@@ -0,0 +1,78 @@
...
+      String emulatorHost) {
+    String extractedEmulatorUrl = tryToExtractEmulatorUrlFromDbUrl(suppliedDatabaseUrl);
+    if (!Strings.isNullOrEmpty(extractedEmulatorUrl)) {
```
schmidt-sebastian: Something still feels a little weird about the control flow here, especially since you are still mixing the concepts of databaseUrl and emulatorHost in somewhat surprising ways.

What do you think about this (pseudocode):

```
public static ParsedUrl getEmulatorUrl(String suppliedDatabaseUrl) {
    ParsedUrl parsedUrl = Utilities.parseUrl(suppliedDatabaseUrl);
    if (isEmulatorUrl(parsedUrl)) {
		return parsedUrl;
	}
	
	String hostEnv = getEmulatorHostFromEnv();
	if (!hostEnv) {
		return null;
	}
	
    ParsedUrl parsedUrl = Utilities.parseUrl(hostEnv);
    if (parsedUrl.scheme("http:") || parsedUrl.path("?ns=")) {
      throw new IllegalArgumentException(
          "emulator host declared in environment variable must be of the format \"host:port\"");
    }
	
    String namespaceName = "default";
    String path = "/";
    if (!Strings.isNullOrEmpty(suppliedDatabaseUrl)) {
      ParsedUrl parsedUrl = Utilities.parseUrl(suppliedDatabaseUrl);
      namespaceName = parsedUrl.repoInfo.namespace;
      path = parsedUrl.path.isEmpty() ? "/" : parsedUrl.path.toString() + "/";
    }
	
    return new ParsedUrl(...)
  }
```

src/main/java/com/google/firebase/database/utilities/Utilities.java
```diff
@@ -18,78 +18,106 @@
...
-        repoInfo.secure = true;
+        String[] parts = host.split("\\.", -1);
+        repoInfo.namespace = parts[0].toLowerCase();
       }
-      String[] parts = repoInfo.host.split("\\.");
 
-      repoInfo.namespace = parts[0].toLowerCase();
       repoInfo.internalHost = repoInfo.host;
+      // use raw (encoded) path for backwards compatibility.
+      String originalPathString = uri.getRawPath();
```
schmidt-sebastian: Oh. I didn't know about this method. Awesome!

Note that you still have to turn `+` into spaces: https://github.com/firebase/firebase-android-sdk/pull/680/files#diff-76637acaa08127d536bbbe0c038845fdR72

src/test/java/com/google/firebase/database/FirebaseDatabaseTest.java
```diff
@@ -44,13 +48,13 @@
...
+  public void setupTestMethod() {
     FirebaseApp.initializeApp(firebaseOptions);
```
schmidt-sebastian: If you remove this line and instead move it to the test cases that need it, can you get rid of the special logic in the new tests that detects whether the default app exists?

src/test/java/com/google/firebase/database/FirebaseDatabaseTest.java
```diff
@@ -171,4 +174,135 @@ public void testInitAfterAppDelete() throws ExecutionException, InterruptedExcep
...
+        // and append it to the emulator URL from env var(if it is valid)
+        new CustomTestCase("https://valid-namespace.firebaseio.com", "localhost:8080",
+            "http://localhost:8080", "valid-namespace"),
+        new CustomTestCase("https://test.firebaseio.com?ns=valid-namespace", "localhost:90",
+            "http://localhost:90", "valid-namespace")
+    );
+    boolean earlierDefaultAppFound = false;
+    try {
+      FirebaseApp.getInstance(FirebaseApp.DEFAULT_APP_NAME).delete();
+      earlierDefaultAppFound = true;
```
schmidt-sebastian: We might be able to remove this if you remove the `@Before` method and inline it to the tests that need it.

src/test/java/com/google/firebase/database/FirebaseDatabaseTest.java
```diff
@@ -171,4 +174,135 @@ public void testInitAfterAppDelete() throws ExecutionException, InterruptedExcep
...
+    }
+    if (earlierDefaultAppFound) {
+      FirebaseApp.initializeApp();
+    }
+  }
+
+
+  @Test
+  public void testDbUrlIsEmulatorUrlForDbRefWithPath() {
+    class CustomTestCase {
```
schmidt-sebastian: Can we make this a shared (static) class between the two test cases?

src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -90,6 +92,17 @@
...
+      new Supplier<GoogleCredentials>() {
+        @Override
+        public GoogleCredentials get() {
+          try {
+            return GoogleCredentials.getApplicationDefault();
```
hiranya911: I think this needs to be scoped too. Perhaps move this to `FirebaseOptions` with package-level visibility, and use it here as `FirebaseOptions.APPLICATION_DEFAULT_CREDENTIALS`.

src/main/java/com/google/firebase/database/util/EmulatorHelper.java
sarajmunjal: Done

src/main/java/com/google/firebase/database/util/EmulatorHelper.java
sarajmunjal: Thanks! Makes things much cleaner.
One caveat though: `isEmulatorUrl(String url)` has to take a string URL instead of `ParsedUrl`. This is because we need to check for the existence of the namespace query parameter.

src/main/java/com/google/firebase/database/utilities/Utilities.java
sarajmunjal: I don't think that we need to replace `+` with space: uri.getRawPath() doesn't seem to be outputting + instead of spaces. We have testcases to verify this.

src/test/java/com/google/firebase/database/FirebaseDatabaseTest.java
sarajmunjal: This adds boilerplate to a lot of tests, but makes the development experience much better while modifying tests. :+1:

src/test/java/com/google/firebase/database/FirebaseDatabaseTest.java
sarajmunjal: Done

src/main/java/com/google/firebase/FirebaseApp.java
sarajmunjal: Done

src/test/java/com/google/firebase/database/FirebaseDatabaseTest.java
sarajmunjal: Done

src/main/java/com/google/firebase/database/utilities/Utilities.java
yuchenshi: Yo, here are some test cases I think we should add: https://github.com/firebase/firebase-android-sdk/pull/680/files#diff-7aea264732c9130ecc3212feae077da0R53-R64
Maybe it will even show some inconsistencies -- let's see.

src/main/java/com/google/firebase/FirebaseOptions.java
```diff
@@ -56,9 +59,21 @@
           "https://www.googleapis.com/auth/cloud-platform",
           "https://www.googleapis.com/auth/datastore");
 
+  static final Supplier<GoogleCredentials> APPLICATION_DEFAULT_CREDENTIALS =
+      new Supplier<GoogleCredentials>() {
+        @Override
+        public GoogleCredentials get() {
+          try {
+            return GoogleCredentials.getApplicationDefault();
```
hiranya911: `.createScoped(FIREBASE_SCOPES)`

src/main/java/com/google/firebase/database/utilities/Utilities.java
sarajmunjal: I added a test case for `+` being converted to ` `, but the non-ASCII characters testcase blew up the Java URI class.

src/test/java/com/google/firebase/database/FirebaseDatabaseTest.java
```diff
@@ -159,16 +177,121 @@ public void testAppDelete() throws ExecutionException, InterruptedException {
...
+    } finally {
+      TestOnlyImplFirebaseTrampolines.clearInstancesForTest();
+    }
+  }
+
+  @Test
+  public void testDbUrlIsEmulatorUrlWhenSettingOptionsManually() {
+
+    List<CustomTestCase> testCases;
+    testCases = ImmutableList.of(
```
schmidt-sebastian: Nit: Combine 199&200 (here and below)

PR #321, opened 2019-10-22
fix(fcm): Increase the multicast request count limit to 500.

chong-shao:
Increase the multicast request count limit to 500.

### Testing

  * Updated the related tests.

RELEASE NOTE: Number of messages per batch increased to 500.

### DIFF
--- a/src/main/java/com/google/firebase/messaging/FirebaseMessaging.java
+++ b/src/main/java/com/google/firebase/messaging/FirebaseMessaging.java
@@ -149,7 +149,7 @@ public class FirebaseMessaging {
    * <p>The responses list obtained by calling {@link BatchResponse#getResponses()} on the return
    * value corresponds to the order of input messages.
    *
-   * @param messages A non-null, non-empty list containing up to 100 messages.
+   * @param messages A non-null, non-empty list containing up to 500 messages.
    * @return A {@link BatchResponse} indicating the result of the operation.
    * @throws FirebaseMessagingException If an error occurs while handing the messages off to FCM for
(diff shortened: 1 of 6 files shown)


Bonythomasv:
@chong-shao is n't it better to make this configurable ?
my understanding is that , we can send upto 1000 devices with a single API call. so making this configurable will provide more flexibility for the clients.

chong-shao:
Hi @Bonythomasv  We are experimenting with the 1000 limit and we would like to set the recommended maximum size in our SDK to be 500 for now. 

> @chong-shao is n't it better to make this configurable ?
> my understanding is that , we can send upto 1000 devices with a single API call. so making this configurable will provide more flexibility for the clients.

Bonythomasv:
@chong-sao. Are you seeing any knows issues with 1000 ?

We used the FCM http directly, we send 100s services per API call and but we would like to switch to admin sdk and this 500 hard limit is going to be a problem for us. 

Can I PR to make this configurable with default value as 500

hiranya911:
FCM v1 API only supported 100 messages per batch until a few days ago. It's the legacy FCM API that supported batches of up to 1000. We want to get the v1 API limit bumped up to 1000, but we want to do it in phases. So until all SDKs are there, we'd advise users to not send batches larger than 500 to the v1 API.

Bonythomasv:
Ah.. I see , we still uses the legacy API.

One side question,
The does v1 API response has the API message delivery status, or is there
any way to find number of failed devices by multicast_id or message_id in
the response.

PR #324, opened 2019-10-29
Fix a typo in AndroidNotification

lahirumaramba:
- Fix a typo in `AndroidNotification`

### DIFF
--- a/src/main/java/com/google/firebase/messaging/AndroidNotification.java
+++ b/src/main/java/com/google/firebase/messaging/AndroidNotification.java
@@ -443,7 +443,7 @@ public class AndroidNotification {
     /**
      * For notifications that inform users about events with an absolute time reference, sets
      * the time that the event in the notification occurred in milliseconds. Notifications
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/firebase/messaging/AndroidNotification.java
```diff
@@ -443,7 +443,7 @@ public Builder setSticky(boolean sticky) {
     /**
      * For notifications that inform users about events with an absolute time reference, sets
      * the time that the event in the notification occurred in milliseconds. Notifications
-     * in the panel are sorted by this time. The time is be formated in RFC3339 UTC "Zulu" 
+     * in the panel are sorted by this time. The time will be formatted in RFC3339 UTC "Zulu"
```
chong-shao: Good catch and thanks for doing this! Could you change it to "is formatted"? egilmore@ looked at this before and suggested to avoid "will."

lahirumaramba: Thanks, Chong. Updated!

PR #326, opened 2019-11-01
Reject rounds=0 for SHA1 hashes

rsgowman:
Port of https://github.com/firebase/firebase-admin-node/pull/677

### DIFF
--- a/src/test/java/com/google/firebase/auth/UserImportHashTest.java
+++ b/src/test/java/com/google/firebase/auth/UserImportHashTest.java
@@ -17,7 +17,6 @@
 package com.google.firebase.auth;
 
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.fail;
 
 import com.google.common.collect.ImmutableMap;
 import com.google.common.io.BaseEncoding;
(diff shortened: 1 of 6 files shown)


src/test/java/com/google/firebase/auth/UserImportHashTest.java
```diff
@@ -145,6 +146,91 @@ public void testBasicHash() {
     }
   }
 
+  private void assertBuilderThrowsIllegalArgumentException(Md5.Builder builder) {
```
rsgowman: An alternative to these helpers is to make all of the builders extend from a common interface (which they do: RepeatableHash.Builder) and make it public (which it isn't; and neither is RepeatableHash) even if it's just `@VisibleForTesting`. I didn't do that, in the (naively?) optimistic hope that we'll update to java8 "soon" which will make all that unnecessary.

hiranya911: May be move this to `InvalidHashTest`. There `RepeatableHash` is already visible.

src/test/java/com/google/firebase/auth/UserImportHashTest.java
```diff
@@ -145,6 +146,91 @@ public void testBasicHash() {
...
+    try {
+      builder.build();
+      fail("Expected IllegalArgumentException to be thrown but no expecption occurred.");
+    } catch (IllegalArgumentException expected) {
+    }
+  }
+
+  private void assertBuilderThrowsIllegalArgumentException(Sha1.Builder builder) {
```
hiranya911: If you decide to keep these helpers, move them further down in the file, after all the public methods.

src/test/java/com/google/firebase/auth/UserImportHashTest.java
rsgowman: Oh, yeah; that's definitely better. (Despite the name, I also used this file to test for valid hashes too.)

src/test/java/com/google/firebase/auth/UserImportHashTest.java
rsgowman: (no longer applicable.)

PR #327, opened 2019-11-06
Fix Date test in another Locale

sakebook:
### Discussion
No issue.

Current Date test is depend on Locale(GMT).

I ran the test on my local machine and it was not successful.
This is because the Japanese locale (JMT) is used.

I defined use Locale.US.

### Testing
Fixed test

### API Changes
None

### DIFF
--- a/src/main/java/com/google/firebase/internal/DateUtils.java
+++ b/src/main/java/com/google/firebase/internal/DateUtils.java
@@ -20,6 +20,7 @@ import java.text.ParsePosition;
 import java.text.SimpleDateFormat;
 import java.util.Calendar;
 import java.util.Date;
+import java.util.Locale;
 import java.util.TimeZone;
 
 /**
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/firebase/internal/DateUtils.java
```diff
@@ -57,9 +58,9 @@
   static final String PATTERN_ASCTIME = "EEE MMM d HH:mm:ss yyyy";
 
   private static final SimpleDateFormat[] DEFAULT_PATTERNS = new SimpleDateFormat[] {
-      new SimpleDateFormat(PATTERN_RFC1123),
-      new SimpleDateFormat(PATTERN_RFC1036),
-      new SimpleDateFormat(PATTERN_ASCTIME)
+      new SimpleDateFormat(PATTERN_RFC1123, Locale.UK),
```
hiranya911: Let's follow what http://svn.apache.org/repos/asf/httpcomponents/httpclient/tags/4.3/httpclient/src/main/java/org/apache/http/client/utils/DateUtils.java does and set to `Locale.US`.

sakebook: I fixed it. https://github.com/firebase/firebase-admin-java/pull/327/commits/3e8ec14402afd88a6ad42b89efdbf2bd47d1e6b5

PR #338, opened 2019-12-19
fix(fcm): Fix issues in event_time timestamp parsing

### LINKED ISSUE #337: Incorrect timestamp format in AndroidNotification.event_time
I think the way we format the `event_time` field on `AndroidNotification` is wrong:

https://github.com/firebase/firebase-admin-java/blob/83b7a64c103fbf20f46c04df229000529e9a8d2d/src/main/java/com/google/firebase/messaging/AndroidNotification.java#L458-L462

There seems to be couple of issues with this.

1. We are not specifying the timezone for the `SimpleDateFormat`. Therefore the resulting timestamp string is in the local timezone while the backend expects it to be in UTC.
2. The way nanoseconds are formatted seems to be wrong. For the milliseconds value `1546304523123L` this generates the string `2019-01-01T01:02:03.000000123Z`. I would expect the nanos to be formatted as `123000000` (i.e. 123 milliseconds in nanos).

chong-shao:
- Set the timezone to always be UTC.
- Pending zeros after the millisecond digits to the right instead of left.

RELEASE NOTE: `AndroidNotification` class now correctly formats the `event_time` field sent to the Cloud Messaging service.

Resolves #337

### DIFF
--- a/src/main/java/com/google/firebase/messaging/AndroidNotification.java
+++ b/src/main/java/com/google/firebase/messaging/AndroidNotification.java
@@ -28,6 +28,7 @@ import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
+import java.util.TimeZone;
 import java.util.concurrent.TimeUnit;
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/firebase/messaging/AndroidNotification.java
```diff
@@ -456,8 +457,9 @@ public Builder setSticky(boolean sticky) {
      * @return This builder.
      */
     public Builder setEventTimeInMillis(long eventTimeInMillis) {
-      this.eventTime = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSSSSSSS'Z'")
-          .format(new Date(eventTimeInMillis));
+      SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS000000'Z'");
+      dateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
+      this.eventTime = dateFormat.format(new Date(1546304523123L));
```
hiranya911: Format eventTimeInMillis

chong-shao: Oops, sorry missed it

PR #342, opened 2020-01-08
Setting up a CI pipeline based on GitHub Actions

hiranya911:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: Java CI
+name: Java Continuous Integration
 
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,17 @@
...
+
+on: [push, pull_request]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
```
lahirumaramba: Just curious, what is the reason to use `checkout@v1` instead of `v2`?

PR #334, opened 2019-12-12
Fixing some failing RTDB tests

hiranya911:
Following test failures are observed when running on environments not set up for ADC:

```
testDbUrlIsEmulatorUrlForDbRefWithPath(com.google.firebase.database.FirebaseDatabaseTest)  Time elapsed: 0.13 sec  <<< ERROR!
java.lang.IllegalStateException: java.io.IOException: The Application Default Credentials are not available. They are available if running in Google Compute Engine. Otherwise, the environment variable GOOGLE_APPLICATION_CREDENTIALS must be defined pointing to a file defining the credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
	at com.google.firebase.database.FirebaseDatabaseTest.testDbUrlIsEmulatorUrlForDbRefWithPath(FirebaseDatabaseTest.java:252)
Caused by: java.io.IOException: The Application Default Credentials are not available. They are available if running in Google Compute Engine. Otherwise, the environment variable GOOGLE_APPLICATION_CREDENTIALS must be defined pointing to a file defining the credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
	at com.google.firebase.database.FirebaseDatabaseTest.testDbUrlIsEmulatorUrlForDbRefWithPath(FirebaseDatabaseTest.java:252)

testDbUrlIsEmulatorUrlWhenSettingOptionsManually(com.google.firebase.database.FirebaseDatabaseTest)  Time elapsed: 0.001 sec  <<< ERROR!
java.lang.IllegalStateException: java.io.IOException: The Application Default Credentials are not available. They are available if running in Google Compute Engine. Otherwise, the environment variable GOOGLE_APPLICATION_CREDENTIALS must be defined pointing to a file defining the credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
	at com.google.firebase.database.FirebaseDatabaseTest.testDbUrlIsEmulatorUrlWhenSettingOptionsManually(FirebaseDatabaseTest.java:215)
Caused by: java.io.IOException: The Application Default Credentials are not available. They are available if running in Google Compute Engine. Otherwise, the environment variable GOOGLE_APPLICATION_CREDENTIALS must be defined pointing to a file defining the credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
	at com.google.firebase.database.FirebaseDatabaseTest.testDbUrlIsEmulatorUrlWhenSettingOptionsManually(FirebaseDatabaseTest.java:215)
```

[...]

Issue #332, opened 2019-12-06
Tenants support in Java Admin SDK


vkercan:
Support for user management based on a Tenant is currently not yet supported in Java Admin SDK, although Google Cloud Identity Platform already supports this (https://cloud.google.com/blog/products/identity-security/multi-tenancy-support-identity-platform-now-generally-available).

Would be helpful to extend this support also in Java Admin SDK.

hiranya911:
@micahstairs FYA

micahstairs:
Thanks for the feature request! Fortunately, I've already begun working on this. We are still in the early stages (working out the design for the API), but once I begin working on the implementation, this issue will track my progress.

micahstairs:
Thanks for your patience! This has now been released. See https://github.com/firebase/firebase-admin-java/releases/tag/v6.15.0.

PR #344, opened 2020-01-09
Add Tenant class and its create and update request classes.

micahstairs:
This pull request adds the Tenant class as part of adding multi-tenancy support (see issue #332).

This is the first file which uses AutoValue, so I've added the relevant dependency.

I cherrypicked a few commits from pull request #334 in order to fix two tests that were failing at the time that the micahstairs-multitenacity branch was created.


src/main/java/com/google/firebase/auth/Tenant.java
```diff
@@ -0,0 +1,80 @@
...
+   * Returns whether to enable email link user authentication.
+   *
+   * @return true if a user can be authenticated using an email link, and false otherwise.
+   */
+  public abstract boolean isEmailLinkSignInEnabled();
+
+  /**
+   * Returns a tenant builder.
+   */
+  public static Builder newBuilder() {
```
micahstairs: I think we should make newBuilder() package-private. We don't expect clients to be able to use this builder directly, right?

Looks like I already had made the Builder class package private, but it's methods are public, which should be changed as well.

pom.xml
```diff
@@ -321,6 +321,13 @@
                 <configuration>
                     <source>1.7</source>
                     <target>1.7</target>
+                    <annotationProcessorPaths>
+                    <path>
```
hiranya911: Indent one level

src/main/java/com/google/firebase/auth/Tenant.java
hiranya911: Actually, do we ever need to build an instance of `Tenant`? It looks like they always get constructed from the JSON responses sent by the Auth backend service, in which case we can just rely on the JSON deserialization support we have.

pom.xml
micahstairs: Done.

src/main/java/com/google/firebase/auth/Tenant.java
micahstairs: Are you suggesting that this class should look more like com.google.firebase.auth.UserRecord? And construct it with some sort of response and a JsonFactory? I'm not sure why this would mean we can't make it an AutoValue.

And I still think the inner CreateRequest and UpdateRequest classes will also make sense as AutoValues.

I don't think I have a good understanding of the API/JSON interaction yet. What is the best way to play around with the REST API and test sending various requests and getting reponses?

hiranya911: We use Google API client to marshal/unmarshal JSON to Java types. If you have a `Tenant` class defined as follows:

```
class Tenant {
  @Key("tenantId")
  private String tenantId;
  
  @Key("displayName")
  private String displayName;
}
```

...then you can construct an instance of it from JSON as follows:

```
Tenant tenant = jsonFactory.createJsonParser(jsonString).parse(Tenant.class);
```

So AutoValue doesn't add anything to this flow as far as I can see. Create and Update classes are essentially builders, and I can see the value in using AutoValue for them.

micahstairs: Alright, I've made the suggested changes.

I got the JSON key names from the Go API (https://github.com/firebase/firebase-admin-go/blob/e088f55036154f0cccf661992f854abd8fb1333f/auth/tenant_mgt.go).

src/main/java/com/google/firebase/auth/Tenant.java
```diff
@@ -0,0 +1,138 @@
...
+
+import com.google.api.client.util.Key;
+import com.google.auto.value.AutoValue;
+
+/**
+ * Contains metadata associated with a Firebase tenant.
+ *
+ * <p>Instances of this class are immutable and thread safe.
+ */
+public class Tenant {
```
hiranya911: `final`

src/main/java/com/google/firebase/auth/Tenant.java
```diff
@@ -0,0 +1,138 @@
+/*
+ * Copyright 2019 Google LLC
```
lahirumaramba: Should this now be 2020? Technically the file was created in 2019 :)

src/main/java/com/google/firebase/auth/Tenant.java
```diff
@@ -0,0 +1,138 @@
...
+
+package com.google.firebase.auth;
+
+import com.google.api.client.util.Key;
+import com.google.auto.value.AutoValue;
+
+/**
+ * Contains metadata associated with a Firebase tenant.
+ *
+ * <p>Instances of this class are immutable and thread safe.
```
lahirumaramba: `<p>` typo? or missing `</p>`?

lahirumaramba: Please ignore this. Just realized how it works in javadoc :)

micahstairs: This was intentional and adheres to standard Javadoc formatting.

The closing tag is unnecessary for certain tag types (e.g. <p>, <li>): https://stackoverflow.com/questions/33787753/javadoc-closing-tags.

src/main/java/com/google/firebase/auth/Tenant.java
micahstairs: Done.

PR #345, opened 2020-01-09
feat(auth): Link federatedid

nrsim:


src/main/java/com/google/firebase/auth/UserRecord.java
```diff
@@ -528,6 +532,17 @@ UpdateRequest setValidSince(long epochSeconds) {
...
     }
 
+    UpdateRequest linkProvider(@NonNull UserProvider userProvider) {
+      properties.put("linkProviderUserInfo", userProvider);
+      return this;
+    }
+
+    UpdateRequest deleteProvider(String providerId) {
+      checkProviderId(providerId);
+      properties.put("deleteProvider", ImmutableList.of(providerId));
```
rsgowman: This would overwrite previous values such that:
```
UpdateRequest ur = ...;
ur.deleteProvider("google.com").deleteProvider("facebook.com")
```
would result in only the facebook.com provider being deleted.

Better might be: `UpdateRequest setDeleteProviders(Iterable<String> providerIds)`. You could optionally add `UpdateRequest addDeleteProvider(String)` and/or `UpdateRequest addDeleteProviders(Iterable<String>)`.

See further discussion here: http://go/java-practices/builders#special. (Second bullet point).

src/main/java/com/google/firebase/auth/UserRecord.java
```diff
@@ -528,6 +532,17 @@ UpdateRequest setValidSince(long epochSeconds) {
       return this;
     }
 
+    UpdateRequest linkProvider(@NonNull UserProvider userProvider) {
```
rsgowman: For our builders (which UpdateRequest effectively is), use `setProperty()` rather than `property()`. (So `setLinkProvider()` here.) Both for consistency with the existing type, (eg setValidSince) and also for consistency with our 'best practices' guide. (http://go/java-practices/builders#api, last bullet point.) Note that this directly conflicts with EJ (as pointed out by the practices doc.)

rsgowman: Note that this part of the api design is currently in flux, so it may change a bit yet. Sorry! :(

No action required (yet).

rsgowman: Should be public (and probably have some javadoc).

src/main/java/com/google/firebase/auth/UserRecord.java
```diff
@@ -528,6 +532,17 @@ UpdateRequest setValidSince(long epochSeconds) {
...
+      properties.put("linkProviderUserInfo", userProvider);
+      return this;
+    }
+
+    UpdateRequest deleteProvider(String providerId) {
```
rsgowman: public/javadoc

src/test/java/com/google/firebase/auth/FirebaseAuthIT.java
```diff
@@ -178,73 +179,134 @@ public void testCreateUserWithParams() throws Exception {
...
+      assertNull(userRecord.getDisplayName());
+      assertEquals(randomUser.email, userRecord.getEmail());
+      assertNull(userRecord.getPhoneNumber());
+      assertNull(userRecord.getPhotoUrl());
+      assertTrue(userRecord.isEmailVerified());
+      assertTrue(userRecord.isDisabled());
+      assertEquals(1, userRecord.getProviderData().length);
+      assertTrue(userRecord.getCustomClaims().isEmpty());
+
+    } finally {
```
rsgowman: Good catch.

src/main/java/com/google/firebase/auth/UserRecord.java
```diff
@@ -526,20 +522,30 @@ public UpdateRequest setCustomClaims(Map<String,Object> customClaims) {
...
 
-    UpdateRequest linkProvider(@NonNull UserProvider userProvider) {
-      properties.put("linkProviderUserInfo", userProvider);
+    /**
+     * Updates the identity providers to unlink from this user\'s account.
+     *
+     * @param providerIds list of identifiers for the identity providers.
+     */
+    public UpdateRequest setDeleteProviders(List<String> providerIds) {
+      checkNotNull(providerIds);
```
rsgowman: It might still be worth checking each providerId to ensure it's not null/empty?

src/main/java/com/google/firebase/auth/UserRecord.java
```diff
@@ -526,20 +522,30 @@ public UpdateRequest setCustomClaims(Map<String,Object> customClaims) {
...
     }
 
-    UpdateRequest linkProvider(@NonNull UserProvider userProvider) {
-      properties.put("linkProviderUserInfo", userProvider);
+    /**
+     * Updates the identity providers to unlink from this user\'s account.
+     *
+     * @param providerIds list of identifiers for the identity providers.
+     */
+    public UpdateRequest setDeleteProviders(List<String> providerIds) {
```
rsgowman: Can `Iterable` be used here (instead of `List`)?

src/main/java/com/google/firebase/auth/UserRecord.java
```diff
@@ -526,20 +522,30 @@ public UpdateRequest setCustomClaims(Map<String,Object> customClaims) {
...
+      checkNotNull(providerIds);
+      properties.put("deleteProvider", providerIds);
```
rsgowman: Check out line 567. (Github doesn't allow you to comment on lines that haven't been changed.) The `deleteProvider` property will be overwritten if the `phone` property is null. I think you'll need to change that to append if it already exists (or set if it does not.)

Michael-Frank:
any progress on this? This feature would be really great to have.

rsgowman:
> any progress on this? This feature would be really great to have.

Yes: We finally got all the approvals necessary to launch this and have started to do so. The node.js port went out last month (https://firebase.google.com/support/release-notes/admin/node#version_950_-_10_february_2021) and the other ports, incl this one, should be released in the coming weeks.

src/main/java/com/google/firebase/auth/UserRecord.java
```diff
@@ -548,6 +548,41 @@ public UpdateRequest setCustomClaims(Map<String,Object> customClaims) {
...
+     * authenticate with this new provider, you need to (a) revoke their
+     * refresh token (see
+     * https://firebase.google.com/docs/auth/admin/manage-sessions#revoke_refresh_tokens),
+     * and (b) ensure no other authentication methods are present on this
+     * account.
+     *
+     * @param providerToLink provider info to be linked to this user\'s account.
+     */
+    public UpdateRequest setProviderToLink(@NonNull UserProvider providerToLink) {
+      properties.put("linkProviderUserInfo", providerToLink);
```
hiranya911, line 567: Do we need a non-null check for the argument?

src/main/java/com/google/firebase/auth/UserRecord.java
```diff
@@ -569,7 +604,24 @@ UpdateRequest setValidSince(long epochSeconds) {
       }
 
       if (copy.containsKey("phoneNumber") && copy.get("phoneNumber") == null) {
-        copy.put("deleteProvider", ImmutableList.of("phone"));
+        Object deleteProvider = copy.get("deleteProvider");
+        if (deleteProvider != null) {
+          if (!(deleteProvider instanceof Iterable<?>)) {
+            throw new IllegalStateException("Contents of deleteProvider was not iterable");
```
hiranya911, line 610: I'd just forego the explicit check and let the cast error pop up. This case should never occur since the writes to the properties map is managed by our code.

src/main/java/com/google/firebase/auth/UserRecord.java
```diff
@@ -569,7 +604,24 @@ UpdateRequest setValidSince(long epochSeconds) {
...
+          }
+
+          // Due to java's type erasure, we can't fully check the type. :(
+          @SuppressWarnings("unchecked")
+          Iterable<String> deleteProviderIterable = (Iterable<String>)deleteProvider;
+
+          copy.put("deleteProvider", new ImmutableList.Builder<String>()
```
hiranya911, line 617: Should we handle the double specification case?

```
user.updateRequest()
  .setPhoneNumber(null)
  .setProvidersToUnlink(ImmutableList.of("phone"))
```

I remember handling this case is other languages.

hiranya911 (changes requested):
Looks good! Just a couple of nits.

src/main/java/com/google/firebase/auth/UserRecord.java
hiranya911: Would it make sense to enforce this constraint in the setters? We'd have to do so in both `setPhoneNumber()` and `setProvidersToUnlink()`, but I think we can reduce the duplication via some sort of a helper function. That would make argument validation consistent with the pattern currently used in this class.

hiranya911 (approved):
LGTM with a suggestion on argument validation.

src/main/java/com/google/firebase/auth/UserRecord.java
rsgowman: Yeah, that's better; done. I've removed the logic here, i.e. if we somehow get into a bad state, we won't notice it. (But now we shouldn't get into that state.)

I didn't create a helper; there wasn't much duplication (though the exception itself is duplicated). If we set the value first and then checked, it would work out better... but that risks leaving the object in a bad state (eg if user catches and ignores the exception and then proceeds to use the object anyways.)

hiranya911 (approved):

PR #350, opened 2020-01-22
Added core types for the error handling revamp

hiranya911:
Starting to implement the error handling revamp. Since this contains sweeping breaking changes, I'm moving the implementation to a separate v7 branch.

* Added `ErrorCode`, `IncomingHttpResponse` and `OutgoingHttpRequest` types.
* Exposed error code and http response from the `FirebaseException` base class.
* Deprecated the old constructors in `FirebaseException`. These will be removed once the error handling revamp is fully implemented.

go/firebase-error-handling-java

### DIFF
--- a/src/main/java/com/google/firebase/ErrorCode.java
+++ b/src/main/java/com/google/firebase/ErrorCode.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Google Inc.
+ * Copyright 2020 Google Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


src/main/java/com/google/firebase/ErrorCode.java
```diff
@@ -0,0 +1,109 @@
...
+
+  /**
+   * Concurrency conflict, such as read-modify-write conflict.
+   */
+  CONFLICT,
+
+  /**
+   * Concurrency conflict, such as read-modify-write conflict.
+   */
+  ABORTED,
```
lahirumaramba: Just to make sure, concurrency conflict and aborted are caused by the same reasons?

hiranya911: `CONFLICT` is really an error code we made up. This is because HTTP 409 can be mapped to both `ABORTED` and `ALREADY_EXISTS`: https://cloud.google.com/apis/design/errors#handling_errors

To mitigate this ambiguity we introduced an error code `CONFLICT`, so that if we only have the HTTP status code to identify an error condition, we will map it to `CONFLICT` instead of one of the other two.

PR #352, opened 2020-01-23
Pull parts of FirebaseAuth into an abstract class.

micahstairs:
This pull request moves part of FirebaseAuth into its abstract class as part of adding multi-tenancy support (see issue #332).

This class will eventually have a getTenantManager() method.


src/main/java/com/google/firebase/auth/FirebaseAuth.java
```diff
@@ -53,35 +33,19 @@
...
-    this.cookieVerifier = threadSafeMemoize(builder.cookieVerifier);
-    this.userManager = threadSafeMemoize(new Supplier<FirebaseUserManager>() {
-      @Override
-      public FirebaseUserManager get() {
-        return new FirebaseUserManager(firebaseApp);
-      }
-    });
-    this.jsonFactory = firebaseApp.getOptions().getJsonFactory();
+    super(
+        checkNotNull(builder.firebaseApp),
```
hiranya911: Do `checkNotNull` in the parent constructor.

src/main/java/com/google/firebase/auth/FirebaseAuth.java
```diff
@@ -1207,4 +148,4 @@ public void destroy() {
       instance.destroy();
     }
   }
-}
+}
```
hiranya911: Add newline

micahstairs: Done.

src/main/java/com/google/firebase/auth/FirebaseAuth.java
micahstairs: Done.

hiranya911:
I think this PR mainly just moves the existing code/docs around. There's probably no need to review docs here.

PR #356, opened 2020-01-28
fix: Enabled automatic HTTP retries for FirebaseProjectManagement

hiranya911:
Enables automatic HTTP retries for the `FirebaseProjectManagement` API. We disable retries for unit tests for faster execution of tests.

RELEASE NOTE: Project management APIs in the `FirebaseProjectManagement` class now automatically retries operations that fail due to retry-eligible HTTP errors.

### DIFF
--- a/src/test/java/com/google/firebase/internal/ApiClientUtilsTest.java
+++ b/src/test/java/com/google/firebase/internal/ApiClientUtilsTest.java
@@ -70,6 +70,19 @@ public class ApiClientUtilsTest {
   }
 
   @Test
+  public void testAuthorizedHttpClientWithoutRetry() throws IOException {
+    FirebaseApp app = FirebaseApp.initializeApp(TEST_OPTIONS);
+
+    HttpRequestFactory requestFactory = ApiClientUtils.newAuthorizedRequestFactory(app, null);
+
+    assertTrue(requestFactory.getInitializer() instanceof FirebaseRequestInitializer);
+    HttpRequest request = requestFactory.buildGetRequest(TEST_URL);
+    assertEquals("Bearer test-token", request.getHeaders().getAuthorization());
+    HttpUnsuccessfulResponseHandler retryHandler = request.getUnsuccessfulResponseHandler();
+    assertFalse(retryHandler instanceof RetryHandlerDecorator);
+  }
+
+  @Test
   public void testUnauthorizedHttpClient() throws IOException {
     FirebaseApp app = FirebaseApp.initializeApp(TEST_OPTIONS);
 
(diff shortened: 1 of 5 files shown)


src/test/java/com/google/firebase/internal/TestApiClientUtils.java
```diff
@@ -0,0 +1,95 @@
...
+    } catch (IOException e) {
+      throw new RuntimeException("Failed to initialize request", e);
+    }
+
+    // Verify authorization
+    assertTrue(request.getHeaders().getAuthorization().startsWith("Bearer "));
+
+    // Verify retry support
+    HttpUnsuccessfulResponseHandler retryHandler = request.getUnsuccessfulResponseHandler();
+    assertTrue(retryHandler instanceof RetryHandlerDecorator);
```
weixifan: Since we are essentially saying that the `unsuccessfulResponseHandler` must be a `RetryHandlerDecorator`, perhaps it is worthwhile to specify this explicitly in the Javadoc of the `newAuthorizedRequestFactory` method. Otherwise, it might seem like this test amounts to forbidding any changes to the implementation rather than a behavioural "black-box" test of the retry mechanism.

A more drastic change might be to add a new subclass of `HttpRequestFactory` called `RetryingHttpRequestFactory` and its `buildGetRequest` returns a new subclass of `HttpRequest` called `RetryingHttpRequest` whose `getUnsuccessfulResponseHandler` returns a `RetryHandlerDecorator` or some suitable base class. This way, by design, we are *compile time* compliant, instead of relying on this `instanceof` test.

src/test/java/com/google/firebase/internal/ApiClientUtilsTest.java
```diff
@@ -69,6 +69,19 @@ public void testAuthorizedHttpClient() throws IOException {
...
+  public void testAuthorizedHttpClientWithoutRetry() throws IOException {
+    FirebaseApp app = FirebaseApp.initializeApp(TEST_OPTIONS);
+
+    HttpRequestFactory requestFactory = ApiClientUtils.newAuthorizedRequestFactory(app, null);
+
+    assertTrue(requestFactory.getInitializer() instanceof FirebaseRequestInitializer);
+    HttpRequest request = requestFactory.buildGetRequest(TEST_URL);
+    assertEquals("Bearer test-token", request.getHeaders().getAuthorization());
+    HttpUnsuccessfulResponseHandler retryHandler = request.getUnsuccessfulResponseHandler();
+    assertFalse(retryHandler instanceof RetryHandlerDecorator);
```
weixifan: Similar comments here -- see below.

PR #358, opened 2020-01-30
Add ListTenantsPage class.

micahstairs:
This pull request adds ListTenantsPage (and some supporting code) as part of adding multi-tenancy support (see issue #332).

This code was very largely based off of ListUsersPage and ListUsersPageTest.

### DIFF
--- a/src/main/java/com/google/firebase/auth/FirebaseUserManager.java
+++ b/src/main/java/com/google/firebase/auth/FirebaseUserManager.java
@@ -42,8 +42,8 @@ import com.google.firebase.auth.UserRecord.CreateRequest;
 import com.google.firebase.auth.UserRecord.UpdateRequest;
 import com.google.firebase.auth.internal.DownloadAccountResponse;
 import com.google.firebase.auth.internal.GetAccountInfoResponse;
-
 import com.google.firebase.auth.internal.HttpErrorResponse;
+import com.google.firebase.auth.internal.ListTenantsResponse;
 import com.google.firebase.auth.internal.UploadAccountResponse;
 import com.google.firebase.internal.FirebaseRequestInitializer;
 import com.google.firebase.internal.NonNull;
@@ -57,6 +57,9 @@ import java.util.Map;
 /**
  * FirebaseUserManager provides methods for interacting with the Google Identity Toolkit via its
  * REST API. This class does not hold any mutable state, and is thread safe.
+ * 
+ * <p>TODO(micahstairs): Consider renaming this to FirebaseAuthManager since this also supports
+ * tenants.
  *
  * @see <a href="https://developers.google.com/identity/toolkit/web/reference/relyingparty">
  *   Google Identity Toolkit</a>
@@ -87,6 +90,7 @@ class FirebaseUserManager {
       .put("INVALID_DYNAMIC_LINK_DOMAIN", "invalid-dynamic-link-domain")
       .build();
 
+  static final int MAX_LIST_TENANTS_RESULTS = 1000;
   static final int MAX_LIST_USERS_RESULTS = 1000;
   static final int MAX_IMPORT_USERS = 1000;
 
@@ -95,10 +99,11 @@ class FirebaseUserManager {
       "iss", "jti", "nbf", "nonce", "sub", "firebase");
 
   private static final String ID_TOOLKIT_URL =
-      "https://identitytoolkit.googleapis.com/v1/projects/%s";
+      "https://identitytoolkit.googleapis.com/%s/projects/%s";
   private static final String CLIENT_VERSION_HEADER = "X-Client-Version";
 
-  private final String baseUrl;
+  private final String userBaseUrl;
+  private final String tenantBaseUrl;
   private final JsonFactory jsonFactory;
   private final HttpRequestFactory requestFactory;
   private final String clientVersion = "Java/Admin/" + SdkUtils.getVersion();
@@ -117,7 +122,8 @@ class FirebaseUserManager {
         "Project ID is required to access the auth service. Use a service account credential or "
             + "set the project ID explicitly via FirebaseOptions. Alternatively you can also "
             + "set the project ID via the GOOGLE_CLOUD_PROJECT environment variable.");
-    this.baseUrl = String.format(ID_TOOLKIT_URL, projectId);
+    this.userBaseUrl = String.format(ID_TOOLKIT_URL, "v1", projectId);
+    this.tenantBaseUrl = String.format(ID_TOOLKIT_URL, "v2", projectId);
     this.jsonFactory = app.getOptions().getJsonFactory();
     HttpTransport transport = app.getOptions().getHttpTransport();
     this.requestFactory = transport.createRequestFactory(new FirebaseRequestInitializer(app));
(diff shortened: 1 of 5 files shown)


hiranya911:
@micahstairs thanks for putting this together. Please remember to set the assignee of the PRs when they are ready to be reviewed. That way it shows up in our radar sooner.

src/main/java/com/google/firebase/auth/FirebaseUserManager.java
```diff
@@ -117,7 +122,8 @@
         "Project ID is required to access the auth service. Use a service account credential or "
             + "set the project ID explicitly via FirebaseOptions. Alternatively you can also "
             + "set the project ID via the GOOGLE_CLOUD_PROJECT environment variable.");
-    this.baseUrl = String.format(ID_TOOLKIT_URL, projectId);
+    this.userBaseUrl = String.format(ID_TOOLKIT_URL, "v1", projectId);
+    this.tenantBaseUrl = String.format(ID_TOOLKIT_URL, "v2", projectId);
```
hiranya911: I noticed that Go SDK uses v2beta1. But if this is indeed supported, then it's ok (and even preferred).

src/main/java/com/google/firebase/auth/FirebaseUserManager.java
```diff
@@ -221,6 +227,25 @@ UserImportResult importUsers(UserImportRequest request) throws FirebaseAuthExcep
...
+  ListTenantsResponse listTenants(int maxResults, String pageToken) throws FirebaseAuthException {
+    ImmutableMap.Builder<String, Object> builder = ImmutableMap.<String, Object>builder()
+        .put("maxResults", maxResults);
+    if (pageToken != null) {
+      checkArgument(!pageToken.equals(
+          ListTenantsPage.END_OF_LIST), "invalid end of list page token");
+      builder.put("nextPageToken", pageToken);
+    }
+
+    GenericUrl url = new GenericUrl(tenantBaseUrl + "/tenants:tenants");
```
hiranya911: I believe this should be just `/tenants`

src/main/java/com/google/firebase/auth/FirebaseUserManager.java
```diff
@@ -221,6 +227,25 @@ UserImportResult importUsers(UserImportRequest request) throws FirebaseAuthExcep
...
   }
 
+  // TODO(micahstairs): Write unit tests for this method.
+  ListTenantsResponse listTenants(int maxResults, String pageToken) throws FirebaseAuthException {
+    ImmutableMap.Builder<String, Object> builder = ImmutableMap.<String, Object>builder()
+        .put("maxResults", maxResults);
+    if (pageToken != null) {
+      checkArgument(!pageToken.equals(
+          ListTenantsPage.END_OF_LIST), "invalid end of list page token");
+      builder.put("nextPageToken", pageToken);
```
hiranya911: I think the query params are called `pageSize` and `pageToken`. Verify and correct if necessary.

src/main/java/com/google/firebase/auth/internal/ListTenantsResponse.java
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  public boolean hasTenants() {
+    return tenants != null && !tenants.isEmpty();
+  }
+
+  public String getPageToken() {
+    return pageToken;
+  }
+}
```
hiranya911: Add newline

src/main/java/com/google/firebase/auth/ListTenantsPage.java
```diff
@@ -0,0 +1,273 @@
...
+      if (response.hasTenants()) {
+        builder.addAll(response.getTenants());
+      }
+      String nextPageToken = response.getPageToken() != null
+          ? response.getPageToken() : END_OF_LIST;
+      return new ListTenantsResult(builder.build(), nextPageToken);
+    }
+  }
+
+  static final class ListTenantsResult {
```
hiranya911: You probably don't need this type. Just try to use the ListTenantsResponse type directly.

`listUsers()` API use this middle-ground type since the REST API returns a `User` type which needs to be mapped to `ExportedUserRecord`. But that requirement is not present here.

src/main/java/com/google/firebase/auth/FirebaseUserManager.java
micahstairs: Yeah, I did notice that the Go implementation was using the beta. But I think the non-beta version is supported: https://identitytoolkit.googleapis.com/$discovery/rest?version=v2.

micahstairs:
@googlebot I fixed it.

src/main/java/com/google/firebase/auth/FirebaseUserManager.java
micahstairs: After taking another look at https://identitytoolkit.googleapis.com/$discovery/rest?version=v2, I'm thinking it should be "/tenants:list".

src/main/java/com/google/firebase/auth/FirebaseUserManager.java
micahstairs: You are correct.

src/main/java/com/google/firebase/auth/internal/ListTenantsResponse.java
micahstairs: Done.

src/main/java/com/google/firebase/auth/ListTenantsPage.java
micahstairs: Okay, but I had to add a constructor to ListTenantsResponse so that I could construct it in the unit test.

src/main/java/com/google/firebase/auth/FirebaseUserManager.java
```diff
@@ -117,7 +122,8 @@
         "Project ID is required to access the auth service. Use a service account credential or "
             + "set the project ID explicitly via FirebaseOptions. Alternatively you can also "
             + "set the project ID via the GOOGLE_CLOUD_PROJECT environment variable.");
-    this.baseUrl = String.format(ID_TOOLKIT_URL, projectId);
+    this.userBaseUrl = String.format(ID_TOOLKIT_URL, "v1", projectId);
```
hiranya911: Nit: Lets call them `userMgtBaseUrl` and `tenantMgtBaseUrl` for clarity.

src/main/java/com/google/firebase/auth/internal/ListTenantsResponse.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+public class ListTenantsResponse {
+
+  @Key("tenants")
+  private List<Tenant> tenants;
+
+  @Key("pageToken")
+  private String pageToken;
+
+  public ListTenantsResponse(List<Tenant> tenants, String pageToken) {
```
hiranya911: You probably won't be able to parse/deserialize into this unless you also define a no-args constructor. Chances are the existing unit tests don't hit that case as of yet. But it's fine to defer adding it until we hit that case.

src/test/java/com/google/firebase/auth/ListTenantsPageTest.java
```diff
@@ -0,0 +1,350 @@
+/*
+ * Copyright 202 Google LLC
```
hiranya911: 2020 (also check other files)

micahstairs: Done.

src/main/java/com/google/firebase/auth/FirebaseUserManager.java

src/main/java/com/google/firebase/auth/internal/ListTenantsResponse.java
micahstairs: Oh, you're right, I think we do need that (I slipped it into this PR).

No, there's not any unit tests which hit this yet. Those will come when I flesh out the other tenant methods in FirebaseUserManager.

PR #361, opened 2020-02-07
Add updateRequest method to Tenant class and add unit tests.

micahstairs:
This pull request adds some things to the Tenant class and adds a few unit tests. This is part of the initiative to adding multi-tenancy support (see issue #332).

It's worth noting that these changes deviate from the API design doc. I realized that the updateTenant method in my design doc doesn’t match the design of the existing updateUser method. To fix this, I'm adding the tenantId to Tenant.UpdateRequest.

This PR also removes the usage of AutoValue since using it was going to deviate too much from the design of the request classes in UserRecord. This change will also make it easier to do JSON serialization when we sending the requests.

### DIFF
--- a/src/main/java/com/google/firebase/auth/Tenant.java
+++ b/src/main/java/com/google/firebase/auth/Tenant.java
@@ -55,6 +55,22 @@ public final class Tenant {
   }
 
   /**
+   * Returns a new {@link UpdateRequest}, which can be used to update the attributes
+   * of this tenant.
+   *
+   * @return a non-null Tenant.UpdateRequest instance.
+   */
+  public UpdateRequest updateRequest() {
+    return UpdateRequest
+      .newBuilder()
+      .setTenantId(getTenantId())
+      .setDisplayName(getDisplayName())
+      .setPasswordSignInAllowed(isPasswordSignInAllowed())
+      .setEmailLinkSignInEnabled(isEmailLinkSignInEnabled())
+      .build();
+  }
+
+  /**
    * Class used to hold the information needs to make a tenant create request.
    */
   @AutoValue
@@ -85,12 +101,16 @@ public final class Tenant {
      * Returns a builder for a tenant create request.
      */
     public static Builder newBuilder() {
+      // TOOD(micahstairs): Figure out if any of the CreateRequest fields should default to anything
+      // particular. If any fields do not have a reasonable default, then the
+      // fields should be made nullable, otherwise AutoValue will throw an
+      // exception if those fields are left unset.
       return new AutoValue_Tenant_CreateRequest.Builder();
     }
 
     /**
-    * Builder class used to construct a create request.
-    */
+     * Builder class used to construct a create request.
+     */
     @AutoValue.Builder
     abstract static class Builder {
       public abstract Builder setDisplayName(String displayName);
@@ -107,7 +127,14 @@ public final class Tenant {
    * Class used to hold the information needs to make a tenant update request. 
    */
   @AutoValue
-  public abstract static class UpdateRequest {
+  public abstract static class UpdateRequest { 
+
+    /**
+     * Returns the id of this tenant.
+     *
+     * @return a non-empty tenant id string.
+     */
+    public abstract String getTenantId();
 
     /**
      * Returns the display name of this tenant.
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/firebase/auth/Tenant.java
```diff
@@ -134,14 +161,20 @@ public static Builder newBuilder() {
      * Returns a builder for a tenant update request.
      */
     public static Builder newBuilder() {
+      // TOOD(micahstairs): Figure out if any of the UpdateRequest fields should default to anything
+      // particular. If any fields do not have a reasonable default, then the
+      // fields should be made nullable, otherwise AutoValue will throw an
+      // exception if those fields are left unset.
```
hiranya911: I think this shouldn't be exposed. `Builder` seems to be a package-protected. Also we don't have a similar method in `UserRecord.UpdateRequest`. May be just expose the constructor.

```
Tenant tenant = ...;
UpdateRequest req = tenant.updateRequest();

// Or alternatively users should be able to:
UpdateRequest req = new Tenant.UpdateReuqst(tenant.getTenantId());
```

src/main/java/com/google/firebase/auth/Tenant.java
```diff
@@ -54,6 +54,22 @@ public boolean isEmailLinkSignInEnabled() {
...
   }
 
+  /**
+   * Returns a new {@link UpdateRequest}, which can be used to update the attributes
+   * of this tenant.
+   *
+   * @return a non-null Tenant.UpdateRequest instance.
+   */
+  public UpdateRequest updateRequest() {
+    return UpdateRequest
```
hiranya911: I think we should just set the tenantId. Other parameters will not be set, which means they won't be updated in the back-end.

micahstairs: Alright this makes sense!

src/main/java/com/google/firebase/auth/Tenant.java
micahstairs: Hmm.. I gave this some thought and ended up rewriting the create/update classes without using AutoValue. This is a lot more consistent with UserRecord's request classes. It will also make it easier to serialize these into json requests.

src/main/java/com/google/firebase/auth/Tenant.java
```diff
@@ -55,100 +60,136 @@ public boolean isEmailLinkSignInEnabled() {
...
+    return new UpdateRequest(getTenantId());
+  }
+
+  /**
+   * A specification class for creating new user accounts.
+   *
+   * <p>Set the initial attributes of the new tenant by calling various setter methods available in
+   * this class. None of the attributes are required.
+   */
+  public static class CreateRequest {
```
hiranya911: final

src/main/java/com/google/firebase/auth/Tenant.java
```diff
@@ -55,100 +60,136 @@ public boolean isEmailLinkSignInEnabled() {
...
-   * Class used to hold the information needs to make a tenant update request. 
+   * A class for updating the attributes of an existing tenant.
+   *
+   * <p>An instance of this class can be obtained via a {@link Tenant} object, or from a tenant ID
+   * string. Specify the changes to be made to the tenant by calling the various setter methods
+   * available in this class.
    */
-  @AutoValue
-  public abstract static class UpdateRequest {
+  public static class UpdateRequest {
```
hiranya911: final

src/test/java/com/google/firebase/auth/TenantTest.java
```diff
@@ -0,0 +1,89 @@
...
+import org.junit.Test;
+
+public class TenantTest {
+
+  private static final JsonFactory jsonFactory = Utils.getDefaultJsonFactory();
+
+  @Test
+  public void testJsonSerialization() throws IOException {
+    Tenant tenant = jsonFactory.fromString(
+        "{"
```
hiranya911: Define a constant for this and avoid duplication.

src/test/java/com/google/firebase/auth/TenantTest.java
```diff
@@ -0,0 +1,89 @@
...
+          + "\"enableEmailLinkSignin\":false"
+          + "}", Tenant.class);
+
+    Tenant.UpdateRequest updateRequest = tenant.updateRequest();
+
+    assertEquals("TENANT_ID", updateRequest.getTenantId());
+    Map<String,Object> properties = updateRequest.getProperties();
+    assertEquals(properties.size(), 1);
+    assertEquals("TENANT_ID", (String) properties.get("tenantId"));
+    assertNull(properties.get("displayName"));
```
hiranya911: These last 3 assertions are probably redundant since you've already tested for size == 1

src/main/java/com/google/firebase/auth/Tenant.java
micahstairs: Done.

src/test/java/com/google/firebase/auth/TenantTest.java
micahstairs: Done.

src/test/java/com/google/firebase/auth/TenantTest.java
micahstairs: Good point.

src/main/java/com/google/firebase/auth/Tenant.java
micahstairs: Done.

PR #369, opened 2020-02-17
Create TenantManager class and wire through listTenants operation.

micahstairs:
This pull request adds the TenantManager class. The listTenants operation has been added and is wired through. I've added the relevant unit tests to FirebaseUserManagerTest (based off of the listUsers unit tests). This is part of the initiative to adding multi-tenancy support (see issue #332).

### DIFF
--- a/src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
+++ b/src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
@@ -690,8 +690,7 @@ public abstract class AbstractFirebaseAuth {
    * UpdateRequest}.
    *
    * @param request A non-null {@link UpdateRequest} instance.
-   * @return A {@link UserRecord} instance corresponding to the updated user account. account, the
-   *     task fails with a {@link FirebaseAuthException}.
+   * @return A {@link UserRecord} instance corresponding to the updated user account.
    * @throws NullPointerException if the provided update request is null.
    * @throws FirebaseAuthException if an error occurs while updating the user account.
    */
@@ -1053,7 +1052,6 @@ public abstract class AbstractFirebaseAuth {
         .callAsync(firebaseApp);
   }
 
-  @VisibleForTesting
   FirebaseUserManager getUserManager() {
     return this.userManager.get();
   }
@@ -1074,7 +1072,7 @@ public abstract class AbstractFirebaseAuth {
     };
   }
 
-  private <T> Supplier<T> threadSafeMemoize(final Supplier<T> supplier) {
+  <T> Supplier<T> threadSafeMemoize(final Supplier<T> supplier) {
     checkNotNull(supplier);
     return Suppliers.memoize(
         new Supplier<T>() {
--- a/src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
+++ b/src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
@@ -1072,7 +1072,7 @@ public abstract class AbstractFirebaseAuth {
     };
   }
 
-  <T> Supplier<T> threadSafeMemoize(final Supplier<T> supplier) {
+  protected <T> Supplier<T> threadSafeMemoize(final Supplier<T> supplier) {
     checkNotNull(supplier);
     return Suppliers.memoize(
         new Supplier<T>() {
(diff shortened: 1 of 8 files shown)


src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
```diff
@@ -1074,7 +1072,7 @@ protected String execute() throws FirebaseAuthException {
     };
   }
 
-  private <T> Supplier<T> threadSafeMemoize(final Supplier<T> supplier) {
+  <T> Supplier<T> threadSafeMemoize(final Supplier<T> supplier) {
```
hiranya911: `protected` makes more sense here.

src/main/java/com/google/firebase/auth/TenantManager.java
```diff
@@ -0,0 +1,118 @@
...
+import com.google.firebase.internal.Nullable;
+
+/**
+ * This class can be used to perform a variety of tenant-related operations, including creating,
+ * updating, and listing tenants.
+ *
+ * <p>TODO(micahstairs): Implement the following methods: getAuthForTenant(), getTenant(),
+ * deleteTenant(), createTenant(), and updateTenant().
+ */
+public class TenantManager {
```
hiranya911: final

src/main/java/com/google/firebase/auth/TenantManager.java
```diff
@@ -0,0 +1,118 @@
...
+
+  private final FirebaseApp firebaseApp;
+  private final FirebaseUserManager userManager;
+
+  public TenantManager(FirebaseApp firebaseApp, FirebaseUserManager userManager) {
```
hiranya911: Make the constructor package-protected

src/main/java/com/google/firebase/auth/internal/ListTenantsResponse.java
```diff
@@ -17,20 +17,24 @@
...
 
 /**
  * JSON data binding for ListTenantsResponse messages sent by Google identity toolkit service.
  */
-public class ListTenantsResponse {
+public final class ListTenantsResponse {
 
   @Key("tenants")
-  private List<Tenant> tenants;
+  private List<Tenant> tenants = ImmutableList.of(); // Default to empty list.
```
hiranya911: We don't use defaults in `DownloadAccountResponse`. Why were they necessary here?

src/test/java/com/google/firebase/auth/FirebaseUserManagerTest.java
```diff
@@ -1264,6 +1310,13 @@ private static void checkUserRecord(UserRecord userRecord) {
     assertEquals("gold", claims.get("package"));
   }
 
+  private static void checkTenant(Tenant tenant) {
```
hiranya911: Suggestion: Just to be in safe side, let's give your test tenants different IDs (say TENANT_ID_1 and TENANT_ID_2). Then change the signature of this method to accept both `Tenant` and `String tenantId`, so the caller can pass them in to verify.

micahstairs: Done.

src/main/java/com/google/firebase/auth/TenantManager.java
micahstairs: Done.

src/main/java/com/google/firebase/auth/TenantManager.java
micahstairs: Done.

src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
micahstairs: Done.

src/main/java/com/google/firebase/auth/internal/ListTenantsResponse.java
micahstairs: The reason we didn't have to use them in DownloadAccountResponse is because of what we do in ListUsersPage.DefaultUserSource.fetch() with the ListUsersResult class as an intermediate storage class. We handle the null case and empty list case there.

But in ListTenantsPage, we decided not to include ListTenantsResult.

Having these defaults set in ListTenantsResponse does feel clunky to me though. Perhaps I should re-add the ListTenantsResult class for consistency?

hiranya911: Perhaps handle them in the getters?

```
public List<Tenant> getTenants() {
  if (tenants != null) {
    return tenants;
  }

  return ImmutableList.of();
}
```

PR #371, opened 2020-02-20
Add getTenant operation to TenantManager.

micahstairs:
This pull request add getTenant to the TenantManager class. I've added the relevant unit tests to FirebaseUserManagerTest (based off of the getUser unit tests). This is part of the initiative to adding multi-tenancy support (see issue #332).

REST API: https://identitytoolkit.googleapis.com/$discovery/rest?version=v2

### DIFF
--- a/src/main/java/com/google/firebase/auth/FirebaseUserManager.java
+++ b/src/main/java/com/google/firebase/auth/FirebaseUserManager.java
@@ -228,27 +228,8 @@ class FirebaseUserManager {
     return new UserImportResult(request.getUsersCount(), response);
   }
 
-<<<<<<< HEAD
-  Tenant getTenantById(String tenantId) throws FirebaseAuthException {
-    GenericUrl url = new GenericUrl(tenantMgtBaseUrl + "/tenants:get");
-    url.put("name", tenantId);
-=======
-  void deleteTenant(String tenantId) throws FirebaseAuthException {
-    GenericUrl url = new GenericUrl(tenantMgtBaseUrl + "/tenants/" + tenantId);
-    GenericJson response = sendRequest("DELETE", url, null, GenericJson.class);
-    if (response == null) {
-      throw new FirebaseAuthException(TENANT_NOT_FOUND_ERROR,
-          "Failed to delete tenant: " + tenantId);
-    }
-  }
-
   Tenant getTenant(String tenantId) throws FirebaseAuthException {
-<<<<<<< HEAD
-    GenericUrl url = new GenericUrl(tenantMgtBaseUrl + "/tenants:" + tenantId);
->>>>>>> 4b78b4e... Address pull request feedback.
-=======
     GenericUrl url = new GenericUrl(tenantMgtBaseUrl + "/tenants/" + tenantId);
->>>>>>> 82ee04c... Assert that the requests are sent to the correct URL.
     Tenant response = sendRequest("GET", url, null, Tenant.class);
     if (Strings.isNullOrEmpty(response.getTenantId())) {
       throw new FirebaseAuthException(TENANT_NOT_FOUND_ERROR, "Failed to get tenant.");
--- a/src/main/java/com/google/firebase/auth/FirebaseUserManager.java
+++ b/src/main/java/com/google/firebase/auth/FirebaseUserManager.java
@@ -65,6 +65,7 @@ import java.util.Map;
  */
 class FirebaseUserManager {
 
+  static final String TENANT_NOT_FOUND_ERROR = "tenant-not-found";
   static final String USER_NOT_FOUND_ERROR = "user-not-found";
   static final String INTERNAL_ERROR = "internal-error";
 
--- a/src/main/java/com/google/firebase/auth/FirebaseUserManager.java
+++ b/src/main/java/com/google/firebase/auth/FirebaseUserManager.java
@@ -83,6 +83,7 @@ class FirebaseUserManager {
       .put("INVALID_PHONE_NUMBER", "invalid-phone-number")
       .put("PHONE_NUMBER_EXISTS", "phone-number-already-exists")
       .put("PROJECT_NOT_FOUND", "project-not-found")
+      .put("TENANT_NOT_FOUND", TENANT_NOT_FOUND_ERROR)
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/firebase/auth/FirebaseUserManager.java
```diff
@@ -226,6 +228,16 @@ UserImportResult importUsers(UserImportRequest request) throws FirebaseAuthExcep
     return new UserImportResult(request.getUsersCount(), response);
   }
 
+  Tenant getTenantById(String tenantId) throws FirebaseAuthException {
```
hiranya911: I think `getTenant` is descriptive enough. We have several different ways of fetching user accounts (by UID, by email etc), but we don't have that requirement here.

src/main/java/com/google/firebase/auth/FirebaseUserManager.java
```diff
@@ -226,6 +228,16 @@ UserImportResult importUsers(UserImportRequest request) throws FirebaseAuthExcep
...
+    GenericUrl url = new GenericUrl(tenantMgtBaseUrl + "/tenants:get");
+    url.put("name", tenantId);
```
hiranya911: This doesn't quite look right. The URL path should be something like `v2/projects/{projectsId}/tenants/{tenantsId}`. That's what we use in other SDKs, and also what's listed in the auto-generated discovery doc.

src/test/java/com/google/firebase/auth/FirebaseUserManagerTest.java
```diff
@@ -430,6 +430,28 @@ public void testImportUsersLargeList() {
...
+    TestResponseInterceptor interceptor = initializeAppForUserManagement(
+        TestUtils.loadResource("getTenant.json"));
+    Tenant tenant = FirebaseAuth.getInstance().getTenantManager().getTenantAsync("TENANT_1").get();
+    checkTenant(tenant, "TENANT_1");
+    checkRequestHeaders(interceptor);
+  }
+
+  @Test
+  public void testGetTenantWithNotFoundError() throws Exception {
+    initializeAppForUserManagement("{}");
```
hiranya911: I believe this endpoint sends a 404 (or similar error response) when the tenant is not found. Lets update the test case accordingly. See https://github.com/firebase/firebase-admin-go/blob/e088f55036154f0cccf661992f854abd8fb1333f/auth/tenant_mgt_test.go#L1096-L1100

micahstairs: I don't think I follow. The Node implementation suggests I check for a response missing a 'name' (which is the tenantId): https://github.com/firebase/firebase-admin-node/blob/2b952d43fd8784151ebdb60b65e88b9fd1888ccf/src/auth/auth-api-request.ts#L1463.

src/main/java/com/google/firebase/auth/FirebaseUserManager.java
micahstairs: Thanks! Fixed.

src/main/java/com/google/firebase/auth/FirebaseUserManager.java

micahstairs:
I think I'm also going to assert that the URL of the outgoing request is correct, like you suggested on my other PR.

micahstairs:
> I think I'm also going to assert that the URL of the outgoing request is correct, like you suggested on my other PR.

Done.

src/test/java/com/google/firebase/auth/FirebaseUserManagerTest.java
hiranya911: That's just some additional validation they do on the "successful" responses. On missing tenants, the backend sends a 404 NOT_FOUND error.

src/test/java/com/google/firebase/auth/FirebaseUserManagerTest.java
```diff
@@ -437,19 +437,23 @@ public void testGetTenant() throws Exception {
     Tenant tenant = FirebaseAuth.getInstance().getTenantManager().getTenantAsync("TENANT_1").get();
     checkTenant(tenant, "TENANT_1");
     checkRequestHeaders(interceptor);
+    checkUrl(interceptor,
+        "https://identitytoolkit.googleapis.com/v2/projects/test-project-id/tenants/TENANT_1");
```
hiranya911: Nit: Define the URL prefix (up to the /tenants portion) in a constant.

src/test/java/com/google/firebase/auth/FirebaseUserManagerTest.java
```diff
@@ -430,6 +430,32 @@ public void testImportUsersLargeList() {
...
+        TestUtils.loadResource("getTenant.json"));
+    Tenant tenant = FirebaseAuth.getInstance().getTenantManager().getTenantAsync("TENANT_1").get();
+    checkTenant(tenant, "TENANT_1");
+    checkRequestHeaders(interceptor);
+    checkUrl(interceptor,
+        "https://identitytoolkit.googleapis.com/v2/projects/test-project-id/tenants/TENANT_1");
+  }
+
+  @Test
+  public void testGetTenantWithNotFoundError() throws Exception {
```
hiranya911: Compare with https://github.com/firebase/firebase-admin-go/blob/e088f55036154f0cccf661992f854abd8fb1333f/auth/tenant_mgt_test.go#L1151-L1161

src/test/java/com/google/firebase/auth/FirebaseUserManagerTest.java
micahstairs: Done.

src/test/java/com/google/firebase/auth/FirebaseUserManagerTest.java
micahstairs: Done.

micahstairs:
I had to rebase since the other PR was submitted first. Hopefully I did that right.

micahstairs:
It took a lot of work to get this rebased properly, but I learned a lot about git in the process :)

The individual commits are bit messed up (the code wasn't always in a state where it could be compiled after each commit), however, it's all about to get squashed anyway.

PR #375, opened 2020-02-25
Allow custom FCM endpoint URLs

### LINKED ISSUE #374: FR: Allow custom FCM endpoint URL
FCM endpoint URL is hardcoded in [`FirebaseMessagingClientImpl#FCM_URL`](https://github.com/firebase/firebase-admin-java/blob/5f1e48c/src/main/java/com/google/firebase/messaging/FirebaseMessagingClientImpl.java#L56) (see also `FCM_BATCH_URL` in the same class).  

Ability to override it would make some kinds of testing easier (you can point it to some mock-FCM to verify sent messages).

pertu:
resolves #374

### DIFF
--- a/src/main/java/com/google/firebase/FirebaseOptions.java
+++ b/src/main/java/com/google/firebase/FirebaseOptions.java
@@ -78,6 +78,8 @@ public final class FirebaseOptions {
   private final String projectId;
   private final String serviceAccountId;
   private final HttpTransport httpTransport;
+  private final String customFcmUrl;
+  private final String customFcmBatchUrl;
   private final int connectTimeout;
   private final int readTimeout;
   private final JsonFactory jsonFactory;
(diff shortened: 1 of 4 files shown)


pertu:
@googlebot I signed it!

hiranya911:
@perty thanks for the pull request.

However, this is essentially trying to white-box test the SDK (i.e. testing whether the SDK sends the expected messages), and that is not really something we wish to encourage. Developers should treat the SDK (and any other third-party library for that matter) as a black-box. Otherwise, if we ever change the internal implementation details of the SDK (e.g. the message format used to communicate with FCM), your tests will break.

To that end we recommend developers to put the SDK API calls behind an interface, and employ black-box testing methods in their own projects:

```
// Custom interface. Used everywhere in the developer code.
interface NotificationSender {
  String send(String title, String body);
}

class FCMNotificationSender implements NotificationSender {
  // Implementation that uses Admin SDK
}

class MockNotificationSender implements NotificationSender {
  // Implementation for testing
}
```

This way the developer doesn't test any SDK functionality, but only tests their own code.

If you insist on white-box testing, then please look at the `HttpTransport` interface, which allows you to control how the SDK makes outgoing HTTP calls. You can pass an implementation of this via `FirebaseOptions`. See here for how we use it in our own unit tests:

https://github.com/firebase/firebase-admin-java/blob/c40a69d04c73e158a0ea1e75b1ea7400c3544c92/src/test/java/com/google/firebase/messaging/FirebaseMessagingClientImplTest.java#L494-L507

If you want, you can even intercept and override the URL of requests in your own custom `HttpTransport` implementation. But like I said, this level of white-box testing is really not recommended, as it requires more work and the end result could become unstable if the SDK implementation ever changes.

perty:
Oh that's not my PR, it's pertu.

hiranya911:
Oops! Sorry about that :)

PR #377, opened 2020-03-02
Add createTenant and updateTenant operations.

micahstairs:
This pull request adds the createTenant and updateTenant operations to the API. I've added the relevant unit tests to FirebaseUserManagerTest. This is part of the initiative to adding multi-tenancy support (see issue #332).

I've also made a few tweaks to the implementation and tests of the other tenant operations:
- Calling the synchronous versions of the API (as opposed to the async versions) makes the tests cleaner.
- I've improved URL assertions in a few places.
- I've simplified the error handling of getTenant and deleteTenant (after realizing that the manual checks I put in place were doing nothing since the server responds with an error and a firebase exception with the proper error code gets thrown).

REST API: https://identitytoolkit.googleapis.com/$discovery/rest?version=v2


src/main/java/com/google/firebase/auth/FirebaseUserManager.java
```diff
@@ -230,20 +230,28 @@ UserImportResult importUsers(UserImportRequest request) throws FirebaseAuthExcep
...
-    if (Strings.isNullOrEmpty(response.getTenantId())) {
-      throw new FirebaseAuthException(TENANT_NOT_FOUND_ERROR, "Failed to get tenant.");
-    }
-    return response;
+    return sendRequest("GET", url, null, Tenant.class);
+  }
+
+  Tenant createTenant(Tenant.CreateRequest request) throws FirebaseAuthException {
+    GenericUrl url = new GenericUrl(tenantMgtBaseUrl + "/tenants");
+    url.putAll(request.getProperties());
```
hiranya911: Put the properties in the request body.

src/main/java/com/google/firebase/auth/FirebaseUserManager.java
```diff
@@ -230,20 +230,28 @@ UserImportResult importUsers(UserImportRequest request) throws FirebaseAuthExcep
...
+    return sendRequest("POST", url, null, Tenant.class);
+  }
+
+  Tenant updateTenant(Tenant.UpdateRequest request) throws FirebaseAuthException {
+    Map<String, Object> properties = request.getProperties();
+    checkArgument(properties.size() > 1,
+        "tenant update must have at least one property other than tenantId set");
+    GenericUrl url = new GenericUrl(tenantMgtBaseUrl + "/tenants/" + request.getTenantId());
+    url.putAll(properties);
+    url.putAll(request.getProperties());
```
hiranya911: Put the properties in the request body. Remove the extra call to `getProperties()`.

src/main/java/com/google/firebase/auth/FirebaseUserManager.java
```diff
@@ -230,20 +230,28 @@ UserImportResult importUsers(UserImportRequest request) throws FirebaseAuthExcep
...
+    GenericUrl url = new GenericUrl(tenantMgtBaseUrl + "/tenants");
+    url.putAll(request.getProperties());
+    return sendRequest("POST", url, null, Tenant.class);
+  }
+
+  Tenant updateTenant(Tenant.UpdateRequest request) throws FirebaseAuthException {
+    Map<String, Object> properties = request.getProperties();
+    checkArgument(properties.size() > 1,
+        "tenant update must have at least one property other than tenantId set");
+    GenericUrl url = new GenericUrl(tenantMgtBaseUrl + "/tenants/" + request.getTenantId());
```
hiranya911: You need to generate an `updateMask` from the given properties, and add it the URL here. See the corresponding Go and Node.js implementations.

micahstairs: Okay, interesting. So for the update request, we encode the properties 2 different ways.

I found the utility which builds the mask in Node (https://github.com/firebase/firebase-admin-node/blob/0f6c02e3377c3337e4f206e176b2d96ec6dd6c3c/src/utils/index.ts#L148).

Any idea if we already have a similar utility in Java? Or do I need to write one? From looking at Node's tests (https://github.com/firebase/firebase-admin-node/blob/0f6c02e3377c3337e4f206e176b2d96ec6dd6c3c/test/unit/utils/index.spec.ts#L307), it looks like the format we want is something like this:

...../tenants/TENANT_ID?updateMask=name.TENANT_ID,displayName.DISPLAY_NAME,allowPasswordSignup.true,enableEmailLinkSignin.false

Does that look right?

hiranya911: I don't think we have a util in Java for this at the moment. The field mask should only consist of field names in a dotted notation. Following json payload:

```
{
  'parentField1': 'some value',
  'parentField2': 'another value',
  'parentField3': {
    'childField1': 'yet another value'
  }
}
```

should become:
```
['parentField1', 'parentField2', 'parentField3.childField1']
```

You can also look at the unit tests of the Node.js utils for more examples.

src/main/java/com/google/firebase/auth/FirebaseUserManager.java

src/main/java/com/google/firebase/auth/FirebaseUserManager.java

src/main/java/com/google/firebase/auth/FirebaseUserManager.java

src/main/java/com/google/firebase/auth/FirebaseUserManager.java
```diff
@@ -230,20 +231,26 @@ UserImportResult importUsers(UserImportRequest request) throws FirebaseAuthExcep
...
+
+  Tenant createTenant(Tenant.CreateRequest request) throws FirebaseAuthException {
+    GenericUrl url = new GenericUrl(tenantMgtBaseUrl + "/tenants");
+    return sendRequest("POST", url, request.getProperties(), Tenant.class);
+  }
+
+  Tenant updateTenant(Tenant.UpdateRequest request) throws FirebaseAuthException {
+    Map<String, Object> properties = request.getProperties();
+    checkArgument(!properties.isEmpty(), "tenant update must have at least one property set");
+    String updateMask = String.join(",", ImmutableSortedSet.copyOf(properties.keySet()));
```
hiranya911: This will break if the request contains any nested properties (we don't have any at the moment). If the request is shaped like `{"foo": {"bar": 100}}`, the update mask should look like `['foo.bar']`.

Let's either implement a small util function to generate the correct update masks for all requests, or add a comment/TODO here to indicate the limitation with the current appraoch.

PR #386, opened 2020-04-01
Add firebase auth destroy check before tenant operations.

micahstairs:
This is part of the initiative to adding multi-tenancy support (see issue #332).

### DIFF
--- a/src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
+++ b/src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
@@ -1097,7 +1097,11 @@ public abstract class AbstractFirebaseAuth {
 
   protected void destroy() {
     synchronized (lock) {
+      doDestroy();
       destroyed.set(true);
     }
   }
+
+  /** Performs any additional required clean up. */
+  protected abstract void doDestroy();
 }
--- a/src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
+++ b/src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
@@ -1086,7 +1086,7 @@ public abstract class AbstractFirebaseAuth {
         });
   }
 
-  private void checkNotDestroyed() {
+  void checkNotDestroyed() {
     synchronized (lock) {
       checkState(
           !destroyed.get(),
@@ -1095,7 +1095,7 @@ public abstract class AbstractFirebaseAuth {
     }
   }
 
-  void destroy() {
+  protected void destroy() {
     synchronized (lock) {
       destroyed.set(true);
     }
--- a/src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
+++ b/src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
@@ -1086,7 +1086,7 @@ public abstract class AbstractFirebaseAuth {
         });
   }
 
-  void checkNotDestroyed() {
+  private void checkNotDestroyed() {
     synchronized (lock) {
       checkState(
           !destroyed.get(),
@@ -1095,7 +1095,7 @@ public abstract class AbstractFirebaseAuth {
     }
   }
 
-  protected void destroy() {
+  final void destroy() {
     synchronized (lock) {
       doDestroy();
       destroyed.set(true);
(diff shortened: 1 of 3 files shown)


micahstairs:
> I think we should try to manage this within `TenantManager` it self. Here's what I'd suggest:
> 
> 1. Add a `destroy()` method to the `TenantManager` class.
> 2. Call it from `FirebaseAuth.destroy()`.
> 3. Handle the `checkNotDestroyed()` check entirely within the `TenantManager` class.
> 
> I think that results in a cleaner flow of control, without `AbstractFirebaseAuth` leaking into `TenantManager`. WDYT?

By #3, I assume you mean "Call it from `FirebaseAuth.destroy()`".

I like the idea of not leaking the firebase auth into `TenantManager`, however, it's not straightforward to call `TenantManager.destroy()` method from `FirebaseAuth.destroy()`, is it? I would need to override `AbstractFirebaseAuth.destroy()`'s method in `FirebaseAuth`. I believe overriding this would require me to expose the lock that it is used to synchronize on.

micahstairs:
It's worth noting that I'm currently working on implementing TenantAwareFirebaseAuth, so perhaps that will shape our decision here.

hiranya911:
You can do something like the following in `AbstractFirebaseAuth`:

```
protected abstract void doDestroy();

void destroy() {
   // Call doDestroy() while holding the lock
}
```

Now `FirebaseAuth` can implement `doDestroy()` to implement any child-specific cleanup logic.

micahstairs:
> You can do something like the following in `AbstractFirebaseAuth`:
> 
> ```
> protected abstract void doDestroy();
> 
> void destroy() {
>    // Call doDestroy() while holding the lock
> }
> ```
> 
> Now `FirebaseAuth` can implement `doDestroy()` to implement any child-specific cleanup logic.
> 
> Basically, pass control in one direction: `AbstractFirebaseAuth --> FirebaseAuth --> TenantManager` and never in both directions.

Okay, this makes sense, but there's an issue. The TenantManager is lazily instantiated since it's a supplier. So if I call getTaskManager().destroy() and it hasn't been instantiated yet, then we end up invoking the destroyed app. The workaround (which I've pushed) isn't particularly elegant.

```
tenantManager = threadSafeMemoize(new Supplier<TenantManager>() {                               
      @Override                                                                                     
      public TenantManager get() {                                                                  
        return new TenantManager(builder.firebaseApp, getUserManager());                            
      }                                                                                             
});   
```

src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
```diff
@@ -1086,7 +1086,7 @@ public T get() {
         });
   }
 
-  private void checkNotDestroyed() {
+  void checkNotDestroyed() {
```
hiranya911, line 1089: I think this can remain private now.

src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
```diff
@@ -1095,9 +1095,13 @@ private void checkNotDestroyed() {
     }
   }
 
-  void destroy() {
+  protected void destroy() {
```
hiranya911, line 1098: Make this just `final void`?

src/main/java/com/google/firebase/auth/FirebaseAuth.java
```diff
@@ -53,7 +57,12 @@ public TenantManager get() {
   }
 
   public TenantManager getTenantManager() {
-    return tenantManager.get();
+    TenantManager suppliedTenantManager;
+    synchronized (lock) {
+      suppliedTenantManager = tenantManager.get();
+      tenantManagerCreated.set(true);
```
hiranya911, line 63: Set this to true in the `Supplier.get()` method (in the FirebaseAuth constructor). That method gets further wrapped in a `threadSafeMemoize` which provides the necessary locking around it.

micahstairs: Oh great idea! This looks much cleaner.

src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
micahstairs: Done.

src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
micahstairs: Done.

PR #389, opened 2020-04-10
Testable BatchResponse

### LINKED ISSUE #388: BatchResponse implementation makes testing difficult.
* Operating System version: N/A
  * Firebase SDK version: 2
  * Library version: 6.12.2
  * Firebase Product: Messaging

#### Description:

BatchResponse is implemented as a final class with a package-private constructor. As such, this means that as a consumer of the library I have to either wrap the response object or use reflection to make it testable.

#### Steps to reproduce:

```java
// This will not compile
new BatchResponse(responses);
```

#### Suggestions

I would be happy to submit a PR for this but I would like guidance on implementation preference.

a) Remove the final keyword from the BatchResponse class so tools like Mockito can mock the class.
b) Make the constructor public
c) Add a builder or some other method of instantiating the class. This doesn't seem like a useful change.

ToxicBakery:
Resolves #388 
- Converts batch response to an interface
- Moves batch response implementation to BatchResponseImpl
- Update FirebaseMessagingClientImpl to use BatchResponseImpl

### DIFF
--- a/src/main/java/com/google/firebase/messaging/BatchResponseImpl.java
+++ b/src/main/java/com/google/firebase/messaging/BatchResponseImpl.java
@@ -26,7 +26,7 @@ import java.util.List;
  * See {@link FirebaseMessaging#sendAll(List)} and {@link
  * FirebaseMessaging#sendMulticast(MulticastMessage)}.
  */
-public class BatchResponseImpl implements BatchResponse {
+class BatchResponseImpl implements BatchResponse {
 
   private final List<SendResponse> responses;
   private final int successCount;
--- a/src/main/java/com/google/firebase/messaging/BatchResponseImpl.java
+++ b/src/main/java/com/google/firebase/messaging/BatchResponseImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright  2019 Google Inc.
+ * Copyright  2020 Google Inc.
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/firebase/messaging/BatchResponseImpl.java
```diff
@@ -0,0 +1,58 @@
...
+import com.google.firebase.internal.NonNull;
+
+import java.util.List;
+
+/**
+ * Response from an operation that sends FCM messages to multiple recipients.
+ * See {@link FirebaseMessaging#sendAll(List)} and {@link
+ * FirebaseMessaging#sendMulticast(MulticastMessage)}.
+ */
+public class BatchResponseImpl implements BatchResponse {
```
hiranya911, line 29: Make this package-protected

ToxicBakery: Good call, apologies

hiranya911:
Hey @ToxicBakery. I just pushed a change to our CI config, so that they will be run on external contributions such as this PR. Can I ask you to please sync against the latest master branch, and make a push?

ToxicBakery:
Done, appears to have run successfully :+1:

PR #391, opened 2020-04-16
Add tenant-aware token generation and verification.

micahstairs:
This pull request incorporates the tenant ID into the token generation and validation when using a tenant-aware client. This is part of the initiative to adding multi-tenancy support (see issue #332).

### DIFF
--- a/src/main/java/com/google/firebase/auth/FirebaseTokenUtils.java
+++ b/src/main/java/com/google/firebase/auth/FirebaseTokenUtils.java
@@ -30,6 +30,7 @@ import com.google.firebase.FirebaseOptions;
 import com.google.firebase.ImplFirebaseTrampolines;
 import com.google.firebase.auth.internal.CryptoSigners;
 import com.google.firebase.auth.internal.FirebaseTokenFactory;
+import com.google.firebase.internal.Nullable;
 
 import java.io.IOException;
 
@@ -52,11 +53,17 @@ final class FirebaseTokenUtils {
   private FirebaseTokenUtils() { }
 
   static FirebaseTokenFactory createTokenFactory(FirebaseApp firebaseApp, Clock clock) {
+    return createTokenFactory(firebaseApp, clock, null);
+  }
+
+  static FirebaseTokenFactory createTokenFactory(
+      FirebaseApp firebaseApp, Clock clock, @Nullable String tenantId) {
     try {
       return new FirebaseTokenFactory(
           firebaseApp.getOptions().getJsonFactory(),
           clock,
-          CryptoSigners.getCryptoSigner(firebaseApp));
+          CryptoSigners.getCryptoSigner(firebaseApp),
+          tenantId);
     } catch (IOException e) {
       throw new IllegalStateException(
           "Failed to initialize FirebaseTokenFactory. Make sure to initialize the SDK "
--- a/src/main/java/com/google/firebase/auth/FirebaseTokenUtils.java
+++ b/src/main/java/com/google/firebase/auth/FirebaseTokenUtils.java
@@ -99,11 +99,6 @@ final class FirebaseTokenUtils {
   }
 
   static FirebaseTokenVerifierImpl createSessionCookieVerifier(FirebaseApp app, Clock clock) {
-    return createSessionCookieVerifier(app, clock, null);
-  }
-
-  static FirebaseTokenVerifierImpl createSessionCookieVerifier(
-      FirebaseApp app, Clock clock, @Nullable String tenantId) {
     String projectId = ImplFirebaseTrampolines.getProjectId(app);
     checkState(!Strings.isNullOrEmpty(projectId),
         "Must initialize FirebaseApp with a project ID to call verifySessionCookie()");
@@ -118,7 +113,6 @@ final class FirebaseTokenUtils {
         .setShortName("session cookie")
         .setMethod("verifySessionCookie()")
         .setDocUrl("https://firebase.google.com/docs/auth/admin/manage-cookies")
-        .setTenantId(tenantId)
         .build();
   }
 
--- a/src/main/java/com/google/firebase/auth/FirebaseTokenUtils.java
+++ b/src/main/java/com/google/firebase/auth/FirebaseTokenUtils.java
@@ -75,6 +75,11 @@ final class FirebaseTokenUtils {
   }
 
   static FirebaseTokenVerifierImpl createIdTokenVerifier(FirebaseApp app, Clock clock) {
+    return createIdTokenVerifier(app, clock, null);
+  }
+
+  static FirebaseTokenVerifierImpl createIdTokenVerifier(
+      FirebaseApp app, Clock clock, @Nullable String tenantId) {
     String projectId = ImplFirebaseTrampolines.getProjectId(app);
     checkState(!Strings.isNullOrEmpty(projectId),
         "Must initialize FirebaseApp with a project ID to call verifyIdToken()");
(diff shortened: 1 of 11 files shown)


src/main/java/com/google/firebase/auth/FirebaseTokenUtils.java
```diff
@@ -82,10 +94,16 @@ static FirebaseTokenVerifierImpl createIdTokenVerifier(FirebaseApp app, Clock cl
...
         .setIdTokenVerifier(idTokenVerifier)
+        .setTenantId(tenantId)
         .build();
   }
 
   static FirebaseTokenVerifierImpl createSessionCookieVerifier(FirebaseApp app, Clock clock) {
+    return createSessionCookieVerifier(app, clock, null);
+  }
+
+  static FirebaseTokenVerifierImpl createSessionCookieVerifier(
```
hiranya911, line 105: Session cookies do not support tenants at the moment. We can remove this.

src/main/java/com/google/firebase/auth/FirebaseTokenVerifierImpl.java
```diff
@@ -278,6 +283,18 @@ private boolean containsLegacyUidField(IdToken.Payload payload) {
     return false;
   }
 
+  private void checkTenantId(final FirebaseToken firebaseToken) throws FirebaseAuthException {
+    String tokenTenantId = firebaseToken.getTenantId();
+    if (!Strings.nullToEmpty(tokenTenantId).equals(Strings.nullToEmpty(tenantId))) {
```
hiranya911, line 288: Following might be a bit more readable:

```
if (!Strings.isNullOrEmpty(this.tenantId)) {
  if (!this.tenantId.equals(tokenTenantId) {
    // ...
  }
}
```

src/test/java/com/google/firebase/auth/FirebaseAuthIT.java
```diff
@@ -705,6 +706,58 @@ public void testCustomTokenWithIAM() throws Exception {
...
+    // Create and decode a token with a tenant-aware client.
+    TenantAwareFirebaseAuth tenantAwareAuth = auth.getTenantManager().getAuthForTenant(tenantId);
+    String customToken = tenantAwareAuth.createCustomTokenAsync("user1").get();
+    String idToken = signInWithCustomToken(customToken, tenantId);
+    FirebaseToken decoded = tenantAwareAuth.verifyIdTokenAsync(idToken).get();
+    assertEquals("user1", decoded.getUid());
+    assertEquals(tenantId, decoded.getTenantId());
+
+    // Delete tenant.
+    tenantManager.deleteTenantAsync(tenantId).get();
```
hiranya911, line 726: Might want to push this into a finally block

src/test/java/com/google/firebase/auth/FirebaseAuthIT.java
```diff
@@ -705,6 +706,58 @@ public void testCustomTokenWithIAM() throws Exception {
...
+      tenantAwareAuth2.verifyIdTokenAsync(idToken).get();
+      fail("No error thrown for verifying a token with the wrong tenant-aware client");
+    } catch (ExecutionException e) {
+      assertTrue(e.getCause() instanceof FirebaseAuthException);
+      assertEquals(FirebaseUserManager.TENANT_ID_MISMATCH,
+          ((FirebaseAuthException) e.getCause()).getErrorCode());
+    }
+
+    // Delete tenants.
+    tenantManager.deleteTenantAsync(tenantId1).get();
```
hiranya911, line 757: Same here

src/test/java/com/google/firebase/auth/internal/FirebaseTokenFactoryTest.java
```diff
@@ -18,6 +18,7 @@
 
 import static com.google.common.base.Preconditions.checkNotNull;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
```
hiranya911, line 21: Can we also have some unit tests in `FirebaseTokenVerifierImplTest`.

src/test/java/com/google/firebase/auth/FirebaseAuthIT.java
micahstairs: There's no existing try block to add the finally block to. Are you suggesting that I wrap the entire "Create and decode a token with a tenant-aware client" section with a try (and no catch)?

src/main/java/com/google/firebase/auth/FirebaseTokenVerifierImpl.java
micahstairs: That's more readable, yes, but it also behaves differently.

Don't we want to throw an exception if `this.tenantId` is null and `tokenTenantId` is non-empty?

src/main/java/com/google/firebase/auth/FirebaseTokenUtils.java
micahstairs: Done.

src/test/java/com/google/firebase/auth/FirebaseAuthIT.java
micahstairs: Done.

src/test/java/com/google/firebase/auth/internal/FirebaseTokenFactoryTest.java
micahstairs: Done.

src/test/java/com/google/firebase/auth/FirebaseAuthIT.java
hiranya911: Yes. You will see this pattern being used in a number of existing test cases. See `testCustomClaims()` for example.

src/main/java/com/google/firebase/auth/FirebaseTokenVerifierImpl.java
hiranya911: Ah I see your point. What if we do this in the constructor then:

```
this.tenantId = Strings.nullToEmpty(tenantId);
```

Then during verification you can do `!this.tenantId.equals(tokenTenantId)`.

micahstairs: Moving `Strings.nullToEmpty` to the constructor is a good idea.

But I think we'd need to do `!this.tenantId.equals(Strings.nullToEmpty(tokenTenantId)))` or else we'd say there was a mismatch if both were initially null.

micahstairs: I pulled the other `Strings.nullToEmpty` to the previous line where tokenTenantId is defined. I think this is better now.

src/test/java/com/google/firebase/auth/FirebaseAuthIT.java
micahstairs: Okay sounds good! Done.

src/main/java/com/google/firebase/auth/FirebaseTokenVerifierImpl.java
```diff
@@ -278,6 +283,18 @@ private boolean containsLegacyUidField(IdToken.Payload payload) {
     return false;
   }
 
+  private void checkTenantId(final FirebaseToken firebaseToken) throws FirebaseAuthException {
+    String tokenTenantId = Strings.nullToEmpty(firebaseToken.getTenantId());
```
hiranya911, line 287: Nit: nullToEmpty is not strictly needed here, since you call equals on `this.tenantId` which is never null. But I'll leave it to you decide.

src/test/java/com/google/firebase/auth/FirebaseAuthIT.java
```diff
@@ -705,6 +706,60 @@ public void testCustomTokenWithIAM() throws Exception {
...
+  }
+
+  @Test
+  public void testVerifyTokenWithWrongTenantAwareClient() throws Exception {
+    // Create tenants to use.
+    TenantManager tenantManager = auth.getTenantManager();
+    Tenant.CreateRequest tenantCreateRequest1 =
+        new Tenant.CreateRequest().setDisplayName("DisplayName1");
+    String tenantId1 = tenantManager.createTenant(tenantCreateRequest1).getTenantId();
+    Tenant.CreateRequest tenantCreateRequest2 =
```
hiranya911, line 738: You can only create one tenant for this test. For the 2nd tenant use a fake tenant ID string. And to ensure proper clean up you will have to do something like this:

```
// create test tenants
try {
   // create token
   try {
     // verify token
   }. catch {
     // assert
   }
} finally {
  // clean up tenants
}
```

You can move the inner try-catch into a new helper method if this looks too ugly.

micahstairs: Done.

src/main/java/com/google/firebase/auth/FirebaseTokenVerifierImpl.java
micahstairs: We actually do need it or else we'd say there was a mismatch if both were initially null. I confirmed this theory by making the change and watching `FirebaseTokenVerifierImplTest.testVerifyToken` fail.

PR #397, opened 2020-04-24
Add OIDC Auth provider config class

micahstairs:
This provider class only includes a `CreateRequest` inner class for the time being. An `UpdateRequest` inner class will be added in the future.

In addition to `OidcProviderConfig`, this pull request also includes a base class, which will be shared by a future `SamlProviderConfig` class.

This is part of adding multi-tenancy support (see issue #332).

### DIFF
--- a/src/main/java/com/google/firebase/auth/OidcProviderConfig.java
+++ b/src/main/java/com/google/firebase/auth/OidcProviderConfig.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/firebase/auth/OidcProviderConfig.java
```diff
@@ -0,0 +1,87 @@
...
+      return this;
+    }
+
+    /**
+     * Sets the issuer for the new provider.
+     *
+     * @param issuer a non-null, non-empty issuer string.
+     */
+    public CreateRequest setIssuer(String issuer) {
+      checkArgument(!Strings.isNullOrEmpty(issuer), "issuer must not be null or empty");
```
hiranya911, line 82: Also check for malformed URL.

micahstairs: Alright, I've added validation for this based off of the check I saw in ActionCodeSettings.java. I also added a quick unit test for this.

PR #399, opened 2020-04-27
Fix CreateRequest chaining and provider ID extraction.

micahstairs:
This makes the base CreateRequest setters return the proper instance type, so that methods can be chained. This also makes it so that the provider ID can be parsed from the resource name.

A package private getProviderId() method was also added, which will be needed by the FirebaseUserManager class when the provider config operations are added there.

### DIFF
--- a/src/main/java/com/google/firebase/auth/AuthProviderConfig.java
+++ b/src/main/java/com/google/firebase/auth/AuthProviderConfig.java
@@ -29,8 +29,11 @@ import java.util.Map;
  */
 public abstract class AuthProviderConfig {
 
-  @Key("name")
+  // Lazily initialized from 'resourceName'.
   private String providerId;
+
+  @Key("name")
+  private String resourceName;
 
   @Key("displayName")
   private String displayName;
@@ -39,6 +42,9 @@ public abstract class AuthProviderConfig {
   private boolean enabled;
 
   public String getProviderId() {
+    if (providerId == null) {
+      providerId = resourceName.substring(resourceName.lastIndexOf("/") + 1);
+    }
     return providerId;
   }
 
@@ -56,20 +62,25 @@ public abstract class AuthProviderConfig {
    * <p>Set the initial attributes of the new provider by calling various setter methods available
    * in this class.
    */
-  public abstract static class CreateRequest {
+  public abstract static class CreateRequest<T extends CreateRequest<T>> {
 
     final Map<String,Object> properties = new HashMap<>();
+    String providerId;
 
     /**
-     * Sets the ID for the new provider.
+     * Sets the display name for the new provider.
      *
      * @param providerId a non-null, non-empty provider ID string.
      */
-    public CreateRequest setProviderId(String providerId) {
+    public T setProviderId(String providerId) {
       checkArgument(
           !Strings.isNullOrEmpty(providerId), "provider ID name must not be null or empty");
-      properties.put("name", providerId);
-      return this;
+      this.providerId = providerId;
+      return getThis();
+    }
+
+    String getProviderId() {
+      return providerId;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/firebase/auth/AuthProviderConfig.java
```diff
@@ -56,45 +62,52 @@ public boolean isEnabled() {
    * <p>Set the initial attributes of the new provider by calling various setter methods available
    * in this class.
    */
-  public abstract static class CreateRequest {
+  public abstract static class CreateRequest<T extends CreateRequest<T>> {
```
hiranya911, line 65: Lets call this `AbstractCreateRequest` or `ProviderConfigCreateRequest` and reduce overusing the name `CreateRequest`.  This also doesn't need to be a nested class. Might be cleaner to put it elsewhere.

src/main/java/com/google/firebase/auth/AuthProviderConfig.java
```diff
@@ -29,16 +29,22 @@
  */
 public abstract class AuthProviderConfig {
 
-  @Key("name")
+  // Lazily initialized from 'resourceName'.
```
hiranya911, line 32: May be just remove this property altogether, and always return the calculated value from the getter. It's quite cheap to compute.

micahstairs: Okay! For consistency, I also made this change to the Tenant class.

src/main/java/com/google/firebase/auth/AuthProviderConfig.java
micahstairs: I like the idea of calling it AbstractCreateRequest instead of CreateRequest. However, I don't think we should move this into a separate class. Keeping it together is more consistent with all of our other CreateRequest/UpdateRequest classes.

On a somewhat related note, should AuthProviderConfig be renamed to AbstractProviderConfig? (Of course, the decision here might be influenced by our generics decision in #400.)

hiranya911: I prefer the shorter name `ProviderConfig` for this type (it's already nested in the `auth` package) so the "Auth" part is kind of redundant. Also may be this doesn't need to be an abstract class at all? Is there any behavior that child classes are expected to implement, or are they just adding more attributes to the parent type?

src/main/java/com/google/firebase/auth/OidcProviderConfig.java
```diff
@@ -53,7 +53,8 @@ public String getIssuer() {
    * <p>Set the initial attributes of the new provider by calling various setter methods available
    * in this class.
    */
-  public static final class CreateRequest extends AuthProviderConfig.CreateRequest<CreateRequest> {
+  public static final class CreateRequest
+      extends AuthProviderConfig.AbstractCreateRequest<CreateRequest> {
```
hiranya911, line 57: You can probably omit the `AuthProviderConfig.` now that the names are not ambiguous.

src/main/java/com/google/firebase/auth/AuthProviderConfig.java
micahstairs: I think renaming to `ProviderConfig` makes sense given that "Auth" is redundant.

However, I think we should keep it abstract for 2 reasons:
1) We never want to instantiate it on its own since it can't be used for anything.
2) If we make it concrete then we will be forced to implement `getThis` in `AuthProvider.AbstractCreateRequest`. The issue with this is that we then lose the ability to force its subclasses to implement `getThis`. If subclasses forget to override `getThis` then it would break how the chaining works.

src/main/java/com/google/firebase/auth/OidcProviderConfig.java
micahstairs: Done.

src/main/java/com/google/firebase/auth/AuthProviderConfig.java
hiranya911: Does 2 actually apply here? Why would the outer class be required to implement a method of an abstract nested class?

micahstairs: Sorry, you're correct. However, I think keeping it abstract makes it more clear that we do not intend on instantiating it on its own.

PR #400, opened 2020-04-28
Add operations to create and delete OIDC provider configs.

micahstairs:
This adds an operation to create OIDC provider configs (can be done using either the tenant-aware or standard Firebase client), as well as an operation to delete provider configs.

I spent some time trying to see if we could make the create operation generic so that it could support both OIDC and SAML provider configs, however, I'm concerned that such a solution would add complexity to the external API, rather than making it simpler. I think it would also introduce some sort of `instanceof` check internally.

### DIFF
--- a/src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
+++ b/src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
@@ -31,8 +31,7 @@ import com.google.firebase.auth.FirebaseUserManager.EmailLinkType;
 import com.google.firebase.auth.FirebaseUserManager.UserImportRequest;
 import com.google.firebase.auth.ListUsersPage.DefaultUserSource;
 import com.google.firebase.auth.ListUsersPage.PageFactory;
-import com.google.firebase.auth.UserRecord.CreateRequest;
-import com.google.firebase.auth.UserRecord.UpdateRequest;
+import com.google.firebase.auth.UserRecord;
 import com.google.firebase.auth.internal.FirebaseTokenFactory;
 import com.google.firebase.internal.CallableOperation;
 import com.google.firebase.internal.NonNull;
@@ -320,7 +319,8 @@ public abstract class AbstractFirebaseAuth {
       @Override
       protected Void execute() throws FirebaseAuthException {
         int currentTimeSeconds = (int) (System.currentTimeMillis() / 1000);
-        UpdateRequest request = new UpdateRequest(uid).setValidSince(currentTimeSeconds);
+        UserRecord.UpdateRequest request =
+            new UserRecord.UpdateRequest(uid).setValidSince(currentTimeSeconds);
         userManager.updateUser(request, jsonFactory);
         return null;
       }
@@ -512,32 +512,33 @@ public abstract class AbstractFirebaseAuth {
 
   /**
    * Creates a new user account with the attributes contained in the specified {@link
-   * CreateRequest}.
+   * UserRecord.CreateRequest}.
    *
-   * @param request A non-null {@link CreateRequest} instance.
+   * @param request A non-null {@link UserRecord.CreateRequest} instance.
    * @return A {@link UserRecord} instance corresponding to the newly created account.
    * @throws NullPointerException if the provided request is null.
    * @throws FirebaseAuthException if an error occurs while creating the user account.
    */
-  public UserRecord createUser(@NonNull CreateRequest request) throws FirebaseAuthException {
+  public UserRecord createUser(@NonNull UserRecord.CreateRequest request)
+      throws FirebaseAuthException {
     return createUserOp(request).call();
   }
 
   /**
-   * Similar to {@link #createUser(CreateRequest)} but performs the operation asynchronously.
+   * Similar to {@link #createUser} but performs the operation asynchronously.
    *
-   * @param request A non-null {@link CreateRequest} instance.
+   * @param request A non-null {@link UserRecord.CreateRequest} instance.
    * @return An {@code ApiFuture} which will complete successfully with a {@link UserRecord}
    *     instance corresponding to the newly created account. If an error occurs while creating the
    *     user account, the future throws a {@link FirebaseAuthException}.
    * @throws NullPointerException if the provided request is null.
    */
-  public ApiFuture<UserRecord> createUserAsync(@NonNull CreateRequest request) {
+  public ApiFuture<UserRecord> createUserAsync(@NonNull UserRecord.CreateRequest request) {
     return createUserOp(request).callAsync(firebaseApp);
   }
 
   private CallableOperation<UserRecord, FirebaseAuthException> createUserOp(
-      final CreateRequest request) {
+      final UserRecord.CreateRequest request) {
     checkNotDestroyed();
     checkNotNull(request, "create request must not be null");
     final FirebaseUserManager userManager = getUserManager();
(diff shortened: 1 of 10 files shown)


src/main/java/com/google/firebase/auth/FirebaseUserManager.java
```diff
@@ -65,6 +65,7 @@
  */
 class FirebaseUserManager {
 
+  static final String CONFIGURATION_NOT_FOUND = "project-not-found";
```
micahstairs, line 68: I find this mapping strange, however, "CONFIGURATION_NOT_FOUND" was already being mapped to "project-not-found".

I've confirmed that the server returns CONFIGURATION_NOT_FOUND if you ask it to delete a provider config ID which doesn't exist.

src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
```diff
@@ -945,6 +936,98 @@ protected String execute() throws FirebaseAuthException {
...
+   * Creates a new provider OIDC Auth config with the attributes contained in the specified {@link
+   * OidcProviderConfig.CreateRequest}.
+   *
+   * @param request A non-null {@link OidcProviderConfig.CreateRequest} instance.
+   * @return An {@link OidcProviderConfig} instance corresponding to the newly created provider
+   *     config.
+   * @throws NullPointerException if the provided request is null.
+   * @throws FirebaseAuthException if an error occurs while creating the provider config.
+   */
+  public OidcProviderConfig createOidcProviderConfig(
```
hiranya911, line 949: This is different from the signature proposed in go/firebase-mt-java. That is fine, but we should think if this is the best we can do. In .NET we used a generics-based solution for this. Can we do something similar here? Something like the following:

```
<T extends ProviderConfig> T createProviderConfig(ProviderConfig.CreateRequest<T> request)
```

WDYT?

src/main/java/com/google/firebase/auth/FirebaseUserManager.java
hiranya911: Already mapped where? Looks like it's being added here.

src/main/java/com/google/firebase/auth/FirebaseUserManager.java
```diff
@@ -106,6 +107,7 @@
   private static final String CLIENT_VERSION_HEADER = "X-Client-Version";
 
   private final String userMgtBaseUrl;
+  private final String oidpMgtBaseUrl;
```
hiranya911, line 110: idpConfigMgtBaseUrl

src/test/java/com/google/firebase/auth/FirebaseUserManagerTest.java
```diff
@@ -1383,6 +1386,140 @@ public void testUnexpectedHttpError() {
...
+        TestUtils.loadResource("oidc.json"));
+    // Only the 'enabled' field can be omitted from an OIDC provider config creation request.
+    OidcProviderConfig.CreateRequest createRequest =
+        new OidcProviderConfig.CreateRequest()
+            .setProviderId("oidc.provider-id")
+            .setDisplayName("DISPLAY_NAME")
+            .setClientId("CLIENT_ID")
+            .setIssuer("https://oidc.com/issuer");
+
+    FirebaseAuth.getInstance().createOidcProviderConfig(createRequest);
```
hiranya911, line 1427: Lets add some assertions here.

src/main/java/com/google/firebase/auth/FirebaseUserManager.java
micahstairs: It was already in the map (`.put("CONFIGURATION_NOT_FOUND", "project-not-found")`), I just gave the constant a name so that it could be referenced.

src/main/java/com/google/firebase/auth/FirebaseUserManager.java

src/test/java/com/google/firebase/auth/FirebaseUserManagerTest.java
micahstairs: Done. Initially, I was thinking that the purpose of the test was to ensure that we didn't mistakenly disallow a create request missing "enabled", however, now I see that can also make meaningful assertions about the request that we are sending.

src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
micahstairs: It's unfortunately not as simple as this. The class signature of AuthProviderConfig is `public abstract static class CreateRequest<T extends CreateRequest<T>>`, not `public abstract static class CreateRequest<T extends AuthProviderConfig<T>>`. In order to make this work, I think we'd need to add another (complex) generic type to `AuthProviderConfig.CreateRequest`. As I mentioned in my PR description, the accumulated complexity outweighed the benefits in my opinion.

For more context about why `CreateRequest<T extends CreateRequest<T>>` is necessary (as opposed to the more intuitive `CreateRequest<T extends AuthProviderConfig>`), see https://stackoverflow.com/a/17165079.

src/main/java/com/google/firebase/auth/FirebaseUserManager.java
hiranya911: I think we ought to fix it. I don't think any of the existing APIs return this error code, so it should be safe to change it to "configuration-not-found".

src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
hiranya911: Wouldn't this work?

```
class CreateRequest<T extends CreateRequest<T>, U extends AuthProviderConfig> {
 
}
```

src/main/java/com/google/firebase/auth/FirebaseUserManager.java
micahstairs: Okay! Done.

src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
micahstairs: Yes, I think that might work. That's the idea I was referring to when I mentioned adding another (complex) generic type to AuthProviderConfig.CreateRequest.

My concern is that this makes the external API too complex. Wouldn't this make the signature for `createProviderConfig` pretty wild?

hiranya911: That's a good point. I guess the best we can do in that case is something like:

```
public <T extends ProviderConfig> T createProviderConfig(AbstractCreateRequest<?, T> request)
```

Users won't have to worry too much about this as they can simply do `createProviderConfig(new OidcProviderConfig.CreateRequest())` most of the time, but `createOidcProviderConfig(request)` is far simpler to wrap one's head around.

Ok, I think I'm sufficiently convinced. Let's go ahead with separate create methods for the 2 types.

micahstairs: Great! Glad we are on the same page now.

micahstairs:
I've merged the other PR into this and changed the base branch to idp-config.

src/main/java/com/google/firebase/auth/FirebaseUserManager.java
```diff
@@ -120,15 +122,18 @@
...
+    final String tenantId = builder.tenantId;
+    if (tenantId == null) {
+      this.userMgtBaseUrl = idToolkitUrlV1;
+      this.idpConfigMgtBaseUrl = idToolkitUrlV2 + "/oauthIdpConfigs";
     } else {
       checkArgument(!tenantId.isEmpty(), "tenant ID must not be empty");
-      this.userMgtBaseUrl =
-          String.format(ID_TOOLKIT_URL, "v1", projectId) + getTenantUrlSuffix(tenantId);
+      this.userMgtBaseUrl = idToolkitUrlV1 + getTenantUrlSuffix(tenantId);
+      this.idpConfigMgtBaseUrl = idToolkitUrlV2 + getTenantUrlSuffix(tenantId) + "/oauthIdpConfigs";
```
hiranya911, line 134: SAML configs are on a different URL path. Better to not append `/oauthIdpConfigs` to the URL here.

PR #404, opened 2020-05-04
Add operation to list OIDC provider configs.

micahstairs:
This adds an operation to list OIDC provider configs (can be done using either the tenant-aware or standard Firebase client). This is the last OIDC provider config operation that needs to be implemented.

Please take note of the generics I have chosen to use in ListProviderConfigsPage. This should help reduce a lot of code duplication when it comes time to add support for SAML provider configs.

### DIFF
--- a/src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
+++ b/src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
@@ -29,10 +29,10 @@ import com.google.common.base.Suppliers;
 import com.google.firebase.FirebaseApp;
 import com.google.firebase.auth.FirebaseUserManager.EmailLinkType;
 import com.google.firebase.auth.FirebaseUserManager.UserImportRequest;
+import com.google.firebase.auth.ListProviderConfigsPage;
 import com.google.firebase.auth.ListProviderConfigsPage.DefaultOidcProviderConfigSource;
-import com.google.firebase.auth.ListProviderConfigsPage.ProviderConfigPageFactory;
+import com.google.firebase.auth.ListUsersPage;
 import com.google.firebase.auth.ListUsersPage.DefaultUserSource;
-import com.google.firebase.auth.ListUsersPage.UserPageFactory;
 import com.google.firebase.auth.UserRecord;
 import com.google.firebase.auth.internal.FirebaseTokenFactory;
 import com.google.firebase.internal.CallableOperation;
@@ -503,7 +503,7 @@ public abstract class AbstractFirebaseAuth {
     checkNotDestroyed();
     final FirebaseUserManager userManager = getUserManager();
     final DefaultUserSource source = new DefaultUserSource(userManager, jsonFactory);
-    final UserPageFactory factory = new UserPageFactory(source, maxResults, pageToken);
+    final ListUsersPage.Factory factory = new ListUsersPage.Factory(source, maxResults, pageToken);
     return new CallableOperation<ListUsersPage, FirebaseAuthException>() {
       @Override
       protected ListUsersPage execute() throws FirebaseAuthException {
@@ -1035,6 +1035,7 @@ public abstract class AbstractFirebaseAuth {
 
   /**
    * Similar to {@link #getOidcProviderConfig(String)} but performs the operation asynchronously.
+   * Page size will be limited to 100 provider configs.
    *
    * @param providerId A provider ID string.
    * @return An {@code ApiFuture} which will complete successfully with an
@@ -1079,7 +1080,7 @@ public abstract class AbstractFirebaseAuth {
 
   /**
    * Similar to {@link #listlistOidcProviderConfigs(String)} but performs the operation
-   * asynchronously.
+   * asynchronously. Page size will be limited to 100 provider configs.
    *
    * @param pageToken A non-empty page token string, or null to retrieve the first page of provider
    *     configs.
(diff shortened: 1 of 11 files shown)


src/main/java/com/google/firebase/auth/FirebaseUserManager.java
```diff
@@ -94,6 +95,7 @@
       .put("INVALID_DYNAMIC_LINK_DOMAIN", "invalid-dynamic-link-domain")
       .build();
 
+  static final int MAX_LIST_PROVIDER_CONFIGS_RESULTS = 100;
```
micahstairs, line 98: Is 100 the right value to use as a maximum? The Go implementation uses 100 but I don't think the Node implementation has a limit at all. It's not clear what limit the server enforces.

src/main/java/com/google/firebase/auth/ListUsersPage.java
```diff
@@ -237,17 +238,17 @@ String getNextPageToken() {
    * before attempting to load any user data (which is expensive, and hence may be performed
    * asynchronously on a separate thread).
    */
-  static class PageFactory {
+  static class UserPageFactory {
```
hiranya911, line 241: May be just call this `Factory` and use the full qualified name of `ListUsersPage.Factory` at callsites?

src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
```diff
@@ -1058,6 +1060,78 @@ protected OidcProviderConfig execute() throws FirebaseAuthException {
...
+   * @throws IllegalArgumentException If the specified page token is empty, or max results value is
+   *     invalid.
+   * @throws FirebaseAuthException If an error occurs while retrieving user data.
+   */
+  public ListProviderConfigsPage<OidcProviderConfig> listOidcProviderConfigs(
+        @Nullable String pageToken, int maxResults) throws FirebaseAuthException {
+    return listOidcProviderConfigsOp(pageToken, maxResults).call();
+  }
+
+  /**
```
hiranya911, line 1080: Mention the default maxResults in the docs.

src/test/java/com/google/firebase/auth/ListUsersPageTest.java
```diff
@@ -45,7 +46,7 @@
   @Test
   public void testSinglePage() throws FirebaseAuthException, IOException {
     TestUserSource source = new TestUserSource(3);
-    ListUsersPage page = new ListUsersPage.PageFactory(source).create();
+    ListUsersPage page = new UserPageFactory(source).create();
```
hiranya911, line 49: Lets see if we can keep this as is, or just rename to `ListUsersPage.Factory`.

src/test/java/com/google/firebase/auth/FirebaseAuthIT.java
```diff
@@ -1066,6 +1066,137 @@ public void testTenantAwareOidcProviderConfigLifecycle() throws Exception {
...
+      }
+      assertEquals(providerIds.size(), collected.get());
+
+      // Test iterate all
+      collected.set(0);
+      page = auth.listOidcProviderConfigsAsync(null).get();
+      for (OidcProviderConfig providerConfig : page.iterateAll()) {
+        if (providerIds.contains(providerConfig.getProviderId())) {
+          collected.incrementAndGet();
+          assertEquals("CLIENT_ID", providerConfig.getClientId());
```
hiranya911, line 1107: Define a helper to reduce some of this duplication.

```
if (checkProviderConfig(providerIds, providerConfig) {
  collected.incrementAndGet();
}

private boolean checkProviderConfig(List<String> providerIds, OidcProviderConfig config) {
  
}
```

src/test/java/com/google/firebase/auth/FirebaseAuthIT.java
```diff
@@ -1066,6 +1066,137 @@ public void testTenantAwareOidcProviderConfigLifecycle() throws Exception {
...
+    } finally {
+      // Delete provider configs
+      for (String providerId : providerIds) {
+        auth.deleteProviderConfigAsync(providerId).get();
+      }
+    }
+  }
+
+  @Test
+  public void testTenantAwareListOidcProviderConfigs() throws Exception {
```
hiranya911, line 1152: I feel like we are creating so many throw away tenants in these tests. I'd suggest that we move all tenant-aware tests to a new class. We can just create one test tenant in a `@BeforeClass` and reuse it as much as possible. Will make the overall logic a lot clear that way too.

But feel free to do that in a separate PR.

micahstairs: This is a really good idea! I think this would clean things up a lot. I've added a TODO to address in a separate PR (since it will involve moving quite a bit of code).

src/main/java/com/google/firebase/auth/ListUsersPage.java
micahstairs: Good idea!

src/test/java/com/google/firebase/auth/ListUsersPageTest.java
micahstairs: Renamed to `ListUsersPage.Factory`.

src/test/java/com/google/firebase/auth/FirebaseAuthIT.java
micahstairs: This doesn't save too many lines, but sure.

src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
micahstairs: Done. I followed the same pattern as in the list user methods.

PR #409, opened 2020-05-11
fix(auth): Ensuring test user account cleanup with a Rule

hiranya911:
Using a JUnit `@Rule` to ensure that temporary user accounts are cleaned up in integration tests. This helps us avoid large try-finally blocks in tests.

I also moved couple of new tenant management tests to a new `TenantManagerIT` class.

### DIFF
--- a/src/test/java/com/google/firebase/auth/FirebaseAuthIT.java
+++ b/src/test/java/com/google/firebase/auth/FirebaseAuthIT.java
@@ -63,8 +63,9 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 
-import org.junit.BeforeClass;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.ExternalResource;
 
 public class FirebaseAuthIT {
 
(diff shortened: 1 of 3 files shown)


src/test/java/com/google/firebase/auth/FirebaseAuthIT.java
```diff
@@ -80,13 +81,11 @@
...
+  private static final FirebaseAuth auth = FirebaseAuth.getInstance(
+      IntegrationTestUtils.ensureDefaultApp());
 
-  @BeforeClass
-  public static void setUpClass() {
-    FirebaseApp masterApp = IntegrationTestUtils.ensureDefaultApp();
-    auth = FirebaseAuth.getInstance(masterApp);
-  }
+  @Rule
+  public final TemporaryUser temporaryUser = new TemporaryUser();
```
micahstairs, line 88: Are you planning on adding a similar rule for tenants or should I take care of that?

micahstairs: I might as well do it so that we don't step on each others' toes. I'm also planning on doing this for the provider configs.

hiranya911: I didn't find that many instances where we create and delete provider configs. But if that's something we do repeatedly, then might be useful to make a similar change.

src/test/java/com/google/firebase/auth/TenantManagerIT.java
```diff
@@ -0,0 +1,155 @@
+/*
+ * Copyright 2020 Google Inc.
```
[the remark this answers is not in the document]
hiranya911: Done

src/test/java/com/google/firebase/auth/FirebaseAuthIT.java
micahstairs: It's not only a matter of reducing how many times we repeat the setup/cleanup. As it currently stands, if an assertion fails right after creating the provider config, it hasn't entered the "try" block yet, so the "finally" won't be executed. So we would either need to extend the try block back further, or we can use temporary provider configs. The temporary objects are nice because it is less of a burden on the developer writing integration tests, and it reduces the required nesting.

PR #411, opened 2020-05-12
Rename OIDC delete operation and refactor integration tests.

micahstairs:
I've renamed the delete operation since we need separate methods for deleting OIDC and SAML provider configs.

I've also created a ProviderConfigTestUtils class to house shared logic between `TenantAwareFirebaseAuthIT` and `FirebaseAuthIT`, and I've added a `TemporaryProviderConfig` class to make it easier for provider configs to be cleaned up automatically. I've structured this class in such a way that we can easily tack on logic to cleanup SAML provider configs as well.

I've also decided to remove `testGetUserWithMultipleTenantIds`. I initially wrote this test to understand how tenant-aware auths and tenant-agnostic auths interacted with one another. In its existing state, it appears to be leaking created users, and in order to resolve this with `TemporaryUser`, we would need to declare multiple `TemporaryUser` objects as rules. I think it's better to avoid this complexity and remove this test. After all, this integration test is mainly making assertions about the server's behavior and not our behavior.

NOTE: I also plan on creating a UserTestUtils class to reduce some code duplication between integration tests, but this will be in a separate PR.

### DIFF
--- a/src/test/java/com/google/firebase/auth/FirebaseAuthIT.java
+++ b/src/test/java/com/google/firebase/auth/FirebaseAuthIT.java
@@ -46,6 +46,7 @@ import com.google.common.util.concurrent.MoreExecutors;
 import com.google.firebase.FirebaseApp;
 import com.google.firebase.FirebaseOptions;
 import com.google.firebase.ImplFirebaseTrampolines;
+import com.google.firebase.auth.ProviderConfigTestUtils.TemporaryProviderConfig;
 import com.google.firebase.auth.hash.Scrypt;
 import com.google.firebase.internal.Nullable;
 import com.google.firebase.testing.IntegrationTestUtils;
@@ -84,8 +85,9 @@ public class FirebaseAuthIT {
   private static final FirebaseAuth auth = FirebaseAuth.getInstance(
       IntegrationTestUtils.ensureDefaultApp());
 
-  @Rule
-  public final TemporaryUser temporaryUser = new TemporaryUser();
+  @Rule public final TemporaryUser temporaryUser = new TemporaryUser();
+  @Rule public final TemporaryProviderConfig temporaryProviderConfig =
+      new TemporaryProviderConfig(auth);
 
   @Test
   public void testGetNonExistingUser() throws Exception {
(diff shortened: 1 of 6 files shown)


src/test/java/com/google/firebase/auth/ProviderConfigTestUtils.java
```diff
@@ -0,0 +1,74 @@
...
+  static final class TemporaryProviderConfig extends ExternalResource {
+
+    private final AbstractFirebaseAuth auth;
+    private final List<String> oidcIds = new ArrayList<>();
+
+    TemporaryProviderConfig(AbstractFirebaseAuth auth) {
+      this.auth = auth;
+    }
+
+    OidcProviderConfig createOidcProviderConfig(OidcProviderConfig.CreateRequest request)
```
hiranya911, line 54: Nit: synchronize this for consistency (so that all access to the underlying List is threadsafe).

src/test/java/com/google/firebase/auth/FirebaseAuthIT.java
```diff
@@ -563,124 +565,113 @@ public void testGenerateSignInWithEmailLink() throws Exception {
...
+    config = auth.updateOidcProviderConfigAsync(updateRequest).get();
+    assertEquals(providerId, config.getProviderId());
+    assertEquals("NewDisplayName", config.getDisplayName());
+    assertFalse(config.isEnabled());
+    assertEquals("NewClientId", config.getClientId());
+    assertEquals("https://oidc.com/new-issuer", config.getIssuer());
 
-    assertOidcProviderConfigDoesNotExist(auth, providerId);
+    // Delete config provider
+    auth.deleteOidcProviderConfigAsync(providerId).get();
```
hiranya911, line 604: Should we also expose a delete method on the `TemporaryProviderConfig` class so that it can drop the provider ID from the List? Otherwise we'd be making an extra rpc to delete an already deleted provider config.

src/test/java/com/google/firebase/auth/TenantAwareFirebaseAuthIT.java
```diff
@@ -61,20 +64,21 @@
   private static final JsonFactory jsonFactory = Utils.getDefaultJsonFactory();
   private static final HttpTransport transport = Utils.getDefaultTransport();
 
-  private static FirebaseAuth auth;
   private static TenantManager tenantManager;
-  private static TenantAwareFirebaseAuth tenantAwareAuth;
+  private static TenantAwareFirebaseAuth auth;
```
hiranya911, line 68: Nit: I kindda liked the more descriptive name `tenantAwareAuth` here. But your call.

src/test/java/com/google/firebase/auth/TenantAwareFirebaseAuthIT.java
```diff
@@ -297,15 +270,15 @@ public void testOidcProviderConfigLifecycle() throws Exception {
             .setEnabled(true)
             .setClientId("ClientId")
             .setIssuer("https://oidc.com/issuer");
-    OidcProviderConfig config = tenantAwareAuth.createOidcProviderConfigAsync(createRequest).get();
+    OidcProviderConfig config = auth.createOidcProviderConfigAsync(createRequest).get();
```
hiranya911, line 273: Use the tempProviderConfig here and unroll the try-finally?

src/test/java/com/google/firebase/auth/TenantAwareFirebaseAuthIT.java
micahstairs: Done.

src/test/java/com/google/firebase/auth/TenantAwareFirebaseAuthIT.java
micahstairs: Whoops! Thanks for catching that.

src/test/java/com/google/firebase/auth/FirebaseAuthIT.java
micahstairs: Okay sure, but we will need to synchronize on the list instead of synchronizing on the method in order to ensure that we don't try to create and delete at the same time.

src/test/java/com/google/firebase/auth/ProviderConfigTestUtils.java
micahstairs: Done.

src/test/java/com/google/firebase/auth/FirebaseAuthIT.java
hiranya911: I'd go for simplicity and just synchronize all the methods that modify the ID List. Performance issue should not really concern at this scale, especially for a test case.

micahstairs: This wasn't a performance optimization. We have to synchronize on the ID list or else we may run into concurrency issues if we delete and add to the list at the same time.

micahstairs: Ah it turns out that I misunderstood how the `synchronized` keyword worked on methods!

PR #413, opened 2020-05-14
[chore] Release 6.13.0

hiranya911:


pom.xml
```diff
@@ -19,7 +19,7 @@
 
     <groupId>com.google.firebase</groupId>
     <artifactId>firebase-admin</artifactId>
-    <version>6.12.3-SNAPSHOT</version>
+    <version>6.13.0</version>
```
lahirumaramba, line 22: I assume we are dropping the `next version-SNAPSHOT` pattern with the new release process, right?

hiranya911: Correct.

PR #418, opened 2020-05-16
Make Javadoc comments consistent and validate provider ID in CreateRequest.

micahstairs:
This makes the formatting of the Javadoc comments more consistent, and it's now more clear when/why `IllegalArgumentException`s get thrown. This also adds validation to the provider ID in the `CreateRequest` to ensure that it has the correct prefix.'

Improvements were also made to the tests. The JSON string is now more readable, the parameters are now in the correct order for `assertEquals` calls, and the `IllegalArgumentException` tests have been properly fleshed out.

### DIFF
--- a/src/main/java/com/google/firebase/auth/OidcProviderConfig.java
+++ b/src/main/java/com/google/firebase/auth/OidcProviderConfig.java
@@ -124,8 +124,8 @@ public final class OidcProviderConfig extends ProviderConfig {
      * information persistently.
      *
      * @param tenantId A non-null, non-empty provider ID string.
-     * @throws IllegalArgumentException If the provider ID is null or empty, or if the format is
-     *     invalid.
+     * @throws IllegalArgumentException If the provider ID is null or empty, or is not prefixed with
+     *     "oidc.".
      */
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/firebase/auth/OidcProviderConfig.java
```diff
@@ -116,9 +123,9 @@ CreateRequest getThis() {
      * {@link AbstractFirebaseAuth#updateOidcProviderConfig(CreateRequest)} to update the provider
      * information persistently.
      *
-     * @param tenantId a non-null, non-empty provider ID string.
-     * @throws IllegalArgumentException If the provider ID is null or empty, or if it's an invalid
-     *     format
+     * @param tenantId A non-null, non-empty provider ID string.
+     * @throws IllegalArgumentException If the provider ID is null or empty, or if the format is
+     *     invalid.
```
hiranya911, line 128: Nit: Perhaps specify the prefix (oidc.*) requirement

micahstairs: Good idea! Done.

PR #419, opened 2020-05-16
Add class for SAML provider config.

micahstairs:
Adds a class for the SAML provider config.


src/main/java/com/google/firebase/auth/SamlProviderConfig.java
```diff
@@ -0,0 +1,212 @@
...
+  public static class IdpCertificate {
+    @Key("x509Certificate")
+    private String x509Certificate;
+
+    public String getX509Certificate() {
+      return x509Certificate;
+    }
+  }
+
+  public static class IdpConfig {
```
hiranya911, line 174: These should be kept out of the public API.

src/main/java/com/google/firebase/auth/SamlProviderConfig.java
```diff
@@ -0,0 +1,212 @@
...
+    CreateRequest getThis() {
+      return this;
+    }
+
+    void assertValidProviderIdFormat(String providerId) {
+      checkArgument(providerId.startsWith("saml."), "Invalid SAML provider ID: " + providerId);
+    }
+  }
+
+  public static class IdpCertificate {
```
hiranya911, line 165: Shouldn't be public.

src/main/java/com/google/firebase/auth/ProviderConfig.java
```diff
@@ -60,6 +62,24 @@ static void assertValidUrl(String url) throws IllegalArgumentException {
     }
   }
 
+  static List<Object> getNestedList(Map<String, Object> outerMap, String id) {
```
hiranya911, line 65: Given these are only used in `SamlProviderConfig`, how about pushing them down to the child class?

src/main/java/com/google/firebase/auth/ProviderConfig.java
```diff
@@ -60,6 +62,24 @@ static void assertValidUrl(String url) throws IllegalArgumentException {
...
+    List<Object> list = (List<Object>) outerMap.get(id);
+    if (list == null) {
+      list = new ArrayList<Object>();
+      outerMap.put(id, list);
+    }
+    return list;
+  }
+
+  static Map<String, Object> getNestedMap(Map<String, Object> outerMap, String id) {
```
hiranya911, line 74: I think `ensureNestedMap()` and `ensureNestedList()` indicate the intent of these a bit better.

src/main/java/com/google/firebase/auth/SamlProviderConfig.java
```diff
@@ -0,0 +1,212 @@
...
+      return this;
+    }
+
+    /**
+     * Adds a x509 certificate to the new provider.
+     *
+     * @param x509Certificate A non-null, non-empty x509 certificate string.
+     * @throws IllegalArgumentException If the x509 certificate is null or empty.
+     */
+    public CreateRequest addX509Certificate(String x509Certificate) {
```
hiranya911, line 129: Nit: Move closer to the other `idpConfig` related methods

hiranya911, line 129: We should also probably have an `addAllX505Certificates(List<String>)` API here.

src/test/java/com/google/firebase/auth/SamlProviderConfigTest.java
```diff
@@ -0,0 +1,158 @@
...
+        + "    ]"
+        + "  },"
+        + "  'spConfig': {"
+        + "    'spEntityId':  'RP_ENTITY_ID',"
+        + "    'callbackUri': 'https://projectId.firebaseapp.com/__/auth/handler'"
+        + "   }"
+        + "}").replace("'", "\"");
+
+  @Test
+  public void testJsonSerialization() throws IOException {
```
hiranya911, line 58: Deserialization

src/main/java/com/google/firebase/auth/SamlProviderConfig.java
micahstairs: I assume you mean `addAllX509Certificates`? And yes, I already have a TODO to add that. I wanted to keep this PR to the bare essentials required.

src/test/java/com/google/firebase/auth/SamlProviderConfigTest.java
micahstairs: Done. I also made the same change in `OidcProviderConfigTest` and `TenantTest`.

src/main/java/com/google/firebase/auth/SamlProviderConfig.java
micahstairs: Done.

src/main/java/com/google/firebase/auth/ProviderConfig.java
micahstairs: Done.

src/main/java/com/google/firebase/auth/ProviderConfig.java
micahstairs: Done.

src/main/java/com/google/firebase/auth/SamlProviderConfig.java
micahstairs: Done.

src/main/java/com/google/firebase/auth/SamlProviderConfig.java
micahstairs: Done.

micahstairs: I had tried this during development but was getting a strange error. However, making this package private seems to work now (it must have been failing for another reason which I have since fixed).

src/main/java/com/google/firebase/auth/SamlProviderConfig.java
```diff
@@ -0,0 +1,230 @@
...
+  public String getRpEntityId() {
+    return spConfig.getRpEntityId();
+  }
+
+  public String getCallbackUrl() {
+    return spConfig.getCallbackUrl();
+  }
+
+  static List<Object> ensureNestedList(Map<String, Object> outerMap, String id) {
+    List<Object> list = (List<Object>) outerMap.get(id);
```
hiranya911, line 68: I think we are missing some imports here causing a compilation error.

src/main/java/com/google/firebase/auth/SamlProviderConfig.java
```diff
@@ -0,0 +1,230 @@
...
+
+  public String getRpEntityId() {
+    return spConfig.getRpEntityId();
+  }
+
+  public String getCallbackUrl() {
+    return spConfig.getCallbackUrl();
+  }
+
+  static List<Object> ensureNestedList(Map<String, Object> outerMap, String id) {
```
hiranya911, line 67: Can be private?

src/main/java/com/google/firebase/auth/SamlProviderConfig.java
micahstairs: So I guess we still were getting these errors (I was just running the tests in the wrong directory).

I've resolved the visibility problem by moving these subtypes into `SamlProviderConfigResponse`.

src/main/java/com/google/firebase/auth/SamlProviderConfig.java
micahstairs: Done.

src/main/java/com/google/firebase/auth/SamlProviderConfig.java
micahstairs: Done.

src/main/java/com/google/firebase/auth/SamlProviderConfig.java
```diff
@@ -0,0 +1,187 @@
...
+import java.util.Map;
+
+/**
+ * Contains metadata associated with a SAML Auth provider.
+ *
+ * <p>Instances of this class are immutable and thread safe.
+ */
+public final class SamlProviderConfig extends ProviderConfig {
+
+  @Key("idpConfig")
```
hiranya911, line 42: ~~Let's move these into the new `SamlProviderConfigResponse` class as well, and handle all deserialization business there. I think this class can just wrap an instance of the response.~~

**EDIT:** I suppose we can't do that since `ProviderConfig` already contains some deserialization logic (providerId, displayName etc). In that case let's define separate top-level classes in the internal package as follows:

* SamlProviderConfigResponse.IdpConfig --> SamlIdpConfig
* SamlProviderConfigResponse.SpConfig --> SamlSpConfig
* SamlProviderConfigResponse.IdpCertificate --> SamlIdpConfig.IdpCertificate

Btw, the option of parsing into maps, and handling the keys manually doesn't sound too onerous in this case either if you want to consider that option:

```
@Key("idpConfig")
private GenericJson idpConfig;

@Key("spConfig")
private GenericJson spConfig;

public String getIdpEntityId() {
    return (String) idpConfig.get("idpEntityId");
}
```

Looks a bit cleaner than adding a bunch of new internal types.

micahstairs: The GenericJson approach ended up working well! I thought it was going to be more complicated than it was, since we have nested types.

PR #422, opened 2020-05-18
chore: Merging master into tenant-mgt

hiranya911:
Merging current master branch into tenant-mgt.

@micahstairs please check if the multi-tenancy code has been affected in any adverse way.

@rsgowman please check if the bulk get/delete code has been affected in any adverse way. We will also need to add some tests for tenant-aware bulk get/delete in the future.


src/main/java/com/google/firebase/auth/FirebaseAuth.java
```diff
@@ -41,7 +40,7 @@
  * custom tokens for use by client-side code, verifying Firebase ID Tokens received from clients, or
  * creating new FirebaseApp instances that are scoped to a particular authentication UID.
  */
-public class FirebaseAuth extends AbstractFirebaseAuth {
+public final class FirebaseAuth extends AbstractFirebaseAuth {
```
micahstairs, line 43: Should we make `TenantAwareFirebaseAuth` final for consistency?

hiranya911: I think we should. Lets do that in a separate PR to keep this merge commit clean.

micahstairs: Alright! I'll slip that into `tenant-mgt` in a separate commit.

PR #424, opened 2020-05-26
Add operation to update SAML provider configs.

micahstairs:
Adds an update operation for SAML provider configs. I've included both unit tests and integration tests.

### DIFF
--- a/src/main/java/com/google/firebase/auth/FirebaseUserManager.java
+++ b/src/main/java/com/google/firebase/auth/FirebaseUserManager.java
@@ -54,6 +54,7 @@ import com.google.firebase.internal.SdkUtils;
 import java.io.IOException;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 /**
  * FirebaseUserManager provides methods for interacting with the Google Identity Toolkit via its
@@ -389,11 +390,11 @@ class FirebaseUserManager {
     sendRequest("DELETE", url, null, GenericJson.class);
   }
 
-  private static ImmutableSortedSet<String> generateMask(Map<String, Object> properties) {
+  private static Set<String> generateMask(Map<String, Object> properties) {
     ImmutableSortedSet.Builder<String> maskBuilder = ImmutableSortedSet.naturalOrder();
     for (Map.Entry<String, Object> entry : properties.entrySet()) {
       if (entry.getValue() instanceof Map) {
-        ImmutableSortedSet<String> childMask = generateMask((Map<String, Object>) entry.getValue());
+        Set<String> childMask = generateMask((Map<String, Object>) entry.getValue());
         for (String childProperty : childMask) {
           maskBuilder.add(entry.getKey() + "." + childProperty);
         }
--- a/src/main/java/com/google/firebase/auth/FirebaseUserManager.java
+++ b/src/main/java/com/google/firebase/auth/FirebaseUserManager.java
@@ -256,7 +256,7 @@ class FirebaseUserManager {
     // CallableOperation.
     checkArgument(!properties.isEmpty(), "tenant update must have at least one property set");
     GenericUrl url = new GenericUrl(tenantMgtBaseUrl + getTenantUrlSuffix(request.getTenantId()));
-    url.put("updateMask", generateMask(properties));
+    url.put("updateMask", Joiner.on(",").join(generateMask(properties)));
     return sendRequest("PATCH", url, properties, Tenant.class);
   }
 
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/firebase/auth/FirebaseUserManager.java
```diff
@@ -384,12 +389,19 @@ void deleteSamlProviderConfig(String providerId) throws FirebaseAuthException {
     sendRequest("DELETE", url, null, GenericJson.class);
   }
 
-  private static String generateMask(Map<String, Object> properties) {
-    // This implementation does not currently handle the case of nested properties. This is fine
-    // since we do not currently generate masks for any properties with nested values. When it
-    // comes time to implement this, we can check if a property has nested properties by checking
-    // if it is an instance of the Map class.
-    return Joiner.on(",").join(ImmutableSortedSet.copyOf(properties.keySet()));
+  private static ImmutableSortedSet<String> generateMask(Map<String, Object> properties) {
```
hiranya911, line 392: Declare return type to be `Set<String>` for extra flexibility.

src/test/java/com/google/firebase/auth/FirebaseUserManagerTest.java
```diff
@@ -1944,6 +1944,136 @@ public void testTenantAwareCreateSamlProvider() throws Exception {
...
+    GenericUrl url = interceptor.getResponse().getRequest().getUrl();
+    assertEquals(
+        "displayName,enabled,idpConfig.idpCertificates,idpConfig.idpEntityId,idpConfig.ssoUrl,"
+          + "spConfig.callbackUri,spConfig.spEntityId",
+        url.getFirst("updateMask"));
+
+    GenericJson parsed = parseRequestContent(interceptor);
+    assertEquals("DISPLAY_NAME", parsed.get("displayName"));
+    assertTrue((boolean) parsed.get("enabled"));
+    Map<String, Object> idpConfig = (Map<String, Object>) parsed.get("idpConfig");
```
hiranya911, line 1978: Some empty lines would make this a bit more pleasing to the eye. How about a newline just before this line?

src/test/java/com/google/firebase/auth/FirebaseUserManagerTest.java
```diff
@@ -1944,6 +1944,136 @@ public void testTenantAwareCreateSamlProvider() throws Exception {
...
+    assertNotNull(idpConfig);
+    assertEquals(3, idpConfig.size());
+    assertEquals("IDP_ENTITY_ID", idpConfig.get("idpEntityId"));
+    assertEquals("https://example.com/login", idpConfig.get("ssoUrl"));
+    List<Object> idpCertificates = (List<Object>) idpConfig.get("idpCertificates");
+    assertNotNull(idpCertificates);
+    assertEquals(2, idpCertificates.size());
+    assertEquals(ImmutableMap.of("x509Certificate", "certificate1"), idpCertificates.get(0));
+    assertEquals(ImmutableMap.of("x509Certificate", "certificate2"), idpCertificates.get(1));
+    Map<String, Object> spConfig = (Map<String, Object>) parsed.get("spConfig");
```
hiranya911, line 1988: Here too.

src/main/java/com/google/firebase/auth/FirebaseUserManager.java
micahstairs: Making the return type `Set<String>` does not give any extra flexibility (since this is not an interface). If this were an input parameter, we would certainly want to accept `Set` or even `Collection` instead of `ImmutableSortedSet`.

In Java, it's best to make the return type as specific as possible. In this case, indicating that the return value is both immutable and sorted is important. See go/java-practices/methods#type_generality.

src/test/java/com/google/firebase/auth/FirebaseUserManagerTest.java
micahstairs: Done.

src/test/java/com/google/firebase/auth/FirebaseUserManagerTest.java
micahstairs: Done.

src/main/java/com/google/firebase/auth/FirebaseUserManager.java
micahstairs: We spoke and decided to change it to `Set` in order to avoid Guava from leaking into the method signatures.

PR #425, opened 2020-05-27
fix(auth): Make user import hashing classes final

MathBunny:
While working on the .NET port of user imports I brought up the fact that these classes weren't final in the Java version; this PR revisits that idea.

API CHANGE: All public implementations of `UserImportHash` class are now marked `final`.


hiranya911:
Thanks @MathBunny. The PR looks good. Unfortunately this is also a breaking API change. So we have to wait until the next major release before we can merge this.

hiranya911:
@MathBunny can you change the base branch of this PR to `v7` (instead of `master`)? Then we can approve and merge it there.

MathBunny:
Looks like I need to rebase my branch, I'll look into this tomorrow

PR #427, opened 2020-05-27
fix: Removed unused FirebaseAppStore abstraction

hiranya911:
This is a bit of dead legacy code left over from the Android SDK, and has never been used in any useful or meaningful way. Therefore I'm removing it from the v7 release branch.


src/main/java/com/google/firebase/FirebaseApp.java
```diff
@@ -249,19 +241,15 @@ static void clearInstancesForTest() {
   }
 
   private static List<String> getAllAppNames() {
-    Set<String> allAppNames = new HashSet<>();
+    List<String> allAppNames = new ArrayList<>();
```
lahirumaramba, line 244: Was there a reason a `Set` was used here before? Could it be to avoid duplicates?

hiranya911: I think that's the idea, but it's redundant. App names are unique by definition. When adding a new app we check this:

```
checkState(
          !instances.containsKey(normalizedName),
          "FirebaseApp name " + normalizedName + " already exists!");
```

This actually gives me an idea. We can use the `instances.keySet()` to populate the List. I'll make the change.

lahirumaramba: Great! Makes sense to me.

PR #445, opened 2020-06-17
Fix documentation strings in Android Notification

lahirumaramba:
- Fix documentation strings
- Bye bye, `will` :)

### DIFF
--- a/src/main/java/com/google/firebase/messaging/AndroidConfig.java
+++ b/src/main/java/com/google/firebase/messaging/AndroidConfig.java
@@ -207,7 +207,7 @@ public class AndroidConfig {
     }
 
     /**
-     * Sets the direct_boot_ok flag, If set to true, messages are allowed to be delivered to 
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/firebase/messaging/AndroidConfig.java
```diff
@@ -207,7 +207,7 @@ public Builder setFcmOptions(AndroidFcmOptions androidFcmOptions) {
     }
 
     /**
-     * Sets the direct_boot_ok flag, If set to true, messages will be allowed to be delivered to 
+     * Sets the {@code direct_boot_ok} flag, If set to true, messages are delivered to
```
hiranya911, line 210: `flag. If set to...`

lahirumaramba: Good catch! Thanks!

PR #451, opened 2020-07-09
fix: Importing GCP dependencies via official BOMs

### LINKED ISSUE #450: Leverage libraries-bom to ensure compatible transitive dependencies between google-cloud-storage & google-cloud-firestore
Given the inherent complexities of dependency management in `com.google.cloud` libraries, `com.google.cloud:libraries-bom` can be used to simplify verification and compatibility.

The release tooling and automation implemented in `com.google.cloud:libaries-bom` ensures all dependencies defined in it are binary compatible across the full dependency graph for all of the included projects (including grpc, guava, google-cloud-* and more. See [The Google Cloud Platform Libraries BOM](https://github.com/GoogleCloudPlatform/cloud-opensource-java/wiki/The-Google-Cloud-Platform-Libraries-BOM) for links to all libraries). 

While updating to use libraries-bom wouldn't fix the issue present in #447 directly, it would provide information to users that the BOM is used for version resolution, and that their projects should do the same when using other GCP libraries in their projects along with `firebase-admin`.

hiranya911:
* Importing guava, google-cloud-* and google-http-client via `libraries-bom`.
* Import google-api-client via `google-api-client-bom`

Resolves #450 

RELEASE NOTE: Admin SDK now imports all Google Cloud Platform dependencies via `com.google.cloud:libraries-bom`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -400,7 +400,7 @@
             <dependency>
                 <groupId>com.google.api-client</groupId>
                 <artifactId>google-api-client-bom</artifactId>
-                <version>1.30.9</version>
+                <version>1.30.10</version>
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -388,49 +388,60 @@
         </plugins>
     </reporting>
 
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>com.google.cloud</groupId>
+                <artifactId>libraries-bom</artifactId>
+                <version>8.0.0</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+            <dependency>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -388,49 +388,60 @@
...
+                <groupId>com.google.cloud</groupId>
+                <artifactId>libraries-bom</artifactId>
+                <version>8.0.0</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+            <dependency>
+                <groupId>com.google.api-client</groupId>
+                <artifactId>google-api-client-bom</artifactId>
+                <version>1.30.9</version>
```
schmidt-sebastian, line 403: Why not jump to 1.30.10?

hiranya911: Oh, did not notice in the Maven listings. Upgraded.

PR #456, opened 2020-07-23
chore: Merged master branch into v7

hiranya911:
I've done the bare minimal changes necessary so this branch would compile and all tests would pass. There's some room for clean up and additional tests, which I will get to in future PRs.

Notable changes/additions:
* `AuthErrorHandler` was moved to the `internal` package so the multitenancy code can use it too (via `AuthHttpClient`)
* Added `AuthErrorCode.TENANT_ID_MISMATCH` enum value.
* `AuthHttpClient` is now a wrapper for the `ErrorHandlingHttpClient`.

Conflicts were detected only in the following files:
```
	both modified:   src/main/java/com/google/firebase/auth/FirebaseAuth.java
	both modified:   src/main/java/com/google/firebase/auth/FirebaseTokenUtils.java
	both modified:   src/main/java/com/google/firebase/auth/FirebaseTokenVerifierImpl.java
	both modified:   src/main/java/com/google/firebase/auth/FirebaseUserManager.java
	both modified:   src/main/java/com/google/firebase/auth/internal/FirebaseTokenFactory.java
	both modified:   src/main/java/com/google/firebase/messaging/FirebaseMessagingClientImpl.java
	both modified:   src/test/java/com/google/firebase/auth/FirebaseAuthIT.java
	both modified:   src/test/java/com/google/firebase/auth/FirebaseAuthTest.java
	both modified:   src/test/java/com/google/firebase/auth/FirebaseTokenVerifierImplTest.java
	both modified:   src/test/java/com/google/firebase/auth/FirebaseUserManagerTest.java
	both modified:   src/test/java/com/google/firebase/messaging/MulticastMessageTest.java
```

Review effort should mostly focus on the above classes.

PR #458, opened 2020-07-24
feat: Added new error codes for IdP management and multitenancy

hiranya911:
* Added `CONFIGURATION_NOT_FOUND` error code. Also note that `TENANT_NOT_FOUND` error code was added while updating the v7 branch in #456.
* Made it possible to test `TenantManager` and `FirebaseTenantClient` with better mock responses.
* Updated error handling test cases to check for new error codes. I also updated some of the existing error handling tests to mock with more realistic errors (e.g. create operations do not usually fail with a NOT_FOUND error).

### DIFF
--- a/src/test/java/com/google/firebase/auth/multitenancy/FirebaseTenantClientTest.java
+++ b/src/test/java/com/google/firebase/auth/multitenancy/FirebaseTenantClientTest.java
@@ -18,6 +18,7 @@ package com.google.firebase.auth.multitenancy;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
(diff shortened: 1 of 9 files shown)


src/test/java/com/google/firebase/auth/multitenancy/FirebaseTenantClientTest.java
```diff
@@ -341,17 +373,16 @@ private static TestResponseInterceptor initializeAppForTenantManagementWithStatu
...
   }
 
-  private static void initializeAppWithResponses(String... responses) {
-    List<MockLowLevelHttpResponse> mocks = new ArrayList<>();
-    for (String response : responses) {
-      mocks.add(new MockLowLevelHttpResponse().setContent(response));
-    }
-    MockHttpTransport transport = new MultiRequestMockHttpTransport(mocks);
-    FirebaseApp.initializeApp(new FirebaseOptions.Builder()
+  private static TenantManager getRetryDisabledTenantManager(MockLowLevelHttpResponse response) {
```
micahstairs, line 376: Maybe `createRetryDisabledTenantManager` would be a better name since you are creating a new instance?

hiranya911: Done

PR #469, opened 2020-08-08
Added toString() in TopicManagementResponse for logging

shambhand:


shambhand:
@googlebot I signed it!

src/main/java/com/google/firebase/messaging/TopicManagementResponse.java
```diff
@@ -122,5 +122,10 @@ public int getIndex() {
     public String getReason() {
       return reason;
     }
+
+    @Override
+    public String toString() {
+      return "Error{" + "index=" + index + ", reason='" + reason + '\'' + '}';
```
hiranya911, line 128: Use `MoreObjects.toStringHelper()` here. See `FirebaseApp.toString()` for an example.

hiranya911:
Thanks @shambhand. Can we also add a unit test for the new functionality?

shambhand:
Thanks @hiranya911 for reviewing it. 
Let me know if the `InstanceIdClientImplTest.java` is an inappropriate place for this unit test case.
`

Issue #446, opened 2020-06-19
FR: Support For Remote Config in JAVA admin SDK


ubarua123:
I see the Java admin SDK doesn't support Remote Config while the node.js does. Is there any plan to support it in the future? or is it final?

lahirumaramba:
Thank you for reaching out! Great news! We have already started working out the design for the Remote Config API for Admin Java SDK. Things are still pretty early stage, and this issue will track our work once we start implementation.

lahirumaramba:
Remote Config API is now available in Admin Java v.7.1.0. Thank you!

PR #479, opened 2020-09-25
Add parameters to Remote Config template

lahirumaramba:
- Introduce `Response Types` and `Public Types`
- Implement `toResponseType()` and `toPublicType()`
- Add `parameters` to `RemoteConfigTemplate`
- Add unit tests

Related Issue: #446


src/main/java/com/google/firebase/remoteconfig/ExplicitParameterValue.java
```diff
@@ -0,0 +1,41 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.firebase.remoteconfig;
+
+import com.google.firebase.remoteconfig.internal.TemplateResponse.ParameterValueResponse;
+
+public final class ExplicitParameterValue extends RemoteConfigParameterValue {
```
hiranya911, line 21: Add javadocs to all public APIs.

src/main/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigClientImpl.java
```diff
@@ -94,8 +95,9 @@ public RemoteConfigTemplate getTemplate() throws FirebaseRemoteConfigException {
     HttpRequestInfo request = HttpRequestInfo.buildGetRequest(remoteConfigUrl)
             .addAllHeaders(COMMON_HEADERS);
     IncomingHttpResponse response = httpClient.send(request);
-    RemoteConfigTemplate parsed = httpClient.parse(response, RemoteConfigTemplate.class);
-    return parsed.setETag(getETag(response));
+    TemplateResponse templateResponse = httpClient.parse(response, TemplateResponse.class);
+    RemoteConfigTemplate template = templateResponse.toPublicType();
```
hiranya911, line 99: `toRemoteConfigTemplate()`

In our chats I used `toPublicType()` and `toResponseType()` as placeholders. Use the typical Java conventions to name these methodds.

src/main/java/com/google/firebase/remoteconfig/InAppDefaultValue.java
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+
+package com.google.firebase.remoteconfig;
+
+import com.google.firebase.remoteconfig.internal.TemplateResponse.ParameterValueResponse;
+
+public final class InAppDefaultValue extends RemoteConfigParameterValue {
+
+  private final boolean inAppDefaultValue;
```
hiranya911, line 23: No need for this. It's always true.

src/main/java/com/google/firebase/remoteconfig/RemoteConfigParameter.java
```diff
@@ -0,0 +1,72 @@
...
+
+public final class RemoteConfigParameter {
+
+  private RemoteConfigParameterValue defaultValue;
+  private String description;
+  private Map<String, RemoteConfigParameterValue> conditionalValues;
+
+  public RemoteConfigParameter() {
+    conditionalValues = new HashMap<>();
+    defaultValue = ExplicitParameterValue.of("");
```
hiranya911, line 33: Why this default value?

src/main/java/com/google/firebase/remoteconfig/RemoteConfigParameter.java
```diff
@@ -0,0 +1,72 @@
...
+    return this;
+  }
+
+  public RemoteConfigParameter setConditionalValues(
+          Map<String, RemoteConfigParameterValue> conditionalValues) {
+    this.conditionalValues = conditionalValues;
+    return this;
+  }
+
+  ParameterResponse toResponseType() {
```
hiranya911, line 64: `toParameterResponse()`

src/main/java/com/google/firebase/remoteconfig/internal/TemplateResponse.java
```diff
@@ -0,0 +1,114 @@
...
+
+  public TemplateResponse() {
+    parameters = Collections.emptyMap();
+  }
+
+  public TemplateResponse(Map<String, ParameterResponse> parameters) {
+    this.parameters = parameters;
+  }
+
+  public RemoteConfigTemplate toPublicType() {
```
hiranya911, line 43: toRemoteConfigTemplate

src/main/java/com/google/firebase/remoteconfig/internal/TemplateResponse.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    public ParameterResponse(ParameterValueResponse defaultValue, String description,
+                             Map<String, ParameterValueResponse> conditionalValues) {
+      this.defaultValue = defaultValue;
+      this.description = description;
+      this.conditionalValues = conditionalValues;
+    }
+
+    public RemoteConfigParameter toPublicType() {
```
hiranya911, line 73: toRemoteConfigParameter

src/main/java/com/google/firebase/remoteconfig/internal/TemplateResponse.java
```diff
@@ -0,0 +1,114 @@
...
+
+    public static ParameterValueResponse ofValue(String value) {
+      return new ParameterValueResponse(value, null);
+    }
+
+    public static ParameterValueResponse ofInAppDefaultValue() {
+      return new ParameterValueResponse(null, true);
+    }
+
+    public RemoteConfigParameterValue toPublicType() {
```
hiranya911, line 107: toRemoteConfigParameterValue

src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigClientImplTest.java
```diff
@@ -63,7 +63,33 @@
           404, ErrorCode.NOT_FOUND,
           500, ErrorCode.INTERNAL);
 
-  private static final String MOCK_TEMPLATE_RESPONSE = "{\"conditions\": [], \"parameters\": {}}";
+  private static final String MOCK_TEMPLATE_RESPONSE = "{"
```
hiranya911, line 66: Probably best to load this from a resource file.

src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigClientImplTest.java
```diff
@@ -86,6 +112,26 @@ public void testGetTemplate() throws Exception {
...
+    assertTrue(parameters.containsKey("header_text"));
+    Map<String, RemoteConfigParameterValue> conditionalValues = parameters
+            .get("welcome_message_text").getConditionalValues();
+    assertEquals(1, conditionalValues.size());
+    assertTrue(conditionalValues.containsKey("ios_en"));
+    ExplicitParameterValue value = (ExplicitParameterValue) conditionalValues.get("ios_en");
+    assertEquals("welcome to app en", value.getValue());
+    checkGetRequestHeader(interceptor.getLastRequest());
+
+    // Check empty template
```
hiranya911, line 127: This should probably be a separate test case.

src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigClientImplTest.java
```diff
@@ -86,6 +112,26 @@ public void testGetTemplate() throws Exception {
     RemoteConfigTemplate template = client.getTemplate();
 
     assertEquals(TEST_ETAG, template.getETag());
+    Map<String, RemoteConfigParameter> parameters = template.getParameters();
```
hiranya911, line 115: Also assert for defaultValues and descriptions

hiranya911 (changes requested):
Looks pretty good. I suggested a few renames and test additions.

src/main/java/com/google/firebase/remoteconfig/RemoteConfigParameter.java
lahirumaramba: If no default value is set (if `defaultValue` is `null`), the REST API sets the default value to an empty string. We can allow `null` here and add a `null` check in the conversion or just set it to an empty value in the `ctor`.

src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigClientImplTest.java
lahirumaramba: Good call!

src/main/java/com/google/firebase/remoteconfig/RemoteConfigParameter.java
```diff
@@ -22,51 +22,99 @@
...
   private String description;
   private Map<String, RemoteConfigParameterValue> conditionalValues;
 
+  /**
+   * Creates a new {@link RemoteConfigParameter}.
+   */
   public RemoteConfigParameter() {
     conditionalValues = new HashMap<>();
-    defaultValue = ExplicitParameterValue.of("");
+    defaultValue = RemoteConfigParameterValue.of("");
```
hiranya911, line 41: I feel like both these should remain null until explicitly set. It's ok to let the backend choose default values for these as necessary.

src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigClientImplTest.java
```diff
@@ -116,19 +91,26 @@ public void testGetTemplate() throws Exception {
     assertEquals(2, parameters.size());
     assertTrue(parameters.containsKey("welcome_message_text"));
     assertTrue(parameters.containsKey("header_text"));
-    Map<String, RemoteConfigParameterValue> conditionalValues = parameters
-            .get("welcome_message_text").getConditionalValues();
+    RemoteConfigParameter parameter = parameters.get("welcome_message_text");
```
hiranya911, line 94: Also assert on `header_text` which has an in app default value param.

hiranya911 (changes requested):
Looks great. Just a couple of suggestions.

src/main/java/com/google/firebase/remoteconfig/RemoteConfigParameter.java
```diff
@@ -0,0 +1,120 @@
...
+   *
+   * @return The {@link String} description of the parameter.
+   */
+  public String getDescription() {
+    return description;
+  }
+
+  /**
+   * Gets the conditional values of the parameter.
+   * The condition name of the highest priority (the one listed first in the
```
egilmorez, line 64: This is interesting . . . you mean this gets all the possible values, including the one actually assigned -- for a particular client instance? -- in the top position in the map?

src/main/java/com/google/firebase/remoteconfig/RemoteConfigParameter.java
```diff
@@ -0,0 +1,120 @@
...
+   *
+   * @return A map of conditional values.
+   */
+  public Map<String, RemoteConfigParameterValue> getConditionalValues() {
+    return conditionalValues;
+  }
+
+  /**
+   * Sets the default value of the parameter.
+   * This is the value to set the parameter to, when none of the named conditions
```
egilmorez, line 75: Does the developer actually set the value?  Or does our backend logic?

src/main/java/com/google/firebase/remoteconfig/RemoteConfigParameter.java
```diff
@@ -0,0 +1,120 @@
...
+   * @param description The description of the parameter.
+   * @return This {@link RemoteConfigParameter}.
+   */
+  public RemoteConfigParameter setDescription(String description) {
+    this.description = description;
+    return this;
+  }
+
+  /**
+   * Sets the conditional values of the parameter.
```
egilmorez, line 99: Again this is maybe just my own confusion, but -- I can see how a developer might want to automate the setting of a list of conditional values.  But isn't the actual value sort of set at runtime, after evaluating the conditions?

src/main/java/com/google/firebase/remoteconfig/RemoteConfigParameterValue.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    @Override
+    ParameterValueResponse toParameterValueResponse() {
+      return ParameterValueResponse.ofValue(this.value);
+    }
+  }
+
+  /**
+   * Represents an in-app-default parameter value.
```
egilmorez, line 75: Suggest removing the hyphen between app and default.

src/main/java/com/google/firebase/remoteconfig/internal/TemplateResponse.java
```diff
@@ -0,0 +1,123 @@
...
+                .put(entry.getKey(), entry.getValue().toRemoteConfigParameterValue());
+      }
+      return new RemoteConfigParameter()
+              .setDefaultValue(defaultValue.toRemoteConfigParameterValue())
+              .setDescription(description).setConditionalValues(conditionalPublicValues);
+    }
+  }
+
+  /**
+   * The DTO for parsing Remote Config parameter value responses from the Remote Config service.
```
egilmorez, line 90: I wonder if we should spell out "Data Transfer Object" -- that's what this is, right?  :)

I searched and couldn't immediately find a single other instance of "DTO" in docs/reference/

egilmorez (commented):
A couple suggested edits and a few questions, thanks!

src/main/java/com/google/firebase/remoteconfig/RemoteConfigParameter.java
lahirumaramba: This gets all the conditional values assigned for that specific parameter. You can set a default value in a parameter to be used when all the conditional values are evaluated to false. Otherwise, the parameter will use the conditional value with the highest priority. The priority is determined by the order of the conditions are stored in the top-level `conditions` array of the `RemoteConfigTemplate`. The top-level `conditions` array is not part of this PR and will be added in a future PR.
https://firebase.google.com/docs/reference/remote-config/rest/v1/RemoteConfigParameter

src/main/java/com/google/firebase/remoteconfig/RemoteConfigParameter.java
lahirumaramba: When getting an existing template the backend will set the values for existing Parameters. If the developer wants to create and add new parameters then they can set a default value to set the parameter to, when none of the named conditions evaluate to true.

src/main/java/com/google/firebase/remoteconfig/RemoteConfigParameter.java
lahirumaramba: When creating new parameters the developer needs to add conditional values manually. The condition names in `conditionalValues` map must match the names in the top-level `conditions` array in `RemoteConfigTemplate`. Otherwise, the API will throw a `condition does not exist` error.

src/main/java/com/google/firebase/remoteconfig/internal/TemplateResponse.java
lahirumaramba: I think the classes in `internal` packages do not show up in the docs. I agree though `Data Transfer Object` makes more sense. I will update the comment here. Thanks!!

src/main/java/com/google/firebase/remoteconfig/RemoteConfigParameter.java
egilmorez: OK, thanks for the explanation.  Parameter evaluation is pretty complex . . . when everything is done, I might try to start a review of our "parameters and conditions" concepts page, just to make sure it's all still accurate.

egilmorez (approved):
Looks good Lahiru, thanks!

src/main/java/com/google/firebase/remoteconfig/RemoteConfigParameter.java
lahirumaramba: I see your point. Updated the code and added `null` checks in conversions.
I have a few concerns with setting these to `null` though. Isn't it recommended not to return `null` collections in Java? If we don't set to `conditionalValues` to an empty map at init, the following will throw an exception on the client side.
```
RemoteConfigParameter p = new RemoteConfigParameter().getConditionalValues().put("ios_en", ParameterValue.of("text")```

This can be prevented with a simple `null` check on the client side, though.

File: src/main/java/com/google/firebase/remoteconfig/internal/TemplateResponse.java [hunk 19]
```diff
@@ -80,14 +90,17 @@ public RemoteConfigParameter toRemoteConfigParameter() {
         conditionalPublicValues
                 .put(entry.getKey(), entry.getValue().toRemoteConfigParameterValue());
       }
+      RemoteConfigParameterValue remoteConfigParameterValue =
+              (defaultValue == null) ? null : defaultValue.toRemoteConfigParameterValue();
       return new RemoteConfigParameter()
-              .setDefaultValue(defaultValue.toRemoteConfigParameterValue())
+              .setDefaultValue(remoteConfigParameterValue)
               .setDescription(description).setConditionalValues(conditionalPublicValues);
```
hiranya911, line 97: Nit: One setter per line

src/main/java/com/google/firebase/remoteconfig/internal/TemplateResponse.java
```diff
@@ -67,8 +75,10 @@ public ParameterResponse() {
       conditionalValues = Collections.emptyMap();
     }
 
-    public ParameterResponse(ParameterValueResponse defaultValue, String description,
-                             Map<String, ParameterValueResponse> conditionalValues) {
+    public ParameterResponse(@Nullable ParameterValueResponse defaultValue,
+                             @Nullable String description,
+                             @NonNull Map<String, ParameterValueResponse> conditionalValues) {
+      checkNotNull(conditionalValues, "conditional values must not be null.");
```
hiranya911, line 81: Probably redundant at this level.

hiranya911 (approved):
The semantics aren't quite sitting well with me. But let's merge this and address any concerns later.

PR #490, opened 2020-10-28
Add Remote Config Parameter Group type

lahirumaramba:
- Add Remote Config Parameter Group type
- Implement `equals()` in all public RC types
- Add unit tests for public RC types
- Refactor existing unit tests to use `equals()` when comparing templates

Related to: #446

### DIFF
--- a/src/main/java/com/google/firebase/remoteconfig/ParameterGroup.java
+++ b/src/main/java/com/google/firebase/remoteconfig/ParameterGroup.java
@@ -1,0 +1,138 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.firebase.remoteconfig;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.firebase.internal.NonNull;
+import com.google.firebase.internal.Nullable;
+import com.google.firebase.remoteconfig.internal.TemplateResponse;
+import com.google.firebase.remoteconfig.internal.TemplateResponse.ParameterGroupResponse;
+import com.google.firebase.remoteconfig.internal.TemplateResponse.ParameterResponse;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * Represents a Remote Config parameter group that can be included in a {@link Template}.
+ * Grouping parameters is only for management purposes and does not affect client-side
+ * fetching of parameter values.
+ */
+public final class ParameterGroup {
+
+  private String description;
+  private Map<String, Parameter> parameters;
+
+  /**
+   * Creates a new {@link ParameterGroup}.
+   */
+  public ParameterGroup() {
+    parameters = new HashMap<>();
+  }
+
+  ParameterGroup(@NonNull ParameterGroupResponse parameterGroupResponse) {
+    checkNotNull(parameterGroupResponse);
+    this.parameters = new HashMap<>();
+    if (parameterGroupResponse.getParameters() != null) {
+      for (Map.Entry<String, TemplateResponse.ParameterResponse> entry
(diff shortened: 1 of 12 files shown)


src/main/java/com/google/firebase/remoteconfig/ParameterGroup.java
```diff
@@ -0,0 +1,138 @@
...
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    ParameterGroup that = (ParameterGroup) o;
+    return Objects.equals(description, that.description)
+            && Objects.equals(parameters, that.parameters);
```
hiranya911, line 131: Does this work on collections like maps?

src/main/java/com/google/firebase/remoteconfig/Template.java
```diff
@@ -132,8 +166,28 @@ TemplateResponse toTemplateResponse() {
...
+    }
     return new TemplateResponse()
             .setParameters(parameterResponses)
-            .setConditions(conditionResponses);
+            .setConditions(conditionResponses)
+            .setParameterGroups(parameterGroupResponse);
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
hiranya911, line 180: Implement hashCode too.

src/test/java/com/google/firebase/remoteconfig/ConditionTest.java
```diff
@@ -18,11 +18,61 @@
...
 
+  @Test
+  public void testConstructor() {
+    Condition c1 = new Condition("ios_en_1", "expression1");
+
+    assertEquals("ios_en_1", c1.getName());
+    assertEquals("expression1", c1.getExpression());
+    assertNull(c1.getTagColor());
+
+    Condition c2 = new Condition("ios_en_2", "expression2", TagColor.BLUE);
```
hiranya911, line 35: Move into a separate test case: testConstructorWithColor

src/test/java/com/google/firebase/remoteconfig/ParameterTest.java
```diff
@@ -0,0 +1,87 @@
...
+    final Parameter p = new Parameter();
+    assertNotNull(p.getConditionalValues());
+    assertEquals(0, p.getConditionalValues().size());
+    assertNull(p.getDefaultValue());
+    assertNull(p.getDescription());
+  }
+
+  @Test(expected = NullPointerException.class)
+  public void testConstructorWithNullParameterResponse() {
+    Parameter p = new Parameter(null);
```
hiranya911, line 43: Just `new Parameter(null)` without the assignment

src/test/java/com/google/firebase/remoteconfig/TemplateTest.java
```diff
@@ -0,0 +1,136 @@
...
+public class TemplateTest {
+
+  @Test
+  public void testConstructor() {
+    Template t = new Template();
+
+    assertNotNull(t.getParameters());
+    assertNotNull(t.getConditions());
+    assertNotNull(t.getParameterGroups());
+    assertEquals(0, t.getParameters().size());
```
hiranya911, line 42: `assertTrue(t.getParameters().isEmpty())`

hiranya911 (changes requested):
Looks great. I only had a few nits.

src/main/java/com/google/firebase/remoteconfig/ParameterGroup.java
lahirumaramba: Yes, it seems to be working fine on maps. `Objects.equals(o1, o2)` will end up calling `o1.equals(o2)` and that should work fine since we have implemented `equals` in the custom types `Parameter` etc.
https://docs.oracle.com/javase/7/docs/api/java/util/Map.html#equals(java.lang.Object)

I have added test cases for equality in `ParameterGroupTest.java`. I can change that test to use two separate collections to make it more explicit if that helps...

src/test/java/com/google/firebase/remoteconfig/ParameterTest.java
lahirumaramba: Good catch! Thanks!
Updated.

src/test/java/com/google/firebase/remoteconfig/TemplateTest.java
lahirumaramba: Good one! Updated. Thanks!

hiranya911 (approved):
LGTM

Nit: Unit tests use a lot of short variable names like `c` and `p`. That's not very Java idiomatic.

lahirumaramba:
> LGTM
> 
> Nit: Unit tests use a lot of short variable names like `c` and `p`. That's not very Java idiomatic.

Thanks for pointing this out!! Updated with proper variable names!

PR #491, opened 2020-11-03
Add Version information to RC template

lahirumaramba:
- Add Version information to the Template
- Create new Version type
- Create new User type
- ~Note: Since Version and User types are output only, I did not implement `equal()` and `hashcode()` operations. Instead, unit tests are checking for each property in `Version` and `User` individually. I also removed the equality operations from `Template`, as well.~
- Based on PR review feedback, implement equality in all public types (including `Version` and `User`) to keep consistency.

Related to: #446

### DIFF
--- a/src/main/java/com/google/firebase/remoteconfig/User.java
+++ b/src/main/java/com/google/firebase/remoteconfig/User.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.firebase.remoteconfig;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.firebase.internal.NonNull;
+import com.google.firebase.internal.Nullable;
+import com.google.firebase.remoteconfig.internal.TemplateResponse.UserResponse;
+
+/**
+ * Represents a Remote Config user. Output only.
+ */
+public final class User {
+
+  private String email;
+  private String name;
+  private String imageUrl;
+
+  User(@NonNull UserResponse userResponse) {
+    checkNotNull(userResponse);
+    this.email = userResponse.getEmail();
+    this.name = userResponse.getName();
+    this.imageUrl = userResponse.getImageUrl();
+  }
+
+  /**
+   * Gets the email of the user.
+   *
+   * @return The email of the user or null.
+   */
+  @Nullable
+  public String getEmail() {
+    return email;
+  }
+
+  /**
+   * Gets the name of the user.
+   *
+   * @return The name of the user or null.
+   */
+  @Nullable
+  public String getName() {
+    return name;
+  }
+
+  /**
+   * Gets the image URL of the user.
+   *
+   * @return The image URL of the user or null.
+   */
+  @Nullable
+  public String getImageUrl() {
+    return imageUrl;
+  }
+}
--- a/src/main/java/com/google/firebase/remoteconfig/User.java
+++ b/src/main/java/com/google/firebase/remoteconfig/User.java
@@ -22,14 +22,16 @@ import com.google.firebase.internal.NonNull;
 import com.google.firebase.internal.Nullable;
 import com.google.firebase.remoteconfig.internal.TemplateResponse.UserResponse;
 
+import java.util.Objects;
+
 /**
  * Represents a Remote Config user. Output only.
  */
-public final class User {
+public class User {
 
-  private String email;
-  private String name;
-  private String imageUrl;
+  private final String email;
+  private final String name;
+  private final String imageUrl;
 
   User(@NonNull UserResponse userResponse) {
     checkNotNull(userResponse);
@@ -66,5 +68,24 @@ public final class User {
   @Nullable
   public String getImageUrl() {
     return imageUrl;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    User user = (User) o;
+    return Objects.equals(email, user.email)
+            && Objects.equals(name, user.name)
+            && Objects.equals(imageUrl, user.imageUrl);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(email, name, imageUrl);
   }
 }
--- a/src/main/java/com/google/firebase/remoteconfig/Version.java
+++ b/src/main/java/com/google/firebase/remoteconfig/Version.java
@@ -40,9 +40,6 @@ import java.util.regex.Pattern;
  */
 public final class Version {
 
-  private static final Pattern ZULU_TIME_NO_NANOSECONDS_PATTERN = Pattern
-          .compile("^(\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2})");
-
   private final String versionNumber;
   private final long updateTime;
   private final String updateOrigin;
@@ -69,19 +66,16 @@ public final class Version {
     if (!Strings.isNullOrEmpty(versionResponse.getUpdateTime())) {
       // Update Time is a timestamp in RFC3339 UTC "Zulu" format, accurate to nanoseconds.
       // example: "2014-10-02T15:01:23.045123456Z"
-      // SimpleDateFormat cannot handle nanoseconds, therefore we drop nanoseconds from the string.
-      Matcher errorMatcher = ZULU_TIME_NO_NANOSECONDS_PATTERN
-              .matcher(versionResponse.getUpdateTime());
-      String updateTimeWithoutNanoseconds = "";
-      if (errorMatcher.find()) {
-        updateTimeWithoutNanoseconds = errorMatcher.group(1);
-      }
+      // SimpleDateFormat cannot handle nanoseconds, therefore we strip nanoseconds from the string.
+      String updateTime = versionResponse.getUpdateTime();
+      int indexOfPeriod = !updateTime.contains(".") ? 0 : updateTime.indexOf(".");
+      String updateTimeWithoutNanoseconds = updateTime.substring(0, indexOfPeriod);
       SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss");
       dateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
       try {
         this.updateTime = dateFormat.parse(updateTimeWithoutNanoseconds).getTime();
       } catch (ParseException e) {
-        throw new IllegalStateException();
+        throw new IllegalStateException("Unable to parse update time.", e);
       }
     } else {
       this.updateTime = 0L;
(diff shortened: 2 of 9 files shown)


src/main/java/com/google/firebase/remoteconfig/Version.java
```diff
@@ -0,0 +1,181 @@
...
+  Version(@NonNull VersionResponse versionResponse) {
+    checkNotNull(versionResponse);
+    this.versionNumber = versionResponse.getVersionNumber();
+    if (!Strings.isNullOrEmpty(versionResponse.getUpdateTime())) {
+      SimpleDateFormat dateFormat = new SimpleDateFormat("EEE, dd MMM yyyy HH:mm:ss z");
+      dateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
+      try {
+        this.updateTime = dateFormat.parse(versionResponse.getUpdateTime()).getTime();
+      } catch (ParseException e) {
+        this.updateTime = 0;
```
lahirumaramba, line 68: I did not want to throw here as the backend ideally should not provide invalid date time strings... Open to suggestions :)

src/main/java/com/google/firebase/remoteconfig/User.java
```diff
@@ -0,0 +1,70 @@
...
+import com.google.firebase.internal.NonNull;
+import com.google.firebase.internal.Nullable;
+import com.google.firebase.remoteconfig.internal.TemplateResponse.UserResponse;
+
+/**
+ * Represents a Remote Config user. Output only.
+ */
+public final class User {
+
+  private String email;
```
hiranya911, line 30: final

src/main/java/com/google/firebase/remoteconfig/Version.java
```diff
@@ -0,0 +1,181 @@
...
+/**
+ * Represents a Remote Config template version.
+ * Output only, except for the version description. Contains metadata about a particular
+ * version of the Remote Config template. All fields are set at the time the specified Remote
+ * Config template is published. A version's description field may be specified when
+ * publishing a template.
+ */
+public final class Version {
+
+  private String versionNumber;
```
hiranya911, line 40: Everything except description can be made final.

src/main/java/com/google/firebase/remoteconfig/Version.java
```diff
@@ -0,0 +1,181 @@
...
+  private boolean legacy;
+
+  /**
+   * Creates a new {@link Version} with a description.
+   */
+  public static Version withDescription(String description) {
+    return new Version().setDescription(description);
+  }
+
+  Version() {
```
hiranya911, line 56: private?

src/main/java/com/google/firebase/remoteconfig/Version.java
```diff
@@ -0,0 +1,181 @@
...
+  private String updateType;
+  private User updateUser;
+  private String description;
+  private String rollbackSource;
+  private boolean legacy;
+
+  /**
+   * Creates a new {@link Version} with a description.
+   */
+  public static Version withDescription(String description) {
```
hiranya911, line 52: List factory methods after the constructors.

src/main/java/com/google/firebase/remoteconfig/Version.java
hiranya911: I think it's reasonable to throw IllegalStateException or something and fail fast.

src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigClientImplTest.java
```diff
@@ -105,21 +108,32 @@ public void testGetTemplate() throws Exception {
...
+    dateFormat.setTimeZone(TimeZone.getTimeZone("GMT"));
+    final String updateTime = dateFormat.format(new Date(receivedVersion.getUpdateTime()));
+
+    assertEquals("17", receivedVersion.getVersionNumber());
+    assertEquals("ADMIN_SDK_NODE", receivedVersion.getUpdateOrigin());
+    assertEquals("INCREMENTAL_UPDATE", receivedVersion.getUpdateType());
+    assertEquals("firebase-user@account.com", receivedVersion.getUpdateUser().getEmail());
+    assertEquals("dev-admin", receivedVersion.getUpdateUser().getName());
+    assertEquals("http://image.jpg", receivedVersion.getUpdateUser().getImageUrl());
+    assertEquals("Wed, 30 Sep 2020 17:56:07 GMT", updateTime);
```
hiranya911, line 135: May be just compare the long values.

hiranya911 (changes requested):
It looks pretty solid. I only had some minor nits. But I do think we should support `equals` on the new classes. A developer might want to do something like the following:

```java
if (!template1.getVersion().equals(template2.getVersion()) {
  // do something
}
```

It's also consistent with other classes in this package. WDYT?

src/main/java/com/google/firebase/remoteconfig/Version.java
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  Version(@NonNull VersionResponse versionResponse) {
+    checkNotNull(versionResponse);
+    this.versionNumber = versionResponse.getVersionNumber();
+
+    if (!Strings.isNullOrEmpty(versionResponse.getUpdateTime())) {
+      // Update Time is a timestamp in RFC3339 UTC "Zulu" format, accurate to nanoseconds.
+      // example: "2014-10-02T15:01:23.045123456Z"
+      // SimpleDateFormat cannot handle nanoseconds, therefore we drop nanoseconds from the string.
```
lahirumaramba, line 72: I noticed that the update time is represented in FC3339 UTC "Zulu" format, accurate to nanoseconds. Added new code here to convert the timestamp to milliseconds properly. `DateTime` does not handle nanoseconds so I had to drop them from the timestamp string before parsing. Otherwise `SimpleDateFormat` consider nanoseconds as milliseconds and add that it back to the time making the parsed datetime invalid.

Apparently, there are better ways to handle nanoseconds in Java8 (`LocalDateTime` and `Instant`) :)

lahirumaramba:
Implemented `equal()` in all public types (including `Version` and `User`) to keep consistency

src/main/java/com/google/firebase/remoteconfig/Version.java
```diff
@@ -37,37 +40,53 @@
...
+        updateTimeWithoutNanoseconds = errorMatcher.group(1);
+      }
+      SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss");
       dateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
       try {
-        this.updateTime = dateFormat.parse(versionResponse.getUpdateTime()).getTime();
+        this.updateTime = dateFormat.parse(updateTimeWithoutNanoseconds).getTime();
       } catch (ParseException e) {
-        this.updateTime = 0;
+        throw new IllegalStateException();
```
hiranya911, line 84: Include a detailed error message.

src/main/java/com/google/firebase/remoteconfig/Version.java
```diff
@@ -37,37 +40,53 @@
...
   Version(@NonNull VersionResponse versionResponse) {
     checkNotNull(versionResponse);
     this.versionNumber = versionResponse.getVersionNumber();
+
     if (!Strings.isNullOrEmpty(versionResponse.getUpdateTime())) {
-      SimpleDateFormat dateFormat = new SimpleDateFormat("EEE, dd MMM yyyy HH:mm:ss z");
+      // Update Time is a timestamp in RFC3339 UTC "Zulu" format, accurate to nanoseconds.
+      // example: "2014-10-02T15:01:23.045123456Z"
+      // SimpleDateFormat cannot handle nanoseconds, therefore we drop nanoseconds from the string.
+      Matcher errorMatcher = ZULU_TIME_NO_NANOSECONDS_PATTERN
```
hiranya911, line 73: Instead of a complex regex, why not just strip away everything (and including) the period? That seems easier and more readable.

src/test/java/com/google/firebase/remoteconfig/TemplateTest.java
```diff
@@ -59,4 +67,77 @@ public void testSetNullParameterGroups() {
     Template template = new Template();
     template.setParameterGroups(null);
   }
+
+  @Test
+  public void testEquality() {
```
hiranya911: I'd advise splitting this into multiple smaller test cases. You can define some class-level constants for values that should be shared across test cases.

src/test/resources/getRemoteConfig.json
```diff
@@ -50,7 +50,7 @@
       "name": "dev-admin",
       "imageUrl": "http://image.jpg"
     },
-    "updateTime": "Wed, 30 Sep 2020 17:56:07 GMT",
+    "updateTime": "2020-11-03T20:24:15.203Z",
```
hiranya911, line 53: This is not quite the format described in the comment in the source (it had 9 digit nano second value).

hiranya911 (changes requested):
Thanks. This looks great. Just a few nits.

src/test/resources/getRemoteConfig.json
lahirumaramba: Good catch!

src/main/java/com/google/firebase/remoteconfig/Version.java
lahirumaramba: Makes sense! Updated the code to use `substring()`

src/test/java/com/google/firebase/remoteconfig/TemplateTest.java
lahirumaramba: I use all the template instances created here in a series of `assertNotEquals()` at the end of the test... would it be fine to move all those instances to class level?

src/main/java/com/google/firebase/remoteconfig/Version.java
```diff
@@ -69,19 +66,16 @@ private Version() {
...
-      Matcher errorMatcher = ZULU_TIME_NO_NANOSECONDS_PATTERN
-              .matcher(versionResponse.getUpdateTime());
-      String updateTimeWithoutNanoseconds = "";
-      if (errorMatcher.find()) {
-        updateTimeWithoutNanoseconds = errorMatcher.group(1);
-      }
+      // SimpleDateFormat cannot handle nanoseconds, therefore we strip nanoseconds from the string.
+      String updateTime = versionResponse.getUpdateTime();
+      int indexOfPeriod = !updateTime.contains(".") ? 0 : updateTime.indexOf(".");
+      String updateTimeWithoutNanoseconds = updateTime.substring(0, indexOfPeriod);
```
hiranya911, line 72: indexOfPeriod can be possibly 0, in which case this will do `substring(0, 0)`.  I think following makes more sense to me:

```java
String updateTime = versionResponse.getUpdateTime();
int indexOfPeriod = updateTime.indexOf('.');
if (indexOfPeriod != -1) {
  updateTime = updateTime.substring(0, indexOfPeriod);
}
```

src/test/java/com/google/firebase/remoteconfig/TemplateTest.java
```diff
@@ -132,5 +132,12 @@ public void testEquality() {
     assertNotEquals(templateThree, templateFive);
     assertNotEquals(templateThree, templateSeven);
     assertNotEquals(templateFive, templateSeven);
+
+    final Template templateNine = new Template()
```
hiranya911, line 136: It's ok to define some of these as class level constants. For example templateNine can be a class-level constant, and templateTen can be initialized in the corresponding test case as a local var. Same goes for others.

hiranya911 (approved):
LGTM with a couple of suggestions.

src/main/java/com/google/firebase/remoteconfig/Version.java
lahirumaramba: I agree! This is better. Update the code. Thanks!

src/test/java/com/google/firebase/remoteconfig/TemplateTest.java
lahirumaramba: I see what you mean now! Moved to class level constants. Thanks!

PR #497, opened 2020-11-24
Add Remote Config rollback operation

lahirumaramba:
- Add `rollback` operations
- Add unit tests

Related to #446

### DIFF
--- a/src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigClientImplTest.java
+++ b/src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigClientImplTest.java
@@ -32,6 +32,7 @@ import com.google.api.client.http.HttpRequest;
 import com.google.api.client.http.HttpResponseException;
 import com.google.api.client.http.HttpResponseInterceptor;
 import com.google.api.client.http.HttpTransport;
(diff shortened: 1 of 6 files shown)


src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigClientImplTest.java
```diff
@@ -803,6 +1008,24 @@ private void checkPutRequestHeader(HttpRequest request, String query, String ifM
     assertEquals(ifMatch, headers.getIfMatch());
   }
 
+  private void checkPostRequestHeader(HttpRequest request, String query) {
```
hiranya911, line 1011: It's not exactly a `query`, but any URL suffix. I'd suggest renaming the argument to reflect that.

hiranya911 (approved):
LGTM with a comment.

src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigClientImplTest.java
lahirumaramba: Good point! Thanks! Updated.

PR #500, opened 2020-12-07
Add toJSON and fromJSON to Remote Config Template

lahirumaramba:
- Add `toJSON()` and `fromJSON()` to `Template`
- Add unit tests

Related to: #446


src/main/java/com/google/firebase/remoteconfig/Template.java
```diff
@@ -73,6 +87,22 @@ public Template() {
...
+  /**
+   * Creates and returns a new Remote Config template from a JSON string.
+   *
+   * @param json A non-null JSON string to populate a Remote Config template.
+   * @return A new {@link Template} instance.
+   * @throws IOException If the input JSON string is not parsable.
+   */
+  public static Template fromJSON(@NonNull String json) throws IOException {
+    checkArgument(!Strings.isNullOrEmpty(json), "JSON String must not be null or empty.");
+    JsonFactory jsonFactory = Utils.getDefaultJsonFactory();
```
hiranya911, line 102: We should use the JsonFactory that comes from AppOptions

src/main/java/com/google/firebase/remoteconfig/Template.java
```diff
@@ -177,6 +207,25 @@ public Template setVersion(Version version) {
...
   }
 
+  /**
+   * Gets the JSON-serializable representation of this template.
+   *
+   * @return A JSON-serializable representation of this {@link Template} instance.
+   */
+  public String toJSON() {
+    String jsonSerialization;
+    Gson gson = new GsonBuilder()
```
hiranya911, line 217: Why not just construct a `TemplateResponse` instance here and then serialize that via `JsonFactory`? Gson shouldn't be needed.

src/main/java/com/google/firebase/remoteconfig/Template.java
```diff
@@ -224,4 +273,33 @@ public boolean equals(Object o) {
...
+    @Override
+    public JsonElement serialize(ParameterValue.InAppDefault src, Type typeOfSrc,
+                                 JsonSerializationContext context) {
+      JsonObject obj = new JsonObject();
+      obj.addProperty("useInAppDefault", true);
+      return obj;
+    }
+  }
+
+  private static class VersionAdapter implements JsonSerializer<Version> {
```
hiranya911, line 288: This wouldn't be needed if we used the existing TemplateResponse and VersionResponse types to handle serialization.

src/main/java/com/google/firebase/remoteconfig/Template.java
```diff
@@ -224,4 +273,33 @@ public boolean equals(Object o) {
   public int hashCode() {
     return Objects.hash(etag, parameters, conditions, parameterGroups, version);
   }
+
+  private static class InAppDefaultAdapter implements JsonSerializer<ParameterValue.InAppDefault> {
```
hiranya911, line 277: Can be handled via `ParameterValueResponse`.

hiranya911 (changes requested):
Haven't gone through everything. But I've added some high-level feedback around how the serialization logic can be improved.

src/main/java/com/google/firebase/remoteconfig/Template.java
lahirumaramba: Note: Discussed offline and decided to use the default json factory here for local serializations. Added a comment 
in code to clarify.

src/main/java/com/google/firebase/remoteconfig/Template.java
lahirumaramba: Good call! Updated the code to use Response types. Thanks!

src/main/java/com/google/firebase/remoteconfig/Template.java
```diff
@@ -73,6 +79,23 @@ public Template() {
...
+  }
+
+  /**
+   * Creates and returns a new Remote Config template from a JSON string.
+   *
+   * @param json A non-null JSON string to populate a Remote Config template.
+   * @return A new {@link Template} instance.
+   * @throws IOException If the input JSON string is not parsable.
+   */
+  public static Template fromJSON(@NonNull String json) throws IOException {
```
hiranya911, line 92: Should we throw a `FirebaseRemoteConfigException` here?

src/main/java/com/google/firebase/remoteconfig/Template.java
```diff
@@ -196,12 +233,13 @@ TemplateResponse toTemplateResponse() {
...
     TemplateResponse.VersionResponse versionResponse = (this.version == null) ? null
-            : this.version.toVersionResponse();
+            : this.version.toVersionResponse(includeAll);
     return new TemplateResponse()
             .setParameters(parameterResponses)
             .setConditions(conditionResponses)
             .setParameterGroups(parameterGroupResponse)
-            .setVersion(versionResponse);
+            .setVersion(versionResponse)
+            .setEtag(includeAll ? this.etag : null);
```
hiranya911, line 242: Nit: Slightly better for readability if this was turned into an if condition.

```java
if (includeAll) {
  return templateResponse.setEtag(this.etag);
}

return templateResponse;
```

src/main/java/com/google/firebase/remoteconfig/Version.java
```diff
@@ -196,8 +190,20 @@ public Version setDescription(String description) {
     return this;
   }
 
-  VersionResponse toVersionResponse() {
+  VersionResponse toVersionResponse(boolean includeAll) {
+    if (!includeAll) {
```
hiranya911, line 194: Nit:

```java
VersionResponse response = new VersionResponse().setDescription(this.description);
if (includeAll) {
  // add others
}

return response;
```

src/main/java/com/google/firebase/remoteconfig/internal/TemplateResponse.java
```diff
@@ -39,6 +39,9 @@
   @Key("version")
   private VersionResponse version;
 
+  @Key("etag")
```
hiranya911, line 42: Add a comment here indicating that this is never set by the HTTP response payload.

src/test/resources/rcTemplateWithETag.json
```diff
@@ -0,0 +1 @@
+{"conditions":[{"expression":"exp ios","name":"ios_en","tagColor":"INDIGO"},{"expression":"exp android","name":"android_en"}],"etag":"etag-0010201","parameterGroups":{"greetings_group":{"description":"description","parameters":{"greeting_text":{"conditionalValues":{"promo":{"useInAppDefault":true},"android":{"value":"hello android"},"ios":{"value":"hello ios"}},"defaultValue":{"useInAppDefault":true},"description":"greeting text"},"greeting_header":{"conditionalValues":{"promo":{"useInAppDefault":true},"android":{"value":"hello android"},"ios":{"value":"hello ios"}},"defaultValue":{"useInAppDefault":true},"description":"greeting header text"}}}},"parameters":{"greeting_text":{"conditionalValues":{"promo":{"useInAppDefault":true},"android":{"value":"hello android"},"ios":{"value":"hello ios"}},"defaultValue":{"useInAppDefault":true},"description":"greeting text"},"greeting_header":{"conditionalValues":{"promo":{"useInAppDefault":true},"android":{"value":"hello android"},"ios":{"value":"hello ios"}},"defaultValue":{"useInAppDefault":true},"description":"greeting header text"}},"version":{"description":"promo version","legacy":false}}
```
hiranya911, line 1: Add newline at eof

src/main/java/com/google/firebase/remoteconfig/RemoteConfigUtil.java
```diff
@@ -30,10 +31,52 @@ static boolean isValidVersionNumber(String versionNumber) {
     return !Strings.isNullOrEmpty(versionNumber) && versionNumber.matches("^\\d+$");
   }
 
+  static long convertToMilliseconds(String dateString) throws ParseException {
+    if (isUTCDateString(dateString)) {
```
hiranya911, line 35: This always ends up parsing at least twice. Can we improve that?

```java
try {
  return convertFromUtcZuluFormat(dateString); // common case first
} catch (ParseException e) {
  return convertFromUtcDateFormat(dateString); // fall back
}
```

src/main/java/com/google/firebase/remoteconfig/RemoteConfigUtil.java
```diff
@@ -30,10 +31,52 @@ static boolean isValidVersionNumber(String versionNumber) {
...
     SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS000000'Z'");
     dateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
     return dateFormat.format(new Date(millis));
   }
+
+  static String convertToUtcDateFormat(long millis) {
+    // sample output date string: Tue, 08 Dec 2020 15:49:51 GMT
+    checkArgument(millis >= 0, "Milliseconds duration must not be negative");
+    SimpleDateFormat dateFormat = new SimpleDateFormat("EEE, dd MMM yyyy HH:mm:ss z");
+    dateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
```
hiranya911, line 53: Can we assign `TimeZone.getTimeZone("UTC")` to a class-level constant?

src/main/java/com/google/firebase/remoteconfig/RemoteConfigUtil.java
```diff
@@ -30,10 +31,52 @@ static boolean isValidVersionNumber(String versionNumber) {
...
     checkArgument(millis >= 0, "Milliseconds duration must not be negative");
     SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS000000'Z'");
     dateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
     return dateFormat.format(new Date(millis));
   }
+
+  static String convertToUtcDateFormat(long millis) {
+    // sample output date string: Tue, 08 Dec 2020 15:49:51 GMT
+    checkArgument(millis >= 0, "Milliseconds duration must not be negative");
+    SimpleDateFormat dateFormat = new SimpleDateFormat("EEE, dd MMM yyyy HH:mm:ss z");
```
hiranya911, line 52: Use constants for pattern strings.

src/test/java/com/google/firebase/remoteconfig/TemplateTest.java
```diff
@@ -153,4 +159,193 @@ public void testEquality() {
...
+
+    assertNotNull(template.getParameters());
+    assertNotNull(template.getConditions());
+    assertNotNull(template.getParameterGroups());
+    assertTrue(template.getParameters().isEmpty());
+    assertTrue(template.getConditions().isEmpty());
+    assertTrue(template.getParameterGroups().isEmpty());
+    assertNull(template.getETag());
+
+    template = Template.fromJSON("{"
```
hiranya911, line 190: Make this a separate test case.

src/test/java/com/google/firebase/remoteconfig/TemplateTest.java
```diff
@@ -153,4 +159,193 @@ public void testEquality() {
...
+
+  @Test
+  public void testToJSON() {
+    // Empty template
+    String jsonString = new Template().toJSON();
+
+    assertEquals("{\"conditions\":[],"
+            + "\"parameterGroups\":{},\"parameters\":{}}", jsonString);
+
+    // Template with parameter values
```
hiranya911, line 263: Split into multiple test cases

src/test/java/com/google/firebase/remoteconfig/TemplateTest.java
```diff
@@ -153,4 +159,193 @@ public void testEquality() {
...
+            + "\"version\":{\"description\":\"template version\",\"legacy\":false,"
+            + "\"rollbackSource\":\"26\",\"updateOrigin\":\"ADMIN_SDK_NODE\","
+            + "\"updateTime\":\"Tue, 08 Dec 2020 15:49:51 UTC\","
+            + "\"updateType\":\"INCREMENTAL_UPDATE\",\"updateUser\":{"
+            + "\"email\":\"user@user.com\"},\"versionNumber\":\"34\"}}", jsonString);
+  }
+
+  @Test
+  public void testToJSONAndFromJSON() throws IOException {
+    String jsonString = new Template().toJSON();
```
hiranya911, line 317: How about something like this?

```java
Template original = new Template();
Template other = Template.fromJSON(original.toJSON());
assertEqual(other, original);
```

hiranya911 (changes requested):
Thanks for making the changes. Looks mostly good. Few comments on cleaning up the PR a bit.

src/main/java/com/google/firebase/remoteconfig/RemoteConfigUtil.java
lahirumaramba: Good point! Thanks!

src/test/resources/rcTemplateWithETag.json
lahirumaramba: We no longer need this file. Removed.

src/main/java/com/google/firebase/remoteconfig/RemoteConfigUtil.java
```diff
@@ -27,66 +27,52 @@
...
+    dateFormat.setTimeZone(UTC_TIME_ZONE);
     return dateFormat.format(new Date(millis));
   }
 
   static long convertFromUtcZuluFormat(String dateString) throws ParseException {
     // sample input date string: 2020-12-08T15:49:51.887878Z
     checkArgument(!Strings.isNullOrEmpty(dateString), "Date string must not be null or empty");
-    SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss");
-    dateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
+    SimpleDateFormat dateFormat = new SimpleDateFormat(ZULU_DATE_NO_FRAC_SECS_PATTERN);
```
hiranya911, line 66: The example input doesn't really match the pattern. You might want to remove the fractional seconds before parsing the timestamp.

src/main/java/com/google/firebase/remoteconfig/Template.java
```diff
@@ -46,11 +46,20 @@
...
-    parameterGroups = new HashMap<>();
+  public Template(String etag) {
+    this.parameters = new HashMap<>();
+    this.conditions = new ArrayList<>();
+    this.parameterGroups = new HashMap<>();
+    this.etag = etag;
+  }
+
+  Template() {
+    this.parameters = new HashMap<>();
```
hiranya911, line 60: `this(null)`

src/main/java/com/google/firebase/remoteconfig/Template.java
```diff
@@ -84,18 +93,24 @@ public Template() {
...
-  public static Template fromJSON(@NonNull String json) throws IOException {
+  public static Template fromJSON(@NonNull String json) throws FirebaseRemoteConfigException {
     checkArgument(!Strings.isNullOrEmpty(json), "JSON String must not be null or empty.");
     // using the default json factory as no rpc calls are made here
     JsonFactory jsonFactory = Utils.getDefaultJsonFactory();
-    JsonParser parser = jsonFactory.createJsonParser(json);
-    TemplateResponse templateResponse = parser.parseAndClose(TemplateResponse.class);
-    return new Template(templateResponse);
+    try {
+      return new Template(jsonFactory
```
hiranya911, line 107: Use a local variable and split into 2 steps.

src/main/java/com/google/firebase/remoteconfig/RemoteConfigUtil.java
```diff
@@ -20,20 +20,59 @@
...
 import com.google.common.base.Strings;
 
+import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.TimeZone;
 
 final class RemoteConfigUtil {
 
+  private static final String ZULU_DATE_PATTERN = "yyyy-MM-dd'T'HH:mm:ss.SSS000000'Z'";
```
hiranya911, line 30: Add a comment here indicating why we need 2 zulu format strings.

hiranya911 (changes requested):
Just a few minor nits.

src/test/java/com/google/firebase/remoteconfig/TemplateTest.java
```diff
@@ -153,4 +166,192 @@ public void testEquality() {
...
+
+    Version expectedVersion = Version.withDescription("promo version");
+    originalTemplate = new Template("etag-0010201")
+            .setParameters(PARAMETERS)
+            .setConditions(CONDITIONS)
+            .setParameterGroups(PARAMETER_GROUPS)
+            .setVersion(expectedVersion);
+    Template template = Template.fromJSON(originalTemplate.toJSON());
+
+    assertEquals("etag-0010201", template.getETag());
```
hiranya911, line 344: Why not run assertEquals on the template instances?

src/main/java/com/google/firebase/remoteconfig/RemoteConfigUtil.java
lahirumaramba: Good call. Thanks!

src/main/java/com/google/firebase/remoteconfig/RemoteConfigUtil.java
lahirumaramba: Moved the date string formatting to this util method and added comments to clarify the steps.

src/main/java/com/google/firebase/remoteconfig/Template.java
lahirumaramba: I have to cast null to String here for this to work... `this((String)null)`. Let me know if this is okay to use.

src/test/java/com/google/firebase/remoteconfig/TemplateTest.java
lahirumaramba: Condition arrays are not quite the same for a direct equality check. The original template contains a condition that does not have a tag color. After the serialization -> deserialization process that tag color will be set to `TagColor.UNSPECIFIED`.

hiranya911: Should we fix that so the behavior remains consistent?

src/main/java/com/google/firebase/remoteconfig/RemoteConfigUtil.java
```diff
@@ -20,20 +20,74 @@
...
+    // nanoseconds (up to 9 fractional seconds digits).
+    // SimpleDateFormat cannot handle fractional seconds, therefore we strip fractional seconds
+    // from the input date string before parsing.
+    // example: input -> "2014-10-02T15:01:23.045123456Z"
+    // formatted -> "2014-10-02T15:01:23"
+    int indexOfPeriod = dateString.indexOf(".");
+    if (indexOfPeriod != -1) {
+      dateString = dateString.substring(0, indexOfPeriod);
+    }
+    checkArgument(!Strings.isNullOrEmpty(dateString), "Date string must not be null or empty");
```
hiranya911, line 80: This should happen before everything else.

hiranya911 (approved):
LGTM with a suggestion.

src/main/java/com/google/firebase/remoteconfig/RemoteConfigUtil.java
lahirumaramba: Ah good catch! Thanks!

src/test/java/com/google/firebase/remoteconfig/TemplateTest.java
lahirumaramba: I think we should. Updated the code not to serialize the tag color as `TagColor.UNSPECIFIED` when `null` or empty.

lahirumaramba:
> LGTM with a suggestion.

Thanks for the review!! Made the suggested updates.

PR #501, opened 2020-12-10
Add Remote Config Integration tests

lahirumaramba:
- Add Remote Config Integration tests

Note: Added `release:stage` to trigger integration tests

Related to: #446

### DIFF
--- a/src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigIT.java
+++ b/src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigIT.java
@@ -138,7 +138,7 @@ public class FirebaseRemoteConfigIT {
 
     // Test list by batches
     final AtomicInteger collected = new AtomicInteger(0);
-    ListVersionsPage page = remoteConfig.listVersionsAsync().get();
+    ListVersionsPage page = remoteConfig.listVersions();
     while (page != null) {
       for (Version version : page.getValues()) {
         if (versions.contains(version.getVersionNumber())) {
@@ -151,7 +151,7 @@ public class FirebaseRemoteConfigIT {
 
     // Test iterate all
     collected.set(0);
-    page = remoteConfig.listVersionsAsync().get();
+    page = remoteConfig.listVersions();
     for (Version version : page.iterateAll()) {
       if (versions.contains(version.getVersionNumber())) {
         collected.incrementAndGet();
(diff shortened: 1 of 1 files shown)


src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigIT.java
```diff
@@ -0,0 +1,199 @@
...
+    assertEquals(String.format("Rollback to version %s", versionNumber),
+            rolledBackTemplate.getVersion().getDescription());
+
+    // get template to verify rollback
+    Template activeTemplate = remoteConfig.getTemplate();
+    assertEquals(rolledBackTemplate, activeTemplate);
+  }
+
+  @Test
+  public void testListVersions() throws Exception {
```
hiranya911, line 131: I'd suggest using the sync APIs here like the previous test case.

src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigIT.java
```diff
@@ -0,0 +1,199 @@
...
+  private static final Version VERSION = Version
+          .withDescription(String.format("promo config %s", timestamp));
+
+  @BeforeClass
+  public static void setUpClass() {
+    remoteConfig = FirebaseRemoteConfig.getInstance(IntegrationTestUtils.ensureDefaultApp());
+  }
+
+  @Test
+  public void testTemplateOperations() throws FirebaseRemoteConfigException {
```
hiranya911, line 83: Nit: Some of these can be split into separate test case with a bit of effort and extra RPCs (I'll leave it to you to decide whether that's worth it)

1. validate can be easily turned into a separate test case
2. getAtVersion and rollback can also be made into separate test cases if we are willing to run a few extra get and publish operations.

hiranya911 (approved):
LGTM with some suggestions.

src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigIT.java
lahirumaramba: Updated to use the sync api

src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigIT.java
lahirumaramba: Good point! As discussed offline let's address this in a future PR. Thanks!

PR #507, opened 2021-01-06
chore(rc): Refactor RC integration tests

lahirumaramba:
- Refactor RC integration tests by adding separate test per operation

### DIFF
--- a/src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigIT.java
+++ b/src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigIT.java
@@ -42,37 +42,11 @@ public class FirebaseRemoteConfigIT {
 
   private static FirebaseRemoteConfig remoteConfig;
 
-  private static final long timestamp = System.currentTimeMillis();
-
-  private static final Map<String, Parameter> PARAMETERS = ImmutableMap.of(
-          "welcome_message_text", new Parameter()
-                  .setDefaultValue(ParameterValue
-                          .of(String.format("welcome to app %s", timestamp)))
-                  .setConditionalValues(ImmutableMap.<String, ParameterValue>of(
-                          "ios_en",
-                          ParameterValue.of(String.format("welcome to app en %s", timestamp))
-                  ))
-                  .setDescription("text for welcome message!"),
-          "header_text", new Parameter()
-                  .setDefaultValue(ParameterValue.inAppDefault()));
-
-  private static final Map<String, ParameterGroup> PARAMETER_GROUPS = ImmutableMap.of(
-          "new menu", new ParameterGroup()
-                  .setDescription(String.format("New Menu %s", timestamp))
-                  .setParameters(ImmutableMap.of(
-                          "pumpkin_spice_season", new Parameter()
-                                  .setDefaultValue(ParameterValue.of("true"))
-                                  .setDescription("Whether it's currently pumpkin spice season."))
-                  ));
-
   private static final List<Condition> CONDITIONS = ImmutableList.of(
           new Condition("ios_en", "device.os == 'ios' && device.country in ['us', 'uk']")
                   .setTagColor(TagColor.INDIGO),
           new Condition("android_en",
                   "device.os == 'android' && device.country in ['us', 'uk']"));
-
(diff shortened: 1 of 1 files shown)


src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigIT.java
```diff
@@ -42,89 +42,133 @@
...
-    assertNotEquals(VERSION, publishedTemplate.getVersion());
+    assertNotEquals(version, publishedTemplate.getVersion());
+  }
+
+  @Test
+  public void testGetTemplate() throws FirebaseRemoteConfigException {
+    // get template to fetch the active template with correct etag
+    // modify and publish a known template to test get template operation.
+    final Template oldTemplate = remoteConfig.getTemplate();
+    final Template inputTemplate = Template.fromJSON(oldTemplate.toJSON());
```
hiranya911, line 101: Are you using this as a way to "clone" the original template? If so might be better to add a `cloneTemplate()` helper method with this logic.

hiranya911: Also we probably don't need to clone it for this test case?

src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigIT.java
```diff
@@ -42,89 +42,133 @@
...
     Template publishedTemplate = remoteConfig.publishTemplate(inputTemplate);
+
     assertNotEquals(inputTemplate.getETag(), publishedTemplate.getETag());
-    assertEquals(PARAMETERS, publishedTemplate.getParameters());
-    assertEquals(PARAMETER_GROUPS, publishedTemplate.getParameterGroups());
+    assertEquals(parameters, publishedTemplate.getParameters());
+    assertEquals(parameterGroups, publishedTemplate.getParameterGroups());
     assertEquals(CONDITIONS, publishedTemplate.getConditions());
-    assertNotEquals(VERSION, publishedTemplate.getVersion());
+    assertNotEquals(version, publishedTemplate.getVersion());
```
hiranya911, line 93: Shouldn't they have the same description?

src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigIT.java
```diff
@@ -42,89 +42,133 @@
...
+  }
+
+  @Test
+  public void testPublishTemplate() throws FirebaseRemoteConfigException {
     // get template to fetch the active template with correct etag
     final Template oldTemplate = remoteConfig.getTemplate();
     final Template inputTemplate = Template.fromJSON(oldTemplate.toJSON());
-    final String versionNumber = oldTemplate.getVersion().getVersionNumber();
-
+    final Map<String, Parameter> parameters = getParameters();
```
hiranya911, line 77: Probably doesn't need to clone the template

hiranya911 (changes requested):
Looks pretty good. Just a few questions to address.

src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigIT.java
lahirumaramba: Good catch! Thanks!

src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigIT.java
lahirumaramba: They should! Added a new assertEquals for description. Thanks!

src/test/java/com/google/firebase/remoteconfig/FirebaseRemoteConfigIT.java
lahirumaramba: Good catch. Thanks! This was only needed in the previous design. Removed the cloning from all the tests.
We only need to copy the etag for `testGetTemplateAtVersion()`. I changed the code there to use the constructor, instead.

hiranya911 (approved):
Thanks. LGTM 👍

Issue #506, opened 2021-01-06
unexpected/unparsable response


qeepcologne:
When calling FirebaseMessaging.getInstance().send(message), sometimes it failes with strange exception.
Sometimes no errors for a day, sometime hundreds or thousands errors. Average error rate was about 3% mid december
but is much higher today.

i checked Firebase Cloud Messaging Api in Google Cloud Platform console and see increasing
404 answer for Api Method google.firebase.fcm.v1.FcmService.SendMessage

I tried to understand the code and i think the http answer is not as expected.
Probably this is related to the firebase server infrastructure, but sdk should not fail on parsing the error and answer
some server Exception instead?.

`java.lang.IllegalArgumentException: key details
        at com.google.api.client.json.JsonParser.parseValue(JsonParser.java:900) ~[google-http-client-1.38.0.jar:1.38.0]
        at com.google.api.client.json.JsonParser.parse(JsonParser.java:360) ~[google-http-client-1.38.0.jar:1.38.0]
        at com.google.api.client.json.JsonParser.parse(JsonParser.java:318) ~[google-http-client-1.38.0.jar:1.38.0]
        at com.google.api.client.json.JsonParser.parseAndClose(JsonParser.java:162) ~[google-http-client-1.38.0.jar:1.38.0]
        at com.google.api.client.json.JsonParser.parseAndClose(JsonParser.java:144) ~[google-http-client-1.38.0.jar:1.38.0]
        at com.google.firebase.messaging.FirebaseMessagingClientImpl$MessagingErrorHandler.safeParse(FirebaseMessagingClientImpl.java:308) ~[firebase-admin-7.1.0.jar:?]
        at com.google.firebase.messaging.FirebaseMessagingClientImpl$MessagingErrorHandler.createException(FirebaseMessagingClientImpl.java:292) ~[firebase-admin-7.1.0.jar:?]
        at com.google.firebase.messaging.FirebaseMessagingClientImpl$MessagingErrorHandler.createException(FirebaseMessagingClientImpl.java:282) ~[firebase-admin-7.1.0.jar:?]
        at com.google.firebase.internal.AbstractHttpErrorHandler.handleHttpResponseException(AbstractHttpErrorHandler.java:57) ~[firebase-admin-7.1.0.jar:?]
        at com.google.firebase.internal.ErrorHandlingHttpClient.send(ErrorHandlingHttpClient.java:108) ~[firebase-admin-7.1.0.jar:?]
[...]

PR #508, opened 2021-01-07
fix(fcm): Gracefully handling non-json error responses

hiranya911:
FCM service seems to occasionally respond with non-json (HTML) error responses, and if these responses contain certain special characters (e.g. `[` or `]`), the SDK throws a cryptic JSON parse exception (see #506). 

The `MessagingErrorHandler` is supposed to catch all parse errors, but today it only handles IOExceptions thrown by the parser. But in the above case parser throws an IllegalArgumentException. This PR addresses this issue by updating the `MessagingErrorHandler` to catch *all* exceptions thrown by the parser. This makes sure the developer gets a `FirebaseMessagingException` with the original error response attached to it, when the above error condition occurs.

RELEASE NOTE: Improved support for handling certain non-JSON error responses sent by the FCM backend service.

### DIFF
--- a/src/test/java/com/google/firebase/messaging/FirebaseMessagingClientImplTest.java
+++ b/src/test/java/com/google/firebase/messaging/FirebaseMessagingClientImplTest.java
@@ -82,7 +82,7 @@ public class FirebaseMessagingClientImplTest {
       .setTopic("test-topic")
       .build();
   private static final List<Message> MESSAGE_LIST = ImmutableList.of(EMPTY_MESSAGE, EMPTY_MESSAGE);
-  
+
   private static final boolean DRY_RUN_ENABLED = true;
   private static final boolean DRY_RUN_DISABLED = false;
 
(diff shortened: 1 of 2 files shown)


src/test/java/com/google/firebase/messaging/FirebaseMessagingClientImplTest.java
```diff
@@ -200,14 +200,14 @@ public void testSendErrorWithZeroContentResponse() {
   @Test
   public void testSendErrorWithMalformedResponse() {
     for (int code : HTTP_ERRORS) {
-      response.setStatusCode(code).setContent("not json");
+      response.setStatusCode(code).setContent("[not json]");
```
chong-shao, line 203: I wonder if a text that looks more like HTML can be more representative here, e.g., <html>"not json"</html> ?

chong-shao (approved):

src/test/java/com/google/firebase/messaging/FirebaseMessagingClientImplTest.java
hiranya911: From my tests and inspection of the `JsonParser` source, we need more than just HTML tags to trigger this IllegalArgumentException. Square brackets seems to be the easiest and most reliable way. Although it makes me wonder what kind of response was served by FCM to cause this error in the first place.

PR #522, opened 2021-04-06
feat: enables OIDC auth code flow

lsirac:
Provides an option for developers to specify the OAuth response type for their OIDC provider (either one of these can be set:):
- id_token
- code (if set, must also set the client secret)

### DIFF
--- a/src/main/java/com/google/firebase/auth/OidcProviderConfig.java
+++ b/src/main/java/com/google/firebase/auth/OidcProviderConfig.java
@@ -79,6 +79,13 @@ public final class OidcProviderConfig extends ProviderConfig {
         "Invalid OIDC provider ID (must be prefixed with 'oidc.'): " + providerId);
   }
 
+  static Map<String, Boolean> ensureResponseType(Map<String,Object> properties) {
+    if (properties.get("responseType") == null) {
+      properties.put("responseType", new HashMap<String, Boolean>());
+    }
+    return (Map<String, Boolean>) properties.get("responseType");
+  }
+
   /**
    * A specification class for creating a new OIDC Auth provider.
    *
@@ -158,10 +165,7 @@ public final class OidcProviderConfig extends ProviderConfig {
      * @param enabled A boolean signifying whether the code response type is supported.
      */
     public CreateRequest setCodeResponseType(boolean enabled) {
-      if (properties.get("responseType") == null) {
-        properties.put("responseType", new HashMap<String, Boolean>());
-      }
-      Map<String, Boolean> map = (Map<String, Boolean>) properties.get("responseType");
+      Map<String, Boolean> map = ensureResponseType(properties);
       map.put("code", enabled);
       return this;
     }
@@ -175,11 +179,7 @@ public final class OidcProviderConfig extends ProviderConfig {
      * @param enabled A boolean signifying whether the ID token response type is supported.
      */
     public CreateRequest setIdTokenResponseType(boolean enabled) {
-      if (properties.get("responseType") == null) {
-        properties.put("responseType", new HashMap<String, Boolean>());
-      }
-
-      Map<String, Boolean> map = (Map<String, Boolean>) properties.get("responseType");
+      Map<String, Boolean> map = ensureResponseType(properties);
       map.put("idToken", enabled);
       return this;
     }
@@ -265,10 +265,7 @@ public final class OidcProviderConfig extends ProviderConfig {
      * @param enabled A boolean signifying whether the code response type is supported.
      */
     public UpdateRequest setCodeResponseType(boolean enabled) {
-      if (properties.get("responseType") == null) {
-        properties.put("responseType", new HashMap<String, Boolean>());
-      }
-      Map<String, Boolean> map = (Map<String, Boolean>) properties.get("responseType");
+      Map<String, Boolean> map = ensureResponseType(properties);
       map.put("code", enabled);
       return this;
     }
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/firebase/auth/OidcProviderConfig.java
```diff
@@ -31,17 +34,31 @@
...
 
+  public String getClientSecret() {
+    return clientSecret;
+  }
+
   public String getIssuer() {
     return issuer;
   }
 
+  public GenericJson getResponseType() {
```
hiranya911, line 58: This cannot really be part of the public API surface. Should use a more standard return type like Map or something. Or may be add two separate getters to mirror the setters you've added below:

```
boolean isCodeResponseType()
boolean isIdTokenResponseType()
```

src/main/java/com/google/firebase/auth/OidcProviderConfig.java
```diff
@@ -113,6 +143,43 @@ public CreateRequest setIssuer(String issuer) {
...
+     * enabled if no response type is specified.
+     *
+     * <p>A client secret must be set for this response type.
+     *
+     * <p>Having both the code and ID token response flows is currently not supported.
+     *
+     * @param enabled A boolean signifying whether the code response type is supported.
+     */
+    public CreateRequest setCodeResponseType(boolean enabled) {
+      if (properties.get("responseType") == null) {
```
hiranya911, line 157: You can simplify this and reduce some of the duplication by introducing a helper:

```
private Map<String, Boolean> ensureResponseType() {
    if (properties.get("responseType") == null) {
        properties.put("responseType", new HashMap<String, Boolean>());
    }

   return (Map<String, Boolean>) properties.get("responseType");
}
```

src/test/java/com/google/firebase/auth/OidcProviderConfigTest.java
```diff
@@ -157,4 +196,14 @@ public void testUpdateRequestMissingIssuer() {
   public void testUpdateRequestInvalidIssuerUrl() {
     new OidcProviderConfig.UpdateRequest("oidc.provider-id").setIssuer("not a valid url");
   }
+
```
hiranya911, line 199: What happens if the developer sets responseTypeCode but doesn't specify a client secret? Should the SDK validate that? Let's at least have a test case to exercise that scenario.

hiranya911 (changes requested):
Looks mostly good. But we cannot have `GenericJson` exposed in our public API surface. Plus few other nits.

src/test/java/com/google/firebase/auth/OidcProviderConfigTest.java
lsirac: I'm letting the backend deal with it - they will return an error. I can't do this check in update, and if you want me to make this check in create then we'll be forcing the dev to call setClientSecret before calling setCodeResponseType. If you prefer I do that let me know.

src/test/java/com/google/firebase/auth/FirebaseUserManagerTest.java
```diff
@@ -1490,7 +1490,10 @@ public void testCreateOidcProvider() throws Exception {
             .setDisplayName("DISPLAY_NAME")
             .setEnabled(true)
             .setClientId("CLIENT_ID")
-            .setIssuer("https://oidc.com/issuer");
+            .setClientSecret("CLIENT_SECRET")
+            .setIssuer("https://oidc.com/issuer")
+            .setCodeResponseType(true)
+            .setIdTokenResponseType(true);
```
hiranya911, line 1496: In the Node.js SDK we are adding checks to make sure that exactly one is true, when both code and idToken response types are specified. If we want we can do something similar here too. But it will probably require some additional changes.

In the `AbstractCreateRequest` and `AbstractUpdateRequest` classes:
```java
protected void validate() { }

Map<String, Object> getProperties() {
    this.validate();
    return ImmutableMap.copyOf(properties);
}
```

Now in `OidcProviderConfig.CreateRequest` and `OidcProviderConfig.UpdateRequest`:
```java
private static void validateResponseTypes(Map<String, Object> properties) {
  // Check if clientSecret is set when code = true
  // Check that when both code and idToken are specified, they are not both true or both false
}

@Override
protected void validate() {
  validateResponseTypes(this.properties);
}
```

Take it as a suggestion. I think for the initial release, the way it's implemented now is probably fine.

src/test/java/com/google/firebase/auth/OidcProviderConfigTest.java
```diff
@@ -47,7 +55,35 @@ public void testJsonDeserialization() throws IOException {
...
+    Map<String, Object> properties = new HashMap<>();
+
+    Map<String, Boolean> responseType = OidcProviderConfig.ensureResponseType(properties);
+
+    assertNotNull(responseType);
+    assertEquals(responseType, properties.get("responseType"));
+  }
+
+  @Test
+  public void testEnsureResponseType_alreadyPresent() {
```
hiranya911, line 75: We prefer regular camel-cased naming for test cases: `testEnsureResponseWhenAlreadyPresent()`

src/test/java/com/google/firebase/auth/OidcProviderConfigTest.java
hiranya911: See my other comment.

hiranya911 (approved):
LGTM with a couple of suggestions.

Issue #493, opened 2020-11-10
FR: add ability to communicate with Auth Emulator


branks:
On the emulator documentation it states the the Node SDK can connect to the Firebase Authentication Emulator by using an environment variable - https://firebase.google.com/docs/emulator-suite/connect_auth

It would be great if this could also be supported in the Java SDK

dv-bv:
Any timelines on this one? Any way we could help getting this through the door?

yuchenshi:
@dv-bv Sure! We welcome Pull Requests that add support for the environment variable to the Java Admin SDK.

Just keep in mind that we're still having some important discussions on `verifyIdToken` so that may not land soon. We'd be happy to review PRs for supporting all other methods. (And by the way, while we do not commit to timelines, your can definitely help to get this through the door.

7agustibm:
Hi! I start this PR: https://github.com/firebase/firebase-admin-java/pull/510

7agustibm:
Hi! When will the new version be available with this feature?

dmitri-gb:
> Just keep in mind that we're still having some important discussions on `verifyIdToken` so that may not land soon.

@yuchenshi, are there any updates on those discussions about `verifyIdToken`? I noticed that the Node Admin SDK [was recently changed](https://github.com/firebase/firebase-admin-node/pull/1148) to always accept unsigned tokens when the emulator is in use. Will the Java SDK follow suit?

yuchenshi:
@dmitri-gb As you've observed, we've reached some conclusions regarding `verifyIdToken` and we're now working on implementing consistent support across Admin SDKs. Namely, `verifyIdToken` will accept unsigned tokens if the environment variable is set and the emulator is running.

While we're unable to provide ETAs, please rest assured we're building on top of @7agustibm's amazing work for network communications and we'll release once we have ID token support as well.

at15:
A snippet for connecting to auth emulator and import user https://firebase.google.com/docs/auth/admin/import-users#java

```java
package foo;

import java.util.ArrayList;
import java.util.List;

import com.google.firebase.FirebaseApp;
import com.google.firebase.FirebaseOptions;
import com.google.firebase.auth.FirebaseAuth;
import com.google.firebase.auth.FirebaseAuthException;
import com.google.firebase.auth.ImportUserRecord;
import com.google.firebase.auth.UserImportOptions;
import com.google.firebase.auth.hash.HmacSha256;
[...]

PR #523, opened 2021-04-08
feat(auth): Add auth emulator support

ssbushi:
### Discussion

Add support to communicate with the Auth Emulator #493.

### Testing

  * Unit tests.
  * Manual testing.

### API Changes

  * N/A.


src/test/java/com/google/firebase/auth/multitenancy/FirebaseTenantClientTest.java
```diff
@@ -312,6 +321,20 @@ public void testDeleteTenantWithNotFoundError() {
     checkUrl(interceptor, "DELETE", TENANTS_BASE_URL + "/UNKNOWN");
   }
 
+  @Test
```
hiranya911, line 324: Ideally we need a way to test all APIs in the class with the emulator.

src/test/java/com/google/firebase/auth/FirebaseTokenVerifierImplTest.java
```diff
@@ -399,20 +399,6 @@ public void testVerifyTokenMissingTenantId() {
     }
   }
```
hiranya911, line 401: Why was this removed?

src/test/java/com/google/firebase/auth/FirebaseEmulatorTokenVerifierTest.java
```diff
@@ -0,0 +1,402 @@
...
+import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.firebase.ErrorCode;
+import java.io.IOException;
+import java.util.concurrent.TimeUnit;
+import org.junit.Before;
+import org.junit.Test;
+
+public class FirebaseEmulatorTokenVerifierTest {
```
hiranya911, line 42: Most of this can be combined with the existing tests once the emulator support is added to the existing `FirebaseTokenVerfierImpl` class.

hiranya911 (changes requested):
Thanks @ssbushi. Looks mostly good. My main observations are:

1. We should avoid duplication by adding emulator support to the existing `FirebaseTokenVerfierImpl` class, as opposed to implementing a whole new class.
2. Avoid duplication by putting env variable lookups in a shared util.
3. We need a better way to test the functionality of all APIs in emulator mode. In other languages we used techniques like test parameterization and fixtures to achieve this. We should try to find an equivalent strategy for junit4 (although it's ok to do so in a separate PR).

src/test/java/com/google/firebase/auth/FirebaseTokenVerifierImplTest.java
ssbushi: Sorry, I forgot to add a comment here.

This test passes, but only because the `verifyToken` does not throw. The current behaviour is that the token verifier silently ignores the `tenantId` in the input token, if the verfier itself is not built with a `tenantId` set.

I was wondering if this is the intended behavior or if we expect the verifier to throw (in which case I can update the implementation and add a `fail()` in the test's try block). 

I haven't tested other tests in this file, but since they lack a `fail()` in the try block, I suspect that there might be a few that are passing regardless of the test conditions.

hiranya911: The verifier is initialized with a tenantId in this case if I'm not mistaken:

```java
fullyPopulatedBuilder()
        .setTenantId("TENANT_ID") // <----
        .build()
        .verifyToken(tokenFactory.createToken());
```

We should be able to add `fail()` statement to the `try` block.

ssbushi: I'm sorry, I don't get it. This test here does not have a verifier with a tenantId.

```
fullyPopulatedBuilder()
        .build()
        .verifyToken(createTokenWithTenantId("TENANT_ID"));
```


This test is same as the `testVerifyTokenWithTenantId` but it's expecting a different behaviour.

hiranya911: This is testing a regular token against a tenant-aware verifier (this should throw):

```
fullyPopulatedBuilder()
        .setTenantId("TENANT_ID")
        .build()
        .verifyToken(tokenFactory.createToken());
```

This is testing a tenant-scoped token against a tenant-unaware verifier:

```
fullyPopulatedBuilder()
        .build()
        .verifyToken(createTokenWithTenantId("TENANT_ID"));
```

This used to throw the same error, but that behavior was changed in #475. This test case seems like something left over from before the change. We can update it to reflect the current behavior.

src/test/java/com/google/firebase/auth/FirebaseTokenVerifierImplTest.java
```diff
@@ -384,21 +425,6 @@ public void testVerifyTokenNoTenantId() {
...
-        .build()
-        .verifyToken(tokenFactory.createToken());
-    } catch (FirebaseAuthException e) {
-      assertEquals(AuthErrorCode.TENANT_ID_MISMATCH, e.getAuthErrorCode());
-      assertEquals(
-          "The tenant ID ('') of the token did not match the expected value ('TENANT_ID')",
-          e.getMessage());
-    }
-  }
-
```
ssbushi, line 401: Removing this test as this is the same as `testVerifyTokenNoTenantId` above.

src/main/java/com/google/firebase/auth/FirebaseEmulatorTokenVerifier.java
```diff
@@ -0,0 +1,280 @@
...
+import com.google.common.base.Strings;
+import com.google.firebase.ErrorCode;
+import com.google.firebase.internal.Nullable;
+import java.io.IOException;
+
+/**
+ * The implementation of the {@link FirebaseTokenVerifier} interface, for the Auth Emulator. Can be
+ * customized to verify both Firebase ID tokens and session cookies.
+ */
+final class FirebaseEmulatorTokenVerifier implements FirebaseTokenVerifier {
```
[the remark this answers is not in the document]
ssbushi: Thanks Hiranya.

I've followed your advice and moved the logic over to `FirebaseTokenVerifierImpl` although I didn't add `isEmulatorMode` as a member. Instead the emulator mode is activated by checking for the respective environment variable. I thought this is more seamless, however I'm curious what you think.

src/main/java/com/google/firebase/auth/FirebaseTokenUtils.java
```diff
@@ -62,7 +63,8 @@ static FirebaseTokenFactory createTokenFactory(
       return new FirebaseTokenFactory(
           firebaseApp.getOptions().getJsonFactory(),
           clock,
-          CryptoSigners.getCryptoSigner(firebaseApp),
+          useEmulator() ? new EmulatorCryptoSigner()
```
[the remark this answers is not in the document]
ssbushi: Done.

src/main/java/com/google/firebase/auth/FirebaseUserManager.java
```diff
@@ -100,6 +102,18 @@ private FirebaseUserManager(Builder builder) {
...
+  private String getIdToolkitUrl(String projectId, String version) {
+    String maybeEmulator = getEmulatorHost();
+    if (maybeEmulator != null) {
+      return String.format(ID_TOOLKIT_URL_EMULATOR, maybeEmulator, version, projectId);
+    }
+    return String.format(ID_TOOLKIT_URL, version, projectId);
+  }
+
+  private String getEmulatorHost() {
+    return System.getenv("FIREBASE_AUTH_EMULATOR_HOST");
```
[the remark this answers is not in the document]
ssbushi: Done.

src/main/java/com/google/firebase/auth/FirebaseUserManager.java
```diff
@@ -100,6 +102,18 @@ private FirebaseUserManager(Builder builder) {
     this.httpClient = new AuthHttpClient(jsonFactory, builder.requestFactory);
   }
 
+  private String getIdToolkitUrl(String projectId, String version) {
+    String maybeEmulator = getEmulatorHost();
+    if (maybeEmulator != null) {
```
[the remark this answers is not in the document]
ssbushi: Done.

src/test/java/com/google/firebase/auth/FirebaseUserManagerTest.java
```diff
@@ -2678,6 +2684,33 @@ public void testTenantAwareDeleteSamlProviderConfig() throws Exception {
     checkUrl(interceptor, "DELETE", expectedUrl);
   }
 
+  @Test
```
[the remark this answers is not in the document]
ssbushi: Agreed. I propose to defer that to a new PR, to avoid making the current one too big.

src/test/java/com/google/firebase/auth/multitenancy/FirebaseTenantClientTest.java
ssbushi: Agreed. See the previous comment.

src/test/java/com/google/firebase/auth/FirebaseEmulatorTokenVerifierTest.java
ssbushi: Done.

src/main/java/com/google/firebase/auth/FirebaseTokenUtils.java
```diff
@@ -87,8 +87,9 @@ static FirebaseTokenVerifierImpl createIdTokenVerifier(
         clock, ID_TOKEN_ISSUER_PREFIX, projectId);
     GooglePublicKeysManager publicKeysManager = newPublicKeysManager(
         app.getOptions(), clock, ID_TOKEN_CERT_URL);
+    String shortName = Utils.isEmulatorMode() ? "Emulator ID token" : "ID token";
```
hiranya911, line 90: I don't think this is necessary. Let's use the same error messages/text as the production mode. That's how this is implemented in other languages too.

src/main/java/com/google/firebase/auth/FirebaseTokenVerifierImpl.java
```diff
@@ -96,7 +97,9 @@ private FirebaseTokenVerifierImpl(Builder builder) {
   public FirebaseToken verifyToken(String token) throws FirebaseAuthException {
     IdToken idToken = parse(token);
     checkContents(idToken);
-    checkSignature(idToken);
+    if (!isEmulatorMode()) {
```
hiranya911, line 100: Personally I prefer the ability to programmatically enable the emulator mode, instead of just tying it to an environment variable. That makes unit testing a lot easier. However, this is the approach we've taken in our Node.js SDK so we can live with this.

But let's make sure we only lookup the env variable once per token verification. So assign the result to a local variable, and use it in rest of the method body:

```
boolean emulator = Utils.isEmulatorMode();
```

src/main/java/com/google/firebase/auth/Utils.java
```diff
@@ -0,0 +1,19 @@
+package com.google.firebase.auth;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Strings;
+
+public class Utils {
```
hiranya911, line 6: This class has to package-protected. If it must be public, then it should be moved to the `auth.internal` package.

src/main/java/com/google/firebase/auth/Utils.java
```diff
@@ -0,0 +1,19 @@
+package com.google.firebase.auth;
```
hiranya911, line 1: Add license header.

src/main/java/com/google/firebase/auth/Utils.java
```diff
@@ -0,0 +1,19 @@
...
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Strings;
+
+public class Utils {
+  @VisibleForTesting
+  public static final String AUTH_EMULATOR_HOST = "FIREBASE_AUTH_EMULATOR_HOST";
+
+  public static boolean isEmulatorMode() {
+    return !Strings.isNullOrEmpty(
+        System.getenv(AUTH_EMULATOR_HOST));
```
hiranya911, line 12: getEmulatorHost()

src/test/java/com/google/firebase/auth/TestTokenFactory.java
```diff
@@ -86,4 +112,11 @@ public String createToken(JsonWebSignature.Header header, JsonWebToken.Payload p
     payload.setSubject(UID);
     return payload;
   }
+
+  public JsonWebSignature.Header createHeaderForEmulator() {
+    JsonWebSignature.Header header = new JsonWebSignature.Header();
+    header.setAlgorithm("NONE");
```
hiranya911, line 118: It should be "none" as per the JWS spec.

src/test/java/com/google/firebase/auth/TestTokenFactory.java
```diff
@@ -69,6 +70,31 @@ public String createToken(JsonWebSignature.Header header, JsonWebToken.Payload p
...
+  }
+
+  public String createUnsignedTokenForEmulator(JsonWebToken.Payload payload) {
+    return createUnsignedTokenForEmulator(createHeaderForEmulator(), payload);
+  }
+
+  public String createUnsignedTokenForEmulator(
+      JsonWebSignature.Header header, JsonWebSignature.Payload payload) {
+    try {
+      String content =
```
hiranya911, line 88: Readability nit:

```
String encodedHeader = Base64.encodeBase64URLSafeString(JSON_FACTORY.toByteArray(header));
String encodedPayload = Base64.encodeBase64URLSafeString(JSON_FACTORY.toByteArray(payload));
return String.format("%s.%s.", encodedHeader, encodedPayload);
```

src/test/java/com/google/firebase/auth/FirebaseTokenVerifierImplTest.java
```diff
@@ -443,6 +455,258 @@ public void testBuilderNoDocUrl() {
     fullyPopulatedBuilder().setDocUrl(null).build();
   }
 
+  @Test
```
hiranya911, line 458: Might be just easier to create a new test class `EmulatorFirebaseTokenVerifierImpleTest` for all these new test cases. You can do the env set up in the `@Before` fixture.

hiranya911 (changes requested):
Thanks for making the changes. Looks pretty good. Just a few more minor updates needed.

src/main/java/com/google/firebase/auth/FirebaseTokenUtils.java
ssbushi: Done.

src/main/java/com/google/firebase/auth/FirebaseTokenVerifierImpl.java
ssbushi: Done.

src/main/java/com/google/firebase/auth/Utils.java
ssbushi: Done.

src/main/java/com/google/firebase/auth/Utils.java
ssbushi: Done.

src/test/java/com/google/firebase/auth/TestTokenFactory.java
ssbushi: Done.

src/test/java/com/google/firebase/auth/TestTokenFactory.java
ssbushi: Done.

src/test/java/com/google/firebase/auth/FirebaseTokenVerifierImplTest.java
ssbushi: Done.

src/main/java/com/google/firebase/auth/Utils.java
ssbushi: Done.

src/main/java/com/google/firebase/auth/FirebaseTokenVerifierImpl.java
```diff
@@ -97,7 +99,7 @@ private FirebaseTokenVerifierImpl(Builder builder) {
   public FirebaseToken verifyToken(String token) throws FirebaseAuthException {
     IdToken idToken = parse(token);
     checkContents(idToken);
-    if (!isEmulatorMode()) {
+    if (!isEmulatorMode) {
```
hiranya911, line 102: If you want you can make `isEmulatorMode` a local variable here instead of a class attribute.

hiranya911 (approved):
LGTM 👍

src/test/java/com/google/firebase/auth/EmulatorFirebaseTokenVerifierImplTest.java
```diff
@@ -0,0 +1,328 @@
...
+    tokenVerifier.verifyToken(token);
+  }
+
+  @Test
+  public void testVerifyTokenIncorrectAlgorithm() {
+    String token = createTokenWithIncorrectAlgorithm();
+
+    try {
+      tokenVerifier.verifyToken(token);
+    } catch (FirebaseAuthException e) {
```
yuchenshi, line 86: This shouldn't throw -- please update accordingly (similar to next test). (FYI it works right now because it doesn't `fail` when no error is thrown, but the intention is still wrong)

yuchenshi (approved):

src/test/java/com/google/firebase/auth/EmulatorFirebaseTokenVerifierImplTest.java
ssbushi: Good catch. Thanks!

src/main/java/com/google/firebase/auth/FirebaseTokenVerifierImpl.java
ssbushi: Done. Thanks!

PR #532, opened 2021-04-27
feat(auth): enables OIDC auth code flow

lsirac:
Provides an option for developers to specify the OAuth response type for their OIDC provider (either one of these can be set:):

- id_token
- code (if set, must also set the client secret)


egilmorez (approved):
LG, thanks!

PR #546, opened 2021-05-03
fix(auth): adds missing EMAIL_NOT_FOUND error code

bojeil-google:
Catch EMAIL_NOT_FOUND when /accounts:sendOobCode is called for password reset on a user that does not exist.

### DIFF
--- a/src/main/java/com/google/firebase/auth/AuthErrorCode.java
+++ b/src/main/java/com/google/firebase/auth/AuthErrorCode.java
@@ -37,6 +37,13 @@ public enum AuthErrorCode {
   EMAIL_ALREADY_EXISTS,
 
   /**
+   * No user record found for the given email, typically raised when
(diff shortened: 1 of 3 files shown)


hiranya911 (approved):
LGTM 👍 

Should we add a test case? Something similar to: https://github.com/firebase/firebase-admin-java/blob/7352a43a45bc81227354d06d1ce95713cf4408b8/src/test/java/com/google/firebase/auth/FirebaseUserManagerTest.java#L2771

bojeil-google:
> LGTM 👍
> 
> Should we add a test case? Something similar to:
> 
> https://github.com/firebase/firebase-admin-java/blob/7352a43a45bc81227354d06d1ce95713cf4408b8/src/test/java/com/google/firebase/auth/FirebaseUserManagerTest.java#L2771

Done

hiranya911 (approved):

PR #573, opened 2021-06-14
fix: Using Java 8 in release and nightly builds

hiranya911:
This makes it possible to upgrade a wide range of the tools/dependencies we use. We will continue to compile source down to Java 1.7 compatibility (required config already present in `pom.xml`), and therefore the resulting artifacts will work on Java 7. Added Java 7, 8 and 11 to the CI build matrix for added coverage.

Environment variable manipulation in tests do not work on Java 8 and above, and therefore implemented some workarounds for that too.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,13 +19,17 @@ jobs:
   build:
     runs-on: ubuntu-latest
 
+    strategy:
+      matrix:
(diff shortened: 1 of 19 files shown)


lahirumaramba (approved):
Great! Thanks, Hiranya!
LGTM!

PR #576, opened 2021-06-28
Drop 500 from the retryable errorcode list

chong-shao:
[Google Cloud documentation](https://cloud.google.com/apis/design/errors#retrying_errors) has been updated to not consider 500 as a retryable error code. Update Admin SDK to align with that


hiranya911 (approved):
I'm pleasantly surprised that this didn't break more unit tests :)

LGTM 👍

chong-shao:
Thanks!

PR #580, opened 2021-07-07
Set trimStackTrace property to false

lahirumaramba:
Set `trimStackTrace` to `false` in mvn failsafe plugin.
This will print the full stack trace for failed integration tests making it easier to investigate nightly build failures.


hiranya911 (approved):
LGTM 👍

PR #585, opened 2021-08-06
fix(auth): check if the user is disabled on checkRevoked=true for verifyIdToken and verifySessionCookie

lsirac:
When `verifyIdToken` or `verifySessionCookie` is called with `checkResolved` set to `true` we also check if the user is disabled. If so, a `FirebaseAuthException` with the new AuthErrorCode `USER_DISABLED` is thrown.

### DIFF
--- a/src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
+++ b/src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
@@ -256,11 +256,13 @@ public abstract class AbstractFirebaseAuth {
    * API call.
    *
    * @param idToken A Firebase ID token string to parse and verify.
-   * @param checkRevoked A boolean denoting whether to check if the tokens were revoked.
+   * @param checkRevoked A boolean denoting whether to check if the tokens were revoked or if
+   *                     the user is disabled.
    * @return A {@link FirebaseToken} representing the verified and decoded token.
    * @throws IllegalArgumentException If the token is null, empty, or if the {@link FirebaseApp}
    *     instance does not have a project ID associated with it.
-   * @throws FirebaseAuthException If an error occurs while parsing or validating the token.
+   * @throws FirebaseAuthException If an error occurs while parsing or validating the token, or if
+   *     the user is disabled.
    */
   public FirebaseToken verifyIdToken(@NonNull String idToken, boolean checkRevoked)
       throws FirebaseAuthException {
@@ -343,8 +345,11 @@ public abstract class AbstractFirebaseAuth {
    * checkRevoked} is true, throws a {@link FirebaseAuthException}.
    *
    * @param cookie A Firebase session cookie string to verify and parse.
-   * @param checkRevoked A boolean indicating whether to check if the cookie was explicitly revoked.
+   * @param checkRevoked A boolean indicating whether to check if the cookie was explicitly revoked
+   *                    or if the user is disabled.
    * @return A {@link FirebaseToken} representing the verified and decoded cookie.
+   * @throws FirebaseAuthException If an error occurs while parsing or validating the token, or if
+   *     the user is disabled.
    */
   public FirebaseToken verifySessionCookie(String cookie, boolean checkRevoked)
       throws FirebaseAuthException {
--- a/src/main/java/com/google/firebase/auth/RevocationCheckDecorator.java
+++ b/src/main/java/com/google/firebase/auth/RevocationCheckDecorator.java
@@ -51,7 +51,14 @@ class RevocationCheckDecorator implements FirebaseTokenVerifier {
   @Override
   public FirebaseToken verifyToken(String token) throws FirebaseAuthException {
     FirebaseToken firebaseToken = tokenVerifier.verifyToken(token);
-    if (isRevoked(firebaseToken)) {
+    validateRevokedOrDisabled(firebaseToken);
+    return firebaseToken;
+  }
+
+  private void validateRevokedOrDisabled(FirebaseToken firebaseToken) throws FirebaseAuthException {
+    UserRecord user = userManager.getUserById(firebaseToken.getUid());
+    long issuedAtInSeconds = (long) firebaseToken.getClaims().get("iat");
+    if (user.getTokensValidAfterTimestamp() > issuedAtInSeconds * 1000) {
       throw new FirebaseAuthException(
           ErrorCode.INVALID_ARGUMENT,
           "Firebase " + shortName + " is revoked.",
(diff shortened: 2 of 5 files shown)


src/main/java/com/google/firebase/auth/RevocationCheckDecorator.java
```diff
@@ -51,22 +51,28 @@ private RevocationCheckDecorator(
...
     }
-
-    return firebaseToken;
-  }
-
-  private boolean isRevoked(FirebaseToken firebaseToken) throws FirebaseAuthException {
-    UserRecord user = userManager.getUserById(firebaseToken.getUid());
-    long issuedAtInSeconds = (long) firebaseToken.getClaims().get("iat");
-    return user.getTokensValidAfterTimestamp() > issuedAtInSeconds * 1000;
+    if (user.isDisabled()) {
```
hiranya911, line 69: Should we check for this first? That's how we implemented this in Python.

src/test/java/com/google/firebase/auth/FirebaseAuthIT.java
```diff
@@ -625,6 +625,45 @@ public void testVerifyIdToken() throws Exception {
...
+    // USER_DISABLED.
+    auth.revokeRefreshTokens("disabledUser");
+    try {
+      auth.verifyIdToken(idToken, /* checkRevoked= */ true);
+      fail("Should throw an exception as the ID tokens are revoked.");
+    } catch (FirebaseAuthException e) {
+      assertEquals(AuthErrorCode.REVOKED_ID_TOKEN, e.getAuthErrorCode());
+    }
+
+    auth.deleteUser("disabledUser");
```
hiranya911, line 664: Instead, do this somewhere earlier in the test (shortly after sign in):

```
temporaryUser.registerUid("disabledUser");
```

That will make sure the user account gets deleted after the test, regardless of the outcome of the test.

src/test/java/com/google/firebase/auth/FirebaseAuthIT.java
```diff
@@ -661,6 +700,51 @@ public void testVerifySessionCookie() throws Exception {
...
+    // USER_DISABLED.
+    auth.revokeRefreshTokens("disabledUser2");
+    try {
+      auth.verifySessionCookie(sessionCookie, /* checkRevoked= */ true);
+      fail("Should throw an exception as the tokens are revoked.");
+    } catch (FirebaseAuthException e) {
+      assertEquals(AuthErrorCode.REVOKED_SESSION_COOKIE, e.getAuthErrorCode());
+    }
+
+    auth.deleteUser("disabledUser2");
```
hiranya911, line 745: Same here. Tie it to the existing auto delete mechanism.

src/test/java/com/google/firebase/auth/FirebaseAuthIT.java
```diff
@@ -625,6 +625,45 @@ public void testVerifyIdToken() throws Exception {
     auth.deleteUserAsync("user2");
   }
 
+  @Test
+  public void testVerifyIdTokenUserDisabled() throws Exception {
+    String customToken = auth.createCustomToken("disabledUser");
```
hiranya911, line 630: Nit: It's often safer to use random UID strings instead of static ones (in case 2 developers were running the tests in parallel against a shared project). You can use this API to generate random UIDs:

```java
RandomUser randomUser = UserTestUtils.generateRandomUserInfo();
String customToken = auth.createCustomToken(randomUser.getUid());
```

src/test/java/com/google/firebase/auth/FirebaseAuthTest.java
```diff
@@ -573,6 +624,25 @@ private FirebaseApp getFirebaseAppForUserRetrieval() {
         .build());
   }
 
+  private FirebaseApp getFirebaseAppForDisabledUserRetrieval() throws IOException {
+    String getUserResponse = TestUtils.loadResource("getUser.json");
+    JsonParser parser = ApiClientUtils.getDefaultJsonFactory().createJsonParser(getUserResponse);
+    GenericJson json =
+        parser.parseAndClose(GenericJson.class);
+    ArrayMap<String, Object> users =
```
hiranya911, line 632: Cast to interface types?

```
Map<String, Object> users = ((List<Map<String, Object>>) json.get("users")).get(0);
```

hiranya911 (changes requested):
Looks pretty good. Just a few minor nits pointed out.

src/main/java/com/google/firebase/auth/RevocationCheckDecorator.java
lsirac: Done. Updated tests to reflect this.

hiranya911 (approved):
Thanks. LGTM 👍

hiranya911:
@egilmorez PTAL the javadoc updates in `AbstractFirebaseAuth` and `AuthErrorCode`.

src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
```diff
@@ -256,11 +256,13 @@ public FirebaseToken verifyIdToken(@NonNull String idToken) throws FirebaseAuthE
    * API call.
    *
    * @param idToken A Firebase ID token string to parse and verify.
-   * @param checkRevoked A boolean denoting whether to check if the tokens were revoked.
+   * @param checkRevoked A boolean denoting whether to check if the tokens were revoked or if
```
egilmorez, line 259: This doesn't seem 100% clear, maybe just to me.

If it's a true/false boolean, then is it token is revoked/token is not revoked, or is it user is disabled/user is not disabled?  Can it be both somehow?

src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
```diff
@@ -343,8 +345,11 @@ public FirebaseToken verifySessionCookie(String cookie) throws FirebaseAuthExcep
    * checkRevoked} is true, throws a {@link FirebaseAuthException}.
    *
    * @param cookie A Firebase session cookie string to verify and parse.
-   * @param checkRevoked A boolean indicating whether to check if the cookie was explicitly revoked.
+   * @param checkRevoked A boolean indicating whether to check if the cookie was explicitly revoked
```
egilmorez, line 348: Same question/comment here as for token revoked.  Maybe "if the user is disabled" doesn't belong in either boolean's description?

egilmorez (commented):
LG, though a couple of comments don't fully make sense to me -- possibly my problem but please clarify if you can, thanks!

src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
lsirac: If the developer wants to check if the token is valid, we also need to check if the user is disabled. The user can be disabled but have an unexpired ID token.

Not sure if that answers your question.

src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
lsirac: I think it belongs here. I'm not quite sure what you mean, maybe my comment above clarifies things?

src/main/java/com/google/firebase/auth/AbstractFirebaseAuth.java
egilmorez: OK got it, thanks!

egilmorez (approved):
Thanks!

PR #587, opened 2021-08-13
Update custom Doclet configuration for javadoc

lahirumaramba:
- Starting with version 3.0 `javadoc` uses [updated parameter names](https://maven.apache.org/plugins/maven-javadoc-plugin/examples/alternate-doclet.html) in custom Doclet configurations, which affects our internal devsite scripts.
-  This PR updates the config to use the new parameters names.
- Tested and verified with the devsite script.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -78,16 +78,6 @@
 
     <profiles>
         <profile>
-            <!-- Disable Doclint on Java 8 and higher -->
-            <id>disable-java8-doclint</id>
-            <activation>
-                <jdk>[1.8,)</jdk>
-            </activation>
-            <properties>
-                <additionalparam>-Xdoclint:none</additionalparam>
-            </properties>
-        </profile>
-        <profile>
             <id>devsite-apidocs</id>
             <activation>
                 <property>
--- a/pom.xml
+++ b/pom.xml
@@ -125,6 +125,7 @@
                             </additionalOptions>
                             <useStandardDocletOptions>false</useStandardDocletOptions>
                             <additionalJOption>-J-Xmx1024m</additionalJOption>
+                            <doclint>none</doclint>
                         </configuration>
                     </plugin>
                     <plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -123,16 +123,16 @@
                                     <version>1.3</version>
                                 </additionalDependency>
                             </additionalDependencies>
-                            <additionalparam>
-                                -hdf book.path /docs/reference/_book.yaml
-                                -hdf project.path /_project.yaml
-                                -hdf devsite.path /docs/reference/admin/java/reference/
-                                -d ${project.build.directory}/apidocs
-                                -templatedir ${devsite.template}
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -123,16 +123,16 @@
...
-                                -hdf book.path /docs/reference/_book.yaml
-                                -hdf project.path /_project.yaml
-                                -hdf devsite.path /docs/reference/admin/java/reference/
-                                -d ${project.build.directory}/apidocs
-                                -templatedir ${devsite.template}
-                                -toroot /docs/reference/admin/java/reference/
-                                -yaml _toc.yaml
-                                -warning 101
-                            </additionalparam>
+                            <additionalOptions>
```
hiranya911, line 126: I think there are couple of other places in this file that use the old syntax. Can we change them too?

hiranya911 (changes requested):
Thanks for putting this together. Just one suggestion.

pom.xml
lahirumaramba: Wasn't sure if the rest should be updated as well. Thanks! I will update all the instances.

lahirumaramba:
Looks like line 87 still uses the current config name. Updated the rest. Thanks!

hiranya911:
Thanks. May be we can remove that profile entirely, and just merge the configuration with the rest? It's reasonable to require jdk 1.8 or higher for all build and release tasks:

```xml
       <profile>
            <!-- Disable Doclint on Java 8 and higher -->
            <id>disable-java8-doclint</id>
            <activation>
                <jdk>[1.8,)</jdk>
            </activation>
            <properties>
                <additionalparam>-Xdoclint:none</additionalparam>
            </properties>
        </profile>
```

pom.xml
```diff
@@ -123,16 +113,16 @@
...
+                                <additionalOption>-hdf book.path /docs/reference/_book.yaml</additionalOption>
```
hiranya911, line 117: We need to pass the `-Xdoclint:none` option here too. Earlier it was provided by the profile whenever JDK 1.8 was available. Now we have to explicitly set it.

hiranya911 (changes requested):

pom.xml
lahirumaramba: Thanks! Added `<doclint>none</doclint>`

pom.xml
```diff
@@ -334,9 +325,9 @@
                             <version>1.3</version>
                         </additionalDependency>
                     </additionalDependencies>
-                    <additionalparam>
-                        -warning 101
-                    </additionalparam>
+                    <additionalOptions>
+                        <additionalOption>-warning 101</additionalOption>
+                    </additionalOptions>
```
hiranya911, line 330: Probably need doclint:none here too.

hiranya911 (approved):
LGTM with a suggestion.

pom.xml
lahirumaramba: Updated. Thanks!

PR #672, opened 2022-05-10
feat(appcheck): Add initial classes for App Check

lahirumaramba:
- Add `verifyToken()` API

### DIFF
--- a/src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
+++ b/src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.firebase.appcheck;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+import com.google.common.collect.ImmutableMap;
+
+import java.util.Map;
+
+/**
+ * A decoded and verified Firebase App Check token. See {@link FirebaseAppCheck#verifyToken(String)}
+ * for details on how to obtain an instance of this class.
+ */
+public final class DecodedAppCheckToken {
+
+  private final Map<String, Object> claims;
+
+  DecodedAppCheckToken(Map<String, Object> claims) {
+    checkArgument(claims != null && claims.containsKey("sub"),
+            "Claims map must at least contain sub");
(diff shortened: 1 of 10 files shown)


src/main/java/com/google/firebase/appcheck/FirebaseAppCheckClientImpl.java
```diff
@@ -0,0 +1,90 @@
...
+  static FirebaseAppCheckClientImpl fromApp(FirebaseApp app) {
+    String projectId = ImplFirebaseTrampolines.getProjectId(app);
+    checkArgument(!Strings.isNullOrEmpty(projectId),
+            "Project ID is required to access App Check service. Use a service "
+                    + "account credential or set the project ID explicitly via FirebaseOptions. "
+                    + "Alternatively you can also set the project ID via the GOOGLE_CLOUD_PROJECT "
+                    + "environment variable.");
+    AppCheckTokenVerifier appCheckTokenVerifier = AppCheckTokenVerifier.builder()
+            .setProjectId(projectId)
+            .build();
```
lahirumaramba, line 46: I think we can use a token factory to clean this up a bit more later. That way we can also configure the JWKS url, cache duration, etc.

src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
```diff
@@ -0,0 +1,68 @@
...
+
+import com.google.common.collect.ImmutableMap;
+
+import java.util.Map;
+
+/**
+ * A decoded and verified Firebase App Check token. See {@link FirebaseAppCheck#verifyToken(String)}
+ * for details on how to obtain an instance of this class.
+ */
+public final class DecodedAppCheckToken {
```
weixifan, line 29: I almost wonder whether we should be exposing all the contents of the token or offer our own opinionated data layer. For example, we can offer just one method `getAppId()` for now like our Node.js (I think).

bhackstock:
Is there an estimate for when this logic will be available in a release ❓🛳️

We are using a java-only backend and would like to verify `app check` tokens for our backend API calls!
If there is anything I can help with or contribute something myself, I'd be glad to 😊

(don't know if this the right place to post/ask about this, please advise if that's not the case 👀)

src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
```diff
@@ -0,0 +1,68 @@
...
+    return (String) claims.get("aud");
+  }
+
+  /** Returns the Expiration Time for this token. */
+  public String getExpirationTime() {
+    return (String) claims.get("exp");
+  }
+
+  /** Returns the Issued At for this token. */
+  public String getIssuedAt() {
```
kaibolay, line 60: ditto. Ideally that'd be `java.time.Instant`, but barring that this should return an integer like `getExpirationTime()`

bard-kristian:
Is this being worked on?

src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
lahirumaramba: Good call. Change to `Long` here and above. Thanks!

tjarvstrand:
Any progress on this? It's kind of strange that AppCheck is not part of the SDK API after such a long time...

alaegin:
Hi, is there any chance that it will be implemented?

dohyungp:
@lahirumaramba Do you have any release plans?

PR #734, opened 2022-11-04
fix(auth): Set emulator credentials when using the Auth emulator

lahirumaramba:
If `FIREBASE_AUTH_EMULATOR_HOST` env var is set the SDK shouldn't connect to Auth servers for OAuth2 handshake. This PR sets the emulator credentials when emulator mode is set.

b/220373445


lahirumaramba:
This change could break Firestore if the `FIREBASE_AUTH_EMULATOR_HOST` is set... changing to draft until I take a closer look.

christhompsongoogle (approved):

kroikie (approved):
LGTM

PR #747, opened 2022-12-08
Fix a small bug in setCustomUserClaims() auth snippet

NhienLam:
I was testing the code snippets in FirebaseAuthSnippets.java, which would be used this doc https://cloud.google.com/identity-platform/docs/multi-tenancy-managing-tenants#controlling_access_with_custom_claims. When testing the setCustomUserClaims(), I got an USER_NOT_FOUND error. 
"No user record found for the given identifier (USER_NOT_FOUND)."

I believe this error happens because `FirebaseAuth.getInstance()` doesn’t specify which tenant we are setting.
Therefore, using TenantAwareAuth instance for the corresponding tenant `tenantAuth` instead of `FirebaseAuth.getInstance()` fixes the issue.

### DIFF
--- a/src/test/java/com/google/firebase/snippets/FirebaseAuthSnippets.java
+++ b/src/test/java/com/google/firebase/snippets/FirebaseAuthSnippets.java
@@ -151,13 +151,13 @@ public class FirebaseAuthSnippets {
     // [END update_user]
   }
 
-  public static void setCustomUserClaims(TenantAwareFirebaseAuth tenantAuth,
+  public static void setCustomUserClaims(
       String uid) throws FirebaseAuthException {
     // [START set_custom_user_claims]
     // Set admin privilege on the user corresponding to uid.
     Map<String, Object> claims = new HashMap<>();
     claims.put("admin", true);
-    tenantAuth.setCustomUserClaims(uid, claims);
+    FirebaseAuth.getInstance().setCustomUserClaims(uid, claims);
     // The new custom claims will propagate to the user's ID token the
     // next time a new one is issued.
     // [END set_custom_user_claims]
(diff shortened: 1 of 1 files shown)


src/test/java/com/google/firebase/snippets/FirebaseAuthSnippets.java
```diff
@@ -151,13 +151,13 @@ public static void updateUser(String uid) throws FirebaseAuthException {
     // [END update_user]
   }
 
-  public static void setCustomUserClaims(
+  public static void setCustomUserClaims(TenantAwareFirebaseAuth tenantAuth,
```
prameshj, line 154: I think it would be better to add a new function like "setCustomUserClaimsTenant" in https://github.com/firebase/firebase-admin-java/blob/1eec932b27276ddf1d6ab491a08edb68a3036170/src/test/java/com/google/firebase/snippets/FirebaseAuthSnippets.java#L1152

NhienLam: Agreed. 
I reverted the changes made in `setCustomUsersClaims` and created a new function `setCustomUserClaimsTenant` that sets custom claims on a user for a specific tenant.

prameshj (approved):

pragatimodi (approved):

lahirumaramba (approved):
Thanks @NhienLam ! LGTM!

lahirumaramba:
@prameshj @kevinthecheung Should we also add Java code samples to https://cloud.google.com/identity-platform/docs/multi-tenancy-managing-tenants#controlling_access_with_custom_claims ?

NhienLam:
@lahirumaramba Thanks for approving the PR! 

> Should we also add Java code samples to https://cloud.google.com/identity-platform/docs/multi-tenancy-managing-tenants#controlling_access_with_custom_claims ?
There is a CL that adds Java code samples to this doc. I was testing these code snippets while reviewing that CL. I will comment on that CL to tell the author to use the new setCustomUserClaimsTenant() which I added. Hopefully, that CL will be merged and the doc will be updated soon.

PR #775, opened 2023-02-22
Update passwordHash and passwordSalt to be base64 decoded before user import

NhienLam:
**Issue**: Using the [importWithScrypt()](https://github.com/firebase/firebase-admin-java/blob/master/src/test/java/com/google/firebase/snippets/FirebaseAuthSnippets.java#L566) code snippet in the [public doc](https://firebase.google.com/docs/auth/admin/import-users#import_users_with_firebase_scrypt_hashed_passwords) will successfully import the user, but we will be unable to sign in because the password is changed.

This may be similar to this issue in Firebase Admin Go: [https://github.com/firebase/firebase-admin-go/issues/479](https://www.google.com/url?q=https://github.com/firebase/firebase-admin-go/issues/479&sa=D&source=buganizer&usg=AOvVaw1NQR0Ya_HBNq6nQolmle6w) and the fixes in [https://github.com/firebase/firebase-admin-go/pull/480](https://www.google.com/url?q=https://github.com/firebase/firebase-admin-go/pull/480&sa=D&source=buganizer&usg=AOvVaw3HuLfgng7APuSd_ZPY1jTH) and [https://github.com/firebase/firebase-admin-python/pull/652](https://www.google.com/url?q=https://github.com/firebase/firebase-admin-python/pull/652&sa=D&source=buganizer&usg=AOvVaw3-Hc9nmXTIZ7npaMro5FAK)

After this change, we can successfully import a user and sign in with the original password.

**Note**: I use `BaseEncoding.base64()` instead of `BaseEncoding.base64Url()` because `BaseEncoding.base64Url().decode` will throw an `Unrecognized character: +` exception if the password hash contains '+'.

### DIFF
--- a/src/test/java/com/google/firebase/snippets/FirebaseAuthSnippets.java
+++ b/src/test/java/com/google/firebase/snippets/FirebaseAuthSnippets.java
@@ -569,8 +569,8 @@ public class FirebaseAuthSnippets {
       List<ImportUserRecord> users = Collections.singletonList(ImportUserRecord.builder()
           .setUid("some-uid")
           .setEmail("user@example.com")
-          .setPasswordHash("password-hash".getBytes())
-          .setPasswordSalt("salt".getBytes())
+          .setPasswordHash(BaseEncoding.base64().decode("password-hash"))
+          .setPasswordSalt(BaseEncoding.base64().decode("salt"))
           .build());
       UserImportOptions options = UserImportOptions.withHash(
           Scrypt.builder()


lahirumaramba (approved):
Thank you @NhienLam ! LGTM!

Do you think we should update other import snippets, too? `importWithBcrypt` etc.

NhienLam:
> Do you think we should update other import snippets, too? importWithBcrypt etc.

Thanks @lahirumaramba! I think the other import snippets don't need to be changed because they are using other hashing algorithms. Quote from https://github.com/firebase/firebase-admin-python/pull/652: "It is fair to assume that the password hash and salt from those other backends has been decoded properly."

PR #785, opened 2023-03-17
feat(fcm): Implement `sendEach`, `sendEachAsync`, `sendEachForMulticast` and `sendEachForMulticastAsync`

Doris-Ge:
1. Add `sendEach`, `sendEachAsync`, `sendEachForMulticast` and `sendEachForMulticastAsync`
2. Deprecate `sendAll`, `sendAllAsync`, `sendMulticast` and `sendMulticastAsync`

`sendEach` vs `sendAll`

 1. `sendEach` sends one HTTP request to V1 Send endpoint for each
      message in the array.
     `sendAll` sends only one HTTP request to V1 Batch Send endpoint
      to send all messages in the array.
 3. `sendEach` calls `messagingClient.send` to send each message
      and constructs a `SendResponse` with the returned `messageId`.
      If `messagingClient.send` throws out an exception, `sendEach`
      will catch the exception and also turn it into a `SendResponse`
      with the exception in it.
      `sendEach` calls `ApiFutures.allAsList().get()` to execute all
      `messagingClient.send` calls asynchronously and wait for all of
      them to complete and construct a `BatchResponse` with all
      `SendResponse`s.
      Therefore, unlike `sendAll`, `sendEach` does not always throw
      an error for a total failure. It can also return a `BatchResponse`
      with only errors in it.  

`sendEachForMulticast` calls `sendEach` under the hood.
`sendEachAsync` is the async version of `sendEach`.
`sendEachForMulticastAsync` is the async version of `sendEachForMulticast`.

Will send another PR for integration tests

### DIFF
--- a/src/main/java/com/google/firebase/messaging/FirebaseMessaging.java
+++ b/src/main/java/com/google/firebase/messaging/FirebaseMessaging.java
@@ -20,18 +20,22 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.api.core.ApiFuture;
+import com.google.api.core.ApiFutures;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
 import com.google.common.base.Supplier;
 import com.google.common.base.Suppliers;
 import com.google.common.collect.ImmutableList;
+import com.google.firebase.ErrorCode;
 import com.google.firebase.FirebaseApp;
 import com.google.firebase.ImplFirebaseTrampolines;
 import com.google.firebase.internal.CallableOperation;
 import com.google.firebase.internal.FirebaseService;
 import com.google.firebase.internal.NonNull;
 
+import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.ExecutionException;
 
 /**
  * This class is the entry point for all server-side Firebase Cloud Messaging actions.
@@ -91,7 +95,7 @@ public class FirebaseMessaging {
    *
    * <p>If the {@code dryRun} option is set to true, the message will not be actually sent. Instead
    * FCM performs all the necessary validations and emulates the send operation. The {@code dryRun}
-   * option is useful for determining whether an FCM registration has been deleted. However, it 
+   * option is useful for determining whether an FCM registration has been deleted. However, it
    * cannot be used to validate APNs tokens.
    *
    * @param message A non-null {@link Message} to be sent.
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/firebase/messaging/FirebaseMessaging.java
```diff
@@ -139,6 +143,191 @@ protected String execute() throws FirebaseMessagingException {
...
+        List<ApiFuture<SendResponse>> list = new ArrayList<>();
+        for (Message message : immutableMessages) {
+          ApiFuture<SendResponse> messageId = sendOpForSendResponse(message, dryRun).callAsync(app);
+          list.add(messageId);
+        }
+        try {
+          List<SendResponse> responses = ApiFutures.allAsList(list).get();
+          return new BatchResponseImpl(responses);
+        } catch (InterruptedException | ExecutionException e) {
+          throw new FirebaseMessagingException(ErrorCode.CANCELLED, SERVICE_ID);
```
Doris-Ge, line 234: We can discuss what error code we want to throw here.

Doris-Ge: Also, we can discuss whether we should make an effort to add a unit test for this case

lahirumaramba: If one message fails to go through would this throw immediately cancelling the rest of the messages?

lahirumaramba (approved):
LGTM! Thank you!!

src/main/java/com/google/firebase/messaging/FirebaseMessaging.java
Doris-Ge: If one message fails because of `FirebaseMesssagingException`, then it won't cancel the rest of the messages because the exception will be caught in `sendOpForSendResponse`.

Based on my read of https://dzone.com/articles/google-guava-%E2%80%93-futures, `allAsList` is fail-fast.
So if one message fails because the `execute()` throws an exception other than `FirebaseMesssagingException`,
```
      protected SendResponse execute() {
        try {
          String messageId = messagingClient.send(message, dryRun);
          return SendResponse.fromMessageId(messageId);
        } catch (FirebaseMessagingException e) {
          return SendResponse.fromException(e);
        }
      }
```
then the rest of the messages will be cancelled. I'm not sure in what cases this could happen, but I think it should be a corner case.

Doris-Ge: I'm going to merge this PR. Let me know if you have any suggested changes for this. Ill fix it in another PR.

sarismet:
Hi @lahirumaramba @Doris-Ge,

Could you please tell a little bit why you deprecated sendAll and recommend us to use sendEach instead? What was the motivation behind it. I know sendEach looks more reliable to debug failed attempts, however, even if it can send http requests async meaning that it may perform as good as sendAll, does it require a single thread from the thread-pool and cause more memory-usage, more networking and less available threads in pool? Therefore, the sendAll method looks more efficient to use threads to me. We are sending millions of notifications in a couple of minutes and we consider to lower our memory usage. Thanks in advance.

Doris-Ge:
Hi @sarismet,
 
> On June 20, 2024, we’re reducing the number of Firebase Cloud Messaging (FCM) legacy register APIs and legacy send APIs that provide similar functionality. This step will allow us to provide you with a more consistent experience and align with Google security standards to improve security, reliability and performance.
> 
> Because of these API decommissions, some already-deprecated SDKs and features will stop working after June 20, 2024.
> 
> Please consult the tables below to find which Firebase Cloud Messaging (FCM) APIs and corresponding 
image

This is because `sendAll` calls the FCM Batch send API` which is deprecated and will be discontinued on June 20, 2024.  
services/SDKs/features will be discontinued and replaced with new alternatives.

> even if it can send http requests async meaning that it may perform as good as sendAll, does it require a single thread from the thread-pool and cause more memory-usage, more networking and less available threads in pool?

I'm not sure whether it requires a single thread to send one async request. For lowering memory usage, I'd suggest to reduce the number of messages to send per `sendEach` call.

amitkumar43:
Hi @sarismet 
What are your learning after moving to sendAll APIs and any optimizations you did for making your system performant at this scale with sendAll APIs.
We are calling sendAll API with a list size of 100, and seeing some gaps around performance.

sarismet:
Hi @amitkumar43 

We overwrite ThreadManager with custom thread executor. It helps a lot. Before that our Kafka consumer gets into rebalancing. 

Nowadays, we are trying to use virtualThreadExecutors in ThreadManager. We reach 2 millions request per minutes but we got errors saying that "Error writing request body to server", "Remote host terminated the handshake", "Unexpected end of file from server". When load is decreased, the error counts decreases as well.

Do you have any idea what causes these errors? Have you ever tried virtual threads in that project? @Doris-Ge @lahirumaramba 

Our java docker image: eclipse-temurin:21.0.3_9-jre-alpine

PR #836, opened 2023-06-23
Mark `sendAll` and `sendMulticast` variants as deprecated

vpavic:
In d1baee36 all `sendAll` and `sendMulticast` variants were deprecated, however those methods were marked as deprecated only in javadoc but not using `@Deprecated`.

This commit adds `@Deprecated` annotation where needed, in order to better signal the deprecation to developers.


Doris-Ge (approved):
Thank you for the PR!

Issue #850, opened 2023-08-07
HttpRequest with 0 write timeout may make program hangs


meor9zcu12:
### Problem encountered:
When under high traffic, FCM admin SDK will throw some errors like:

> Error 1. com.google.firebase.messaging.FirebaseMessagingException: Unexpected HTTP response with status: 408

> Error 2. com.google.firebase.messaging.FirebaseMessagingException: Unknown error while making a remote service call: Unexpected end of file from server

> Error 3. com.google.firebase.messaging.FirebaseMessagingException: Unknown error while making a remote service call: Connection timed out (Connection timed out)

At the same time, after throwing these errors, FCM admin SDK may still make java program never stop because of FCM admin SDK still waiting response from [fcm.googleapis.com](fcm.googleapis.com)

### Findings:
FCM admin uses com.google.http-client, which writeTimeout = 0 by default, [link](https://github.com/googleapis/google-http-java-client/blob/13edd1357eb79535d943f5288ab3adf4d9dfb52a/google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java#L154)

I can't find how [ErrorHandlingHttpClient ](https://github.com/firebase/firebase-admin-java/blob/master/src/main/java/com/google/firebase/internal/ErrorHandlingHttpClient.java#L92) to set writeTimeout, it may be one of result that make program keep waiting response from [fcm.googleapis.com](fcm.googleapis.com) under high traffic.

TheRittler:
I think we also ran into this Issue. Are there any news on this topic?

chong-shao:
Sorry for the delay in reply. We are currently investigating this issue. Will update here later.

GarreltMock:
Any updates here?

PR #900, opened 2024-01-10
fix: Make writeTimeout configurable (similar to readTimeout)

TheRittler:
Solves #850

### DIFF
--- a/src/main/java/com/google/firebase/FirebaseOptions.java
+++ b/src/main/java/com/google/firebase/FirebaseOptions.java
@@ -214,7 +214,7 @@ public final class FirebaseOptions {
    * Returns the write timeout in milliseconds, which is applied to outgoing REST calls
    * made by the SDK.
    *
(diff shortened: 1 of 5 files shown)


lahirumaramba (approved):
LGTM! Thank you for the contribution!

src/main/java/com/google/firebase/FirebaseOptions.java
```diff
@@ -207,6 +210,16 @@ public int getReadTimeout() {
     return readTimeout;
   }
 
+  /**
+   * Returns the write timeout in milliseconds, which is applied to outgoing REST calls
```
egilmorez, line 214: Since this is a case for "that" instead of which, let's just rephrase to something like:

"Returns the write timeout applied to outgoing REST calls in milliseconds."

If you think we can go ahead and assume "by the SDK."

egilmorez (approved):
LG with a nit, thanks!

Issue #802, opened 2023-04-24
[FR] - Need a setter to set FCM host for Functional tests/Service Level tests.


yashjain12yj:
**Summary**:
I recently integrated FCM SDK to send Push Notifications. The integration was very simple and smooth but while I started writing Functional Tests, I got to know that I can't give FCM host while constructing a `FcmMessaging` object. The `FCM_URL` is hardcoded in the `FirebaseMessagingClientImpl` class, which is restricting me from writing functional tests.

**Solution**:
While constructing the `FcmMessaging` object or while initializing `FirebaseApp` there can be an option to pass the FCM host. If the host is not passed then the default hardcoded host must be used.

**Describe alternatives you've considered**
Currently, due to this, I'm blocked from writing FTs. I'm currently only dependent on UTs

bashbers:
Hey @yashjain12yj, might be too late for you since it's currently november but I implemented a work-around for this issue using reflection.

```java
/**
   * Override the internal fcm.googleapis.com URL with a gateway URL. This function is needed due to
   * the FirebaseClient interface missing a property to configure the FCM url.
   *
   * @throws Exception ReflectionBased errors, could be encountered once the {@link FirebaseClient}
   *     interface is changed.
   */
  private void overrideFcmUrl() throws Exception { // NOSONAR
    var instance = FirebaseMessaging.getInstance();
    var clientSupplierField = instance.getClass().getDeclaredField("messagingClient"); // NOSONAR
    clientSupplierField.setAccessible(true); // NOSONAR
    var clientSupplier = (Supplier<?>) clientSupplierField.get(instance); // NOSONAR
    var messager = clientSupplier.get(); // NOSONAR
    var urlField = messager.getClass().getDeclaredField("fcmSendUrl"); // NOSONAR
    setFinal(messager, urlField, gatewaySendUrl); // NOSONAR
  }

  /**
   * Helper function to override private final fields in Objects.
   *
   * @param instance Object that contains {@link Field}
   * @param field Field that should be changed
   * @param newValue any value
   * @throws Exception ReflectionBased errors, could be encountered once the {@link FirebaseClient}
   *     interface is changed
   */
  private void setFinal(Object instance, Field field, Object newValue) throws Exception { // NOSONAR
    field.setAccessible(true); // NOSONAR
    Field modifiersField = Field.class.getDeclaredField("modifiers"); // NOSONAR
    modifiersField.setAccessible(true); // NOSONAR
[...]

PR #912, opened 2024-02-13
Get FCM Root Url from Environment variable

yashjain12yj:
### Summary
Added functionality to override FCM root URL via Env Variable so that before constructing the FirebaseApp object we can set the Fcm root url in the environment variable. This will help in changing the base URL to localhost while writing FTs(Functional Tests)/Integration Tests.

**Global Variable**: `GOOGLE_FCM_HOST` 

**Open Issue**: [#802](https://github.com/firebase/firebase-admin-java/issues/802)

**How setting the base url to localhost help with your functional tests?**
While writing functional tests/Integration tests, we do mock all the external services and create mock servers which behaves the same as external service. So here in case of sending push notifications, we are using Firebase admin SDK and to write tests for this use-case, we would need to mock the FCM server with a mock server created in tests.

**For example**
image

Here, Suppose we are writing tests for Service 1. It exposes a GET API to get some data. The data contains multiple attributes few attributes from DB, a few from the cache, and a few from Service 2(HTTP API).

While setting up the tests we will create a DB, cache server, and a mock HTTP server. Once the server is up, we want Service 1 to point to these servers instead of actual hosts. Since the DB, cache, and mock HTTP server are running in the same machine on different ports. We would want to set the Host to localhost with respective ports.

Now in the case of Firebase admin SDK, the host is hardcoded in the library. Hence, we are not able to override the Host to localhost.


yashjain12yj:
@kroikie Could you please review the PR?

ajax-semenov-y:
@kroikie @lahirumaramba, these changes will increase the testability of applications that use `firebase-admin-java`. My team is also currently blocked. I will be grateful for accepting this contribution from @yashjain12yj.

ajax-semenov-y (approved):

aditya-gupta-swiggy (approved):
thanks for raising this PR @yashjain12yj ! this is much needed for our use case as well, grateful for your contribution. can we merge this @kroikie @lahirumaramba ?

ajax-semenov-y:
@kroikie @lahirumaramba any updates on this?

mygreen:
Excellent PR !

Furthermore, It would be better if it is possible to change the value of IID_HOST defined in [InstanceIdClientImpl](https://github.com/firebase/firebase-admin-java/blob/master/src/main/java/com/google/firebase/messaging/InstanceIdClientImpl.java#L43).

InstanceIdClientImpl is used when subscribing to a pre-work topic that sends push notifications to the topic.
ex) `FirebaseMessaging.getInstance().subscribeToTopic("topic", List.of("token"))`

rashineema:
This change will significantly improve the testability of the application. By using the admin SDK, we avoid the need to implement the HTTP v1 client and manage FCM-specific errors ourselves.

@kroikie @lahirumaramba, could you please review and merge this? It will be beneficial for everyone.

rashineema (approved):

martinboue:
Thanks for the contribution. Any update on this PR? This is also something we need for testing purposes.

ajax-semenov-y (commented):
Sorry, that was just a misclick.

tdfeslo:
Is there any update? Is it possible to take this change into account to override the host "https://fcm.googleapis.com" either via an environment variable or an option in FirebaseOptions?

Thank you in advance.

Issue #834, opened 2023-06-20
FCM Batch sending deprecated but no migration path


19lmyers:
* Library version: 9.1.1
  * Firebase Product: Messaging

I received an email from Firebase about the deprecation of legacy FCM APIs. Notably, the `sendAll()` method that employs batch sending has been deprecated, and I saw the Node Admin SDK has a new `sendEach()` method, but the Java SDK does not provide this method.

How should I approach migrating from the `sendAll()` method?

AndrWeisR:
Ditto. I have firebase-admin-9.1.1.jar which only exposes these methods:

![image](https://github.com/firebase/firebase-admin-java/assets/34931916/0c040fc6-2081-4371-a938-b6219aff81f9)

adhirajsinghchauhan:
> The question is: when 9.2.0 will be released?

@xak2000 v9.2.0 was released [just a day after your comment][release-9.2.0].

---
But I wonder if it'd be feasible to walk back the decision to discontinue the batch send API, or at least move it into the Blaze plan? For people who want a fast, fire-and-forget delivery system, `sendEach` does not fit the bill. Both theoretically and in real-world tests, `sendAll` far outperforms its replacement.

Quick tests with 2000 messages (each to a different topic, but with the same config otherwise) on a few of my own servers. No code changes apart from calling a different function, no retry or delay mechanisms in between, apart from what the library itself defaults to:
```yaml
 sendAll: 5-10s
sendEach: 1-5m
```

The overhead of 2000 HTTP calls instead of just 4 has obvious effects. This only becomes worse as the number of messages, meant to be sent approximately "together", increases.

[release-9.2.0]: https://github.com/firebase/firebase-admin-java/releases/tag/v9.2.0

xak2000:
In the [FAQ](https://firebase.google.com/support/faq#fcm-depr-multiple-tokens) there is a dedicated topic about sending message to multiple tokens in one request.

> Q.  Does the HTTP v1 API support sending messages to multiple tokens in one request??
> A. No. This feature, called "multicast" in legacy HTTP APIs, is not supported by the HTTP v1 API.
> 
> For use cases where **end-to-end latency** is critical, or where **total fanout size is small** (fewer than 1 million), Google recommends the HTTP V1 API. The HTTP V1 API over HTTP/2 can achieve up to ~1/3 of total message throughput rate compared to legacy multicast when sending on a single connection.
> 
[...]

PR #979, opened 2024-08-20
feat: Add HTTP/2 enabled transport as default transport 

jonathanedey:
This changes the default `HttpTransport` to an HTTP/2 enabled Apache 5 HttpClient.


emindeniz99:
Thank you for efforts. I think, if a new transport is implemented and used instead of tested built-in google http client, they should be very well tested. we can copy test cases from this not merged PR to this PR in order to test new http2transport layer.

https://github.com/googleapis/google-http-java-client/pull/1960/files#diff-f6e7e17e639b08337930ac900026abfe1531b7c7e9bea3b6866835526ed1fa81

emindeniz99:
It fixes one of the big problem with current state of sdk but it is could be efficient if thread per request structure of library is changed. CallOp causes blocking a thread per request. 

> This is better, it does not open many sockets, etc. HTTP2 is used. It solves our main problem.
But the fcm library creates a thread for each request, for example sending 6 batch requests with each batch consisting of 500 device tokens. 6*500= 3k threads will be created.
As I know, each thread requires 1 MB memory, this is another problem.
Copied from: https://github.com/firebase/firebase-admin-java/issues/834#issuecomment-1890969169

jonathanedey:
Hi @emindeniz99, Thanks for your patience on this!

Went ahead and mirrored most of those tests here.

The issue of threads is something we have been looking into along side http2. The main culprit being the default ThreadExecutor (although it reuses idle threads) has no thread limit which causes more harm than good. We're planning to address this in a separate PR.

Ideally we would want to make full use of Apache HttpClient 5's async clients and use CompletableFutures instead of a thread per request. However from my understanding this requires the google http client to also support async clients or an effort on our side to modify their classes to do so.

This transport in itself is a bandaid solution until async clients required for HTTP/2 are supported in google http client.

honorhs:
Thank you for your efforts in providing the HTTP/2 functionality.
It would be beneficial to have a report comparing the performance of the version with HTTP/2 applied and the previous version. This could include metrics such as system resources and processing completion time, under the assumption of sending a large number of push messages

src/main/java/com/google/firebase/internal/ApacheHttp2AsyncEntityProducer.java
```diff
@@ -0,0 +1,149 @@
...
+import java.nio.ByteBuffer;
+import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.atomic.AtomicReference;
+
+import org.apache.hc.core5.http.ContentType;
+import org.apache.hc.core5.http.nio.AsyncEntityProducer;
+import org.apache.hc.core5.http.nio.DataStreamChannel;
+
+@SuppressWarnings("deprecation")
```
lahirumaramba, line 33: Which deprecation is this? Maybe we should add a note

src/main/java/com/google/firebase/internal/ApacheHttp2AsyncEntityProducer.java
```diff
@@ -0,0 +1,149 @@
...
+  @Override
+  public void releaseResources() {
+    bytebuf.clear();
+  }
+
+  @VisibleForTesting
+  ByteBuffer getBytebuf() {
+    return bytebuf;
+  }
+}
```
lahirumaramba, line 149: nit: new line

src/main/java/com/google/firebase/internal/ApacheHttp2AsyncEntityProducer.java
```diff
@@ -0,0 +1,149 @@
...
+
+  @Override
+  public void produce(DataStreamChannel channel) throws IOException {
+    if (bytebuf == null) {
+      if (content != null) {
+        try {
+          content.writeTo(baos);
+        } catch (IOException e) {
+          writeFuture.completeExceptionally(e);
+          // failed(e);
```
lahirumaramba, line 110: nit: can we remove this now?

src/main/java/com/google/firebase/internal/ApacheHttp2Transport.java
```diff
@@ -0,0 +1,104 @@
...
+  public static CloseableHttpAsyncClient newDefaultHttpAsyncClient() {
+    return defaultHttpAsyncClientBuilder().build();
+  }
+
+  public static HttpAsyncClientBuilder defaultHttpAsyncClientBuilder() {
+    PoolingAsyncClientConnectionManager connectionManager = 
+        new PoolingAsyncClientConnectionManager();
+    connectionManager.setMaxTotal(100);
+    connectionManager.setDefaultMaxPerRoute(100);
+    connectionManager.closeIdle(TimeValue.of(30, TimeUnit.SECONDS));
```
lahirumaramba, line 67: Should we add a note on how we decided on these limits? 100 max connections and 30 seconds timeout?

src/test/java/com/google/firebase/internal/ApacheHttp2TransportIT.java
```diff
@@ -0,0 +1,289 @@
...
+import org.junit.Test;
+
+public class ApacheHttp2TransportIT {
+  private static FirebaseApp app;
+  private static final GoogleCredentials MOCK_CREDENTIALS = new MockGoogleCredentials("test_token");
+  private static final ImmutableMap<String, Object> payload = 
+      ImmutableMap.<String, Object>of("foo", "bar");
+
+  // Sets a 5 second delay before response
+  private static final String DELAY_URL = "https://nghttp2.org/httpbin/delay/5";
```
lahirumaramba, line 60: can we add a note explaining why?

src/main/java/com/google/firebase/internal/ApiClientUtils.java
```diff
@@ -88,6 +88,6 @@ public static JsonFactory getDefaultJsonFactory() {
   }
 
   public static HttpTransport getDefaultTransport() {
-    return Utils.getDefaultTransport();
+    return new ApacheHttp2Transport();
```
lahirumaramba, line 91: Are we dropping the google api client DefaultTransport and moving to the new apache client? This change affects the core SDK right?

diegomarquezp:
> This transport in itself is a bandaid solution until async clients required for HTTP/2 are supported in google http client.

cc: @burkedavison @ldetmer

src/main/java/com/google/firebase/internal/ApacheHttp2AsyncEntityProducer.java
jonathanedey: This is related to the google api client's `StreamingContent`. Removed the suppression to match other files where we don't suppress this warning.

src/main/java/com/google/firebase/internal/ApacheHttp2AsyncEntityProducer.java
jonathanedey: Done!

src/main/java/com/google/firebase/internal/ApacheHttp2AsyncEntityProducer.java
jonathanedey: This should have been uncommented. Fixed that and fixed an issue where exceptions were not being re-thrown.

src/main/java/com/google/firebase/internal/ApacheHttp2Transport.java
jonathanedey: Good point, I changed the settings here to match the google api client's apache http transport and added a comment stating that. 

Also correctly limited the max concurrent streams to 100 and added a comment on the reasoning for this.

Also added more additional comments throughout this file.

src/main/java/com/google/firebase/internal/ApiClientUtils.java
jonathanedey: This doesn't require core SDK changes since we are still using a base `HttpTransport` and developers can still pass in their custom transports instead. However since the underlying transport is changing, there are still minor side effects in how requests are sent under the hood. 

Keys ones being:
- Timeouts
- Low level request exceptions

For non http/2 backend service endpoints, requests will fallback to use http/1. As we discussed, we can decide on whether this should replace the current default transport or simply expose the new http/2 transport for developers to add manually, after we verify proxy functionality.

SmikeSix2: i build this yesterday and we are testing it in production .. :) since we completely overlooked the change.
had conflicts with pushy dependencies, so this was the last option.

im basically using your apachehttp2transport with more connections. any tip on max output settings? the 9.3.0 with sendEachForMulticastAsync required to use semaphores or it crashed our push server.

we are a soccer company and need to send out like ~80k pushes regulary as fasts as possible.
this solution seem to be stable in terms of pushes send & received, but processing seems kinda slow. and its crazy hard to get settings and processing queues correct.

emindeniz99: > i build this yesterday and we are testing it in production .. :) since we completely overlooked the change. had conflicts with pushy dependencies, so this was the last option.
> 
> im basically using your apachehttp2transport with more connections. any tip on max output settings? the 9.3.0 with sendEachForMulticastAsync required to use semaphores or it crashed our push server.
> 
> we are a soccer company and need to send out like ~80k pushes regulary as fasts as possible. this solution seem to be stable in terms of pushes send & received, but processing seems kinda slow. and its crazy hard to get settings and processing queues correct.


you can use current library for creating auth token and body, after that, send via spring webclient or reactor netty. 
https://github.com/firebase/firebase-admin-java/issues/834#issuecomment-2107964352
you check the code i shared on the comment and read the other issue for details.

SmikeSix2: hi emindeniz99 thanks for the reply. i tried it before but run in dependencies nightmare with pushy.

our issue is the delay. it used to be 1-2s till everything was out, now its like 1s - 30s on large loads.
firebase seems to have added some anti burst behavior. with warmups etc. which is fine for normal stuff, but people want to know when a goal was shot asap.

like can we get out 80k~ in a second?

emindeniz99: @SmikeSix2 Have you checked the GCP quota? Also, you're correct—the FCM endpoints are unreliable, and their response times fluctuate. We're using Pushy as well. FCM, WebClient, Reactor Netty, and Pushy work well together in our setup. We can reach 150k rpm on single pod.(I do not remember specs).

emindeniz99: @SmikeSix2 80k rps is very high, 5million rpm :) fcm does not like it. Did not you reach the quota? Do you have 5m+ limit for fcm? Did you check on gcp console?

SmikeSix2: hehe its up to 350k rps... in bad scenarios. for example when the national team plays. old solution did fine with it, I was always complimenting firebase. but yeah probably need to switch to topics for the large teams.

my colleague has the access ill ask around, thanks for your input!

lahirumaramba (approved):

src/main/java/com/google/firebase/internal/ApiClientUtils.java
guitaoliu, line 91: This changed the default builder behaviour of the FirebaseOptions. Before the change, calling FirebaseOptions.Builder::build without explicitly providing an HttpTransport would reuse the HttpTransport initialised in Utils.TransportInstanceHolder.INSTANCE, but after the change, every FirebaseOptions builder build method creates a new ApacheHttp2Transport.

jonathanedey: Hey @guitaoliu, Thank you for catching this! We are looking to get this resolved for or next release.

PR #1056, opened 2025-01-08
chore: Fix non-existing bucket test

jonathanedey:
Changed bucket name to be globally unique and un-used.


lahirumaramba (approved):
Thanks! LGTM!

PR #1059, opened 2025-01-20
fix: `ApiClientUtils.getDefaultTransport()` should return a cached instance to an `ApacheHttp2Transport`

jonathanedey:
This fixes an unintentional change in `ApiClientUtils.getDefaultTransport()` which caused firebaseOptions to create a new instance of the default transport on `firebaseOptions` build. This only affected developers who made multiple calls to `FirebaseOptions.builder(...).build()` or initialized multiple app instances.


lahirumaramba (approved):
Thanks! LGTM!

PR #1065, opened 2025-01-25
fix: use .equals() instead of == for strings

yanikapitanov:
Simple fix for string comparison.

From what I could see these changes are already covered by unit tests and I was not sure how to add new ones to test this specific scenario.


jonathanedey (approved):
LGTM, thanks @yanikapitanov!

As for testing we will add a follow up PR with updated tests.

Issue #1045, opened 2024-12-09
NoClassDefFoundError when using service account credentials


sebphil:
I'm using the following code to initialize the firebase admin sdk:
```java
GoogleCredentials credentials = GoogleCredentials.fromStream(credentialsStream, () -> transport)
                .createScoped(myProperties.getScopes());
FirebaseOptions options = FirebaseOptions.builder()
                .setCredentials(credentials)
                .setHttpTransport(transport)
                .build();
```
The credentialsStream I'm injecting is read from a service account file.
This code, which worked fine in 9.4.1, now fails in 9.4.2 with the followin error:
`java.lang.NoClassDefFoundError: com/google/auth/CredentialTypeForMetrics`

As far as I can tell, this issue is caused by a dependency conflict between the following libs:
* google-api-client:2.7.0
  * google-auth-library-credentials:**1.23.0**
* google-auth-library-oauth2-http:1.29.0
  * google-auth-library-credentials:**1.29.0**
 
maven selects google-auth-library-credentials:1.23.0 which causes the ServiceAccountCredentials class to fail to load due to the CredentialTypeForMetrics class only existing since google-auth-library-credentials version 1.28.0.

Adding an explicit dependency to google-auth-library-credentials:1.29.0 in my maven project seems to workaround the issue.

janverhoeckx:
We have the same problem.

ashwin-nath-m:
Same problem. downgrading to 9.4.1 helped

lqiu96:
Could we see your pom.xml/ build.gradle files to help with troubleshooting and narrowing down the issue.

Pulling down v9.4.1 of Firebase, I see this from running `mvn dependency:tree`:
```
[INFO] --- maven-dependency-plugin:2.8:tree (default-cli) @ firebase-admin ---
[INFO] com.google.firebase:firebase-admin:jar:9.4.1
[INFO] +- com.google.api-client:google-api-client:jar:2.7.0:compile
[INFO] |  +- commons-codec:commons-codec:jar:1.17.0:compile
[INFO] |  +- com.google.oauth-client:google-oauth-client:jar:1.36.0:compile
[INFO] |  +- com.google.auth:google-auth-library-credentials:jar:1.27.0:compile <- Managed by Libraries-Bom
[INFO] |  +- com.google.http-client:google-http-client-gson:jar:1.45.0:compile
[INFO] |  +- com.google.http-client:google-http-client-apache-v2:jar:1.45.0:compile
[...]

PR #1066, opened 2025-01-27
chore: Rearrange firebase admin pom deps

lqiu96:
Following https://github.com/firebase/firebase-admin-java/issues/1045

Changes:
- Remove the duplicate google-api-client-bom as it is already part of libaries-bom
- Move google-auth-library-oauth2-http in front of google-api-client. This is to help ensure that the two auth modules (oauth2 and credentials) have the versions match. Maven's dep resolution strategy searches the path of the first deps it sees if the possible deps have the same number of layers away (https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Transitive_Dependencies)


jonathanedey (approved):
LGTM, Thanks again Lawrence!

PR #1068, opened 2025-01-28
chore: Update tests for getting and deleting provider uid via phone number and email.

jonathanedey:
This is a follow up of #1065


lahirumaramba (approved):
Thanks! LGTM!

Issue #1072, opened 2025-02-05
Error codes in TopicManagementResponse are incomplete, causing loss of important debug information


JosRoseboom:
### [REQUIRED] Step 2: Describe your environment

  * Operating System version: MacOs 15.1.1 and docker from arm64v8/amazoncorretto:21-al2023-headless
  * Firebase SDK version: 9.4.3
  * Library version: firebase-admin-java
  * Firebase Product: firebase-admin
  
### [REQUIRED] Step 3: Describe the problem

TopicManagementResponse has currently 4 error codes:

```
  private static final Map<String, String> ERROR_CODES = ImmutableMap.<String, String>builder()
      .put("INVALID_ARGUMENT", "invalid-argument")
      .put("NOT_FOUND", "registration-token-not-registered")
      .put("INTERNAL", "internal-error")
      .put("TOO_MANY_TOPICS", "too-many-topics")
```

Looking at https://developers.google.com/instance-id/reference/server , I see that it is currently also possible that RESOURCE_EXHAUSTED is returned.  We log the response if there are any errors and currently see a lot of:

[Error{index=0, reason=unknown-error}]

This is not helpful, while it could be caused by RESOURCE_EXHAUSTED, which would be really good to know.

The second problem is that the error string comes from :

```
 private Error(int index, String reason) {
      this.index = index;
this.reason = ERROR_CODES.containsKey(reason)
        ? ERROR_CODES.get(reason) : UNKNOWN_ERROR;
```

If the reason is not in the predefined ERROR_CODES it would be more helpful to use the reason String that came in as an parameter than UNKNOWN_ERROR. UNKNOWN_ERROR could be used for null or empty Strings.

#### Steps to reproduce:

Cause a RESOURCE_EXHAUSTED and see that the error description is unknown-error

#### Solution

I went ahead and patched this for our own use. I will create a PR containing a commit that adds RESOURCE_EXHAUSTED and a commit that uses the parameter reason as fallback in case of a non empty string.

JosRoseboom:
I created a pull request for this issue: https://github.com/firebase/firebase-admin-java/pull/1073

PR #1073, opened 2025-02-05
Issue 1072

JosRoseboom:
This pr contains 2 commits that provide a solution for the 2 problems described in https://github.com/firebase/firebase-admin-java/issues/1072

### DIFF
--- a/src/main/java/com/google/firebase/messaging/TopicManagementResponse.java
+++ b/src/main/java/com/google/firebase/messaging/TopicManagementResponse.java
@@ -41,6 +41,7 @@ public class TopicManagementResponse {
       .put("NOT_FOUND", "registration-token-not-registered")
       .put("INTERNAL", "internal-error")
       .put("TOO_MANY_TOPICS", "too-many-topics")
+      .put("RESOURCE_EXHAUSTED", "resource-exhausted")
       .build();
 
   private final int successCount;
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/firebase/messaging/TopicManagementResponse.java
```diff
@@ -101,8 +102,11 @@ public static class Error {
 
     private Error(int index, String reason) {
       this.index = index;
-      this.reason = ERROR_CODES.containsKey(reason)
-        ? ERROR_CODES.get(reason) : UNKNOWN_ERROR;
+      if (reason == null || reason.trim().isEmpty()) {
+        this.reason = UNKNOWN_ERROR;
+      } else {
+        this.reason = ERROR_CODES.getOrDefault(reason, reason);
```
jonathanedey: I agree that `unknown-error` isn't very helpful here and in the future we want to avoid needlessly abstracting information like this. This is a reasonable stop gap until we can address this fully in a breaking change.

For consistency with the current format of the error reasons, could you add some additional logic here to map them from `UPPER_SNAKE_CASE` to `lower-kebab-case`.

JosRoseboom: Thanks for checking this @jonathanedey .

I addressed this in commit https://github.com/firebase/firebase-admin-java/pull/1073/commits/869b553639ef2f40a14e25d22f31e48f14cc0448

With commit message:

> For consistency with the current format of the error reasons, logic is added to map them from UPPER_SNAKE_CASE to lower-kebab-case. Entries from the ERROR_CODES that can be computed this way are removed from the map. Extra tests are added for those 3 cases.

Please let me know if there are any other issues.

Thanks,
Jos

PR #1108, opened 2025-06-25
Implement custom signal support for server side remote config

athiramanu:
Hey there! So you want to contribute to a Firebase SDK? 
Before you file this pull request, please read these guidelines:

### Discussion

  * Read the contribution guidelines (CONTRIBUTING.md).
  * If this has been discussed in an issue, make sure to link to the issue here. 
    If not, go file an issue about this **before creating a pull request** to discuss.

### Testing

  * Make sure all existing tests in the repository pass after your change.
  * If you fixed a bug or added a feature, add a new test to cover your code.

### API Changes

  * At this time we cannot accept changes that affect the public API.  If you'd like to help 
    us make Firebase APIs better, please propose your change in an issue so that we 
    can discuss it together.


src/main/java/com/google/firebase/remoteconfig/ConditionEvaluator.java
```diff
@@ -0,0 +1,266 @@
...
+            (customSignal, targetSignal) -> customSignal.contains(targetSignal));
+      case STRING_DOES_NOT_CONTAIN:
+        return !compareStrings(targetCustomSignalValues, customSignalValue,
+            (customSignal, targetSignal) -> customSignal.contains(targetSignal));
+      case STRING_EXACTLY_MATCHES:
+        return compareStrings(targetCustomSignalValues, customSignalValue,
+            (customSignal, targetSignal) -> customSignal.equals(targetSignal));
+      case STRING_CONTAINS_REGEX:
+        return compareStrings(targetCustomSignalValues, customSignalValue,
+            (customSignal, targetSignal) -> Pattern.compile(targetSignal)
```
lahirumaramba: Does it make sense to compile and cache this (if we expect this to be used more frequently)

lahirumaramba: Should we also wrap this in a try-catch? Could this throw `java.util.regex.PatternSyntaxException`?

lahirumaramba (approved):
Thank you! Added a couple comments

src/main/java/com/google/firebase/remoteconfig/ConditionEvaluator.java
athiramanu: Skipping the caching part since the tagret signal compilation happens once for every request call. The function is moved to a try catch block where in case of error, it evaluates to false

PR #1112, opened 2025-07-08
chore: Fixed broken google cloud doc reference links

jonathanedey:
Updated deprecated google cloud reference doc links.


egilmorez (approved):
Good catch, thanks!

PR #1117, opened 2025-08-14
Ssrc bugbash fix

rathovarun1032:
This resolves two bugs identified during the recent bug bash related to server template evaluation and parameter handling.

Corrected Custom Signal Evaluation with Cached Templates
A bug was preventing custom signals from being evaluated correctly when the server template was initialized from a cache. The root cause was an issue with template serialization/deserialization  in ServerVersion and Condtion. The key percentOperator was being used instead of the correct percentConditionOperator. This has been fixed, ensuring that cached templates are parsed correctly and evaluations work as intended.

Fixed Malformed Logging Warning for getDouble()
When getDouble() was called on a parameter with a string value, it correctly returned the default static value but produced a malformed warning in the logs (e.g., "Unable to convert %s to double type."). The string formatting in the logger has been corrected to properly display the value that failed conversion.


src/main/java/com/google/firebase/remoteconfig/ServerVersion.java
ashish-kothari, line 1: This file looks identical to `Version.java`. Any reason to add it separately?

src/test/java/com/google/firebase/remoteconfig/ServerTemplateImplTest.java
```diff
@@ -63,6 +63,21 @@ public void testServerTemplateWithoutCacheValueThrowsException()
...
+        new ServerTemplateImpl.Builder(null)
+            .defaultConfig(defaultConfig)
+            .cachedTemplate(cacheTemplate)
+            .build();
+
+    ServerConfig evaluatedConfig = template.evaluate();
+
+    assertEquals("Default value", evaluatedConfig.getString("Custom"));
+  }
+
```
ashish-kothari, line 76: Should we also add a test to verify that parsing of cached template works as expected (all conditions, parameters, parameter groups are correctly parsed)?

src/main/java/com/google/firebase/remoteconfig/ServerTemplateImpl.java
```diff
@@ -93,8 +95,11 @@ public ServerConfig evaluate(KeysAndValues context) throws FirebaseRemoteConfigE
     }
 
     ConditionEvaluator conditionEvaluator = new ConditionEvaluator();
-    ImmutableMap<String, Boolean> evaluatedCondition = ImmutableMap.copyOf(
-        conditionEvaluator.evaluateConditions(cache.getServerConditions(), context));
+    ImmutableMap<String, Boolean> evaluatedCondition = 
+        context != null && !cache.getServerConditions().isEmpty()
+        ? ImmutableMap.copyOf(
+          conditionEvaluator.evaluateConditions(cache.getServerConditions(), context))
```
ashish-kothari: Can we directly return an `ImmutableMap` result here?

src/main/java/com/google/firebase/remoteconfig/ServerTemplateImpl.java
```diff
@@ -93,8 +95,11 @@ public ServerConfig evaluate(KeysAndValues context) throws FirebaseRemoteConfigE
     }
 
     ConditionEvaluator conditionEvaluator = new ConditionEvaluator();
-    ImmutableMap<String, Boolean> evaluatedCondition = ImmutableMap.copyOf(
-        conditionEvaluator.evaluateConditions(cache.getServerConditions(), context));
+    ImmutableMap<String, Boolean> evaluatedCondition = 
+        context != null && !cache.getServerConditions().isEmpty()
```
ashish-kothari: Move this inside `conditionEvaluator.evaluateConditions()`?

src/test/java/com/google/firebase/remoteconfig/ServerTemplateImplTest.java
rathovarun1032: Added Unit Testcase "testBuilderParsesCachedTemplateCorrectly" to verify parsing of cached template

src/main/java/com/google/firebase/remoteconfig/ServerTemplateImpl.java
rathovarun1032: moved inside conditionEvaluator.evaluateConditions()

src/main/java/com/google/firebase/remoteconfig/ServerTemplateImpl.java
rathovarun1032: yes, its good idea to return ImmutableMap

src/main/java/com/google/firebase/remoteconfig/ServerVersion.java
rathovarun1032: removed serverVersion.java

src/main/java/com/google/firebase/remoteconfig/ConditionEvaluator.java
```diff
@@ -35,44 +35,46 @@
...
+    if (context == null || conditions.isEmpty()) {
+      return ImmutableMap.of();
+    }
+    KeysAndValues evaluationContext =
+        context != null ? context : new KeysAndValues.Builder().build();
+
+    Map<String, Boolean> evaluatedConditions =
+        conditions.stream()
+            .collect(
+                ImmutableMap.toImmutableMap(
```
ashish-kothari: Consider using a static import.

rathovarun1032: done

src/main/java/com/google/firebase/remoteconfig/ConditionEvaluator.java
```diff
@@ -35,44 +36,46 @@
...
-        ? context 
-        : new KeysAndValues.Builder().build();
-
-    Map<String, Boolean> evaluatedConditions = conditions.stream()
-        .collect(Collectors.toMap(
-            ServerCondition::getName,
-            condition -> 
-                evaluateCondition(condition.getCondition(), evaluationContext, /* nestingLevel= */0)
-        ));
+    if (context == null || conditions.isEmpty()) {
```
lahirumaramba, line 59: if the context is null do we ignore the conditions even if it is a non-empty list?

lahirumaramba (changes requested):
LGTM! Added one comment. Thanks!

src/main/java/com/google/firebase/remoteconfig/ConditionEvaluator.java
rathovarun1032: Yes, we cannot evaluate conditions without context so we should ignore it

PR #1122, opened 2025-08-28
feat(rc): Add support for Server-Side Remote Config

rathovarun1032:
Introduce Server-Side Remote Config SDK API

This change adds a new API to the Firebase Admin SDK for  Java. The new API allows developers to fetch, cache, and evaluate Remote Config server templates, addressing the need for a performant, server-side solution for dynamic configuration. It includes new classes like ServerTemplate and ServerConfig, and a feature for Custom Targeting Signals to enable granular user segmentation.

RELEASE NOTES: Added support for configuring Remote Config server applications.


nk-1983 (commented):
test comment

src/main/java/com/google/firebase/remoteconfig/ServerTemplate.java
```diff
@@ -0,0 +1,45 @@
...
+  public interface Builder {
+
+    Builder defaultConfig(KeysAndValues config);
+
+    Builder cachedTemplate(String templateJson);
+    
+    ServerTemplate build();
+  }
+  /**
+  * Proccess the template data with a condition evaluator
```
nk-1983: typo

src/main/java/com/google/firebase/remoteconfig/ServerTemplate.java
```diff
@@ -0,0 +1,45 @@
...
+    
+    ServerTemplate build();
+  }
+  /**
+  * Proccess the template data with a condition evaluator 
+  * based on the provided context. 
+  */
+  ServerConfig evaluate(KeysAndValues context) throws FirebaseRemoteConfigException;
+  /**
+  * Proccess the template data without context.
```
nk-1983: typo

src/main/java/com/google/firebase/remoteconfig/ServerTemplateImpl.java
```diff
@@ -0,0 +1,188 @@
...
+  public ServerConfig evaluate(@Nullable KeysAndValues context)
+      throws FirebaseRemoteConfigException {
+    if (this.cache == null) {
+      throw new FirebaseRemoteConfigException(ErrorCode.FAILED_PRECONDITION,
+          "No Remote Config Server template in cache. Call load() before calling evaluate().");
+    }
+
+    Map<String, Value> configValues = new HashMap<>();
+    ImmutableMap<String, String> defaultConfigValues = defaultConfig.keysAndValues;
+    // Initializes config Value objects with default values.
```
nk-1983: configValue

src/main/java/com/google/firebase/remoteconfig/ServerTemplate.java
rathovarun1032: done

src/main/java/com/google/firebase/remoteconfig/ServerTemplateImpl.java
rathovarun1032: done

src/main/java/com/google/firebase/remoteconfig/ServerTemplate.java
rathovarun1032: done

src/main/java/com/google/firebase/remoteconfig/AndCondition.java
```diff
@@ -0,0 +1,62 @@
...
+    checkArgument(!conditionList.isEmpty(), 
+        "List of conditions for AND operation must not be empty");
+    this.conditions = conditionList.stream()
+                                  .map(OneOfCondition::new) 
+                                  .collect(ImmutableList.toImmutableList());
+  }
+
+  @NonNull
+  List<OneOfCondition> getConditions() {
+    return new ArrayList<>(conditions);
```
lahirumaramba: `conditions` field is already an `ImmutableList`. Do we need to create a mutable copy here? Or could we simply just return the immutable list?

```java
@NonNull
List<OneOfCondition> getConditions() {
  return this.conditions;
}
```

src/main/java/com/google/firebase/remoteconfig/OrCondition.java
```diff
@@ -0,0 +1,59 @@
...
+    checkNotNull(conditionList, "List of conditions for AND operation cannot be null.");
+    checkArgument(!conditionList.isEmpty(), "List of conditions for AND operation cannot be empty");
+    this.conditions = conditionList.stream()
+                                  .map(OneOfCondition::new) 
+                                  .collect(ImmutableList.toImmutableList());
+  }
+
+  @NonNull
+  List<OneOfCondition> getConditions() {
+    return new ArrayList<>(conditions);
```
lahirumaramba: Same as in `AndCondition.java`:
`conditions` field is already an `ImmutableList`. Do we need to create a mutable copy here? Or could we simply just return the immutable list?

```java
@NonNull
List<OneOfCondition> getConditions() {
  return this.conditions;
}
```

src/main/java/com/google/firebase/remoteconfig/ServerTemplateImpl.java
```diff
@@ -0,0 +1,188 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class ServerTemplateImpl implements ServerTemplate {
+
+  private final KeysAndValues defaultConfig;
+  private FirebaseRemoteConfigClient client;
+  private ServerTemplateData cache;
```
lahirumaramba: I wonder if we should make the `cache` filed atomic to improve thread safety... I was thinking about this because one thread could call `load()` while another thread is calling `evaluate()`. This can lead to `evaluate()` operating on stale data.

Since you are more familiar with the use-cases and the workflows I will leave it up to you to make the final call. I see that you have an atomic `cachedTemplate` but `evaluate()` reads directly from `this.cache`, the non-thread-safe field.

src/main/java/com/google/firebase/remoteconfig/ConditionEvaluator.java
```diff
@@ -0,0 +1,346 @@
...
+    return 0;
+  }
+
+  private List<Integer> parseSemanticVersion(String versionString) {
+    return Arrays.stream(versionString.split("\\."))
+        .map(Integer::parseInt)
+        .collect(Collectors.toList());
+  }
+
+  private boolean validateSemanticVersion(String version) {
```
lahirumaramba, line 343: Depends on how often this helper function will be called in runtime. We could optimize it by caching the result from the RegEx compilation.

```java
private static final Pattern SEMVER_PATTERN =  Pattern.compile("^[0-9]+(?:\\.[0-9]+){0,4}$");

private boolean validateSemanticVersion(String version) {
  return SEMVER_PATTERN.matcher(version).matches();
}
```

src/main/java/com/google/firebase/remoteconfig/AndCondition.java
athiramanu: Updated the return type

src/main/java/com/google/firebase/remoteconfig/OrCondition.java
athiramanu: Updated the return type

src/main/java/com/google/firebase/remoteconfig/ConditionEvaluator.java
athiramanu: Done

src/main/java/com/google/firebase/remoteconfig/ServerTemplateImpl.java
rathovarun1032: @lahirumaramba - Thank you for reviewing this code changes
 I have changed the cache field to an AtomicReference<ServerTemplateData>. This will guarantee that the reference to the ServerTemplateData object is updated atomically and that changes are visible across threads.

PR #1145, opened 2025-10-17
Add support for new bandwidthConstrainedOk flag used to deliver notifications to satellite networks

andrew-signal:
### Discussion

Hi! The Android Developers page for [Developing for constrained satellite networks](https://developer.android.com/develop/connectivity/satellite/constrained-networks) indicates that we need to pass a new flag, `bandwidth_constrained_ok` in our AndroidConfig options, like so, to ensure delivery of our pushes on constrained satellite networks. 

```json
{
  "message":{
    "token":"bk3RNwTe3H0:CI2k_HHwgIpoDKCIZvvDMExUdFQ3P1...",
    "notification":{
      "title":"Portugal vs. Denmark",
      "body":"great match!"
    }
    "android": {
       "bandwidth_constrained_ok": true
    }
  }
}
```

In order to do this in a server app based on the firebase-admin-java SDK, we need to expose a config option in `AndroidConfigBuilder`.

(In practice, things seem to be okay without this flag, but the documentation says we should add it, so I'm inclined to add it.)

### Testing

- I added a new unit test to match the already extant one for `direct_boot_ok`.  
- `mvn test` passes locally on my machine. 
- I expect CI will also run on this PR. 

### API Changes

  * This does add the minimal new config parameter to set this parameter in `AndroidConfigBuilder`, which is the least intrusive change I can make to the public API to expose this now-needed functionality.


jonathanedey:
cc: @lahirumaramba, I'll merge this PR as is and have the docs updated and reviewed in a follow up pr along with the `RestrictedSatelliteOk` api.

PR #1170, opened 2026-01-13
feat(auth):  Add support for FPNV

boikoa-gl:
Added support for FPNV


src/main/java/com/google/firebase/fpnv/internal/FirebasePnvTokenVerifier.java
```diff
@@ -0,0 +1,204 @@
...
+   *
+   * @param token String input data
+   * @return {@link FirebasePnvToken}
+   * @throws FirebasePnvException Can throw {@link FirebasePnvException}
+   */
+  public FirebasePnvToken verifyToken(String token) throws FirebasePnvException {
+    checkArgument(!Strings.isNullOrEmpty(token), "FPNV token must not be null or empty");
+
+    try {
+      // Parse the token first to inspect header
```
sam-gc: small nit: let's remove these comments; the code is clear-enough as-is

src/main/java/com/google/firebase/fpnv/internal/FirebasePnvTokenVerifier.java
```diff
@@ -0,0 +1,204 @@
...
+          "Check your project: " + projectId + ". "
+          + "Failed to verify FPNV token signature: " + e.getMessage(),
+          e
+      );
+    }
+  }
+
+  private void verifyHeader(JWSHeader header) throws FirebasePnvException {
+    // Check Algorithm (alg)
+    if (!JWSAlgorithm.ES256.equals(header.getAlgorithm())) {
```
sam-gc: Here and in related places throughout: let's reverse the equality here; I think it reads a bit cleaner (i.e. `if (!header.getAlgorithm().equals(JWSAlgorithm.ES256))`

I know there are some schools of thought which prefer putting the constant first, but our styleguide has converged on having the variable come first since it's the more common pattern.

src/main/java/com/google/firebase/fpnv/internal/FirebasePnvTokenVerifier.java
```diff
@@ -0,0 +1,204 @@
...
+              + " but got " + header.getAlgorithm());
+    }
+    // Check Key ID (kid)
+    if (Strings.isNullOrEmpty(header.getKeyID())) {
+      throw new FirebasePnvException(
+          FirebasePnvErrorCode.INVALID_ARGUMENT,
+          "FPNV has no 'kid' claim."
+      );
+    }
+    // Check Typ (typ)
```
sam-gc: `// Check Type (typ)`

src/main/java/com/google/firebase/fpnv/FirebasePnvException.java
```diff
@@ -0,0 +1,60 @@
...
+      FirebasePnvErrorCode errorCode,
+      String message
+  ) {
+    this(errorCode, message, null);
+  }
+
+  public FirebasePnvErrorCode getFpnvErrorCode() {
+    return errorCode;
+  }
+}
```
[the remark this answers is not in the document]
sam-gc: I'm not sure, but I think the robot is right: if other modules extend from FirebaseException, that's what we should do here too

src/main/java/com/google/firebase/fpnv/FirebasePnvToken.java
```diff
@@ -0,0 +1,96 @@
...
+   */
+  public long getIssuedAt() {
+    return ((java.util.Date) claims.get("iat")).getTime() / 1000L;
+  }
+
+  /**
+   * Returns the entire map of claims.
+   */
+  public Map<String, Object> getClaims() {
+    return claims;
```
sam-gc, line 93: We should probably return this as a copy in an `ImmutableMap`

src/test/java/com/google/firebase/fpnv/internal/FpnvTokenVerifierTest.java
```diff
@@ -0,0 +1,327 @@
...
+        .issuer(ISSUER)
+        .audience(ISSUER)
+        .subject("+1555")
+        .expirationTime(new Date(System.currentTimeMillis() + 10000))
+        .build();
+
+    String tokenString = createToken(header, claims);
+    ExpiredJWTException error = new ExpiredJWTException("Bad token");
+
+    when(mockJwtProcessor.process(any(SignedJWT.class), any())).thenThrow(error);
```
sam-gc, line 277: Is it possible to test this by setting an expiration time in the past instead, instead of just forcing the processor to throw an error?

src/test/java/com/google/firebase/fpnv/internal/FpnvTokenVerifierTest.java
```diff
@@ -0,0 +1,327 @@
...
+  public void testVerifyToken_Claims_WrongAudience() throws Exception {
+    JWTClaimsSet badClaims = new JWTClaimsSet.Builder()
+        .issuer("https://wrong.com") // Wrong issuer
+        .audience(ISSUER)
+        .subject("+1555")
+        .expirationTime(new Date(System.currentTimeMillis() + 10000))
+        .build();
+
+    String tokenString = createToken(header, badClaims);
+    when(mockJwtProcessor.process(any(SignedJWT.class), any())).thenReturn(badClaims);
```
sam-gc, line 296: Is this actually testing if the issuer is wrong?

pom.xml
```diff
@@ -446,6 +446,11 @@
             <artifactId>httpclient5</artifactId>
             <version>5.3.1</version>
         </dependency>
+        <dependency>
+            <groupId>com.nimbusds</groupId>
+            <artifactId>nimbus-jose-jwt</artifactId>
+            <version>10.6</version>
```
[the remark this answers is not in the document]
sam-gc: Does this not apply?

src/main/java/com/google/firebase/fpnv/internal/FirebasePnvTokenVerifier.java
boikoa-gl: Sure. I will.
Can I suggest another option?
```
(!Objects.equals(header.getAlgorithm(), JWSAlgorithm.ES256))
```
Make sense?

src/test/java/com/google/firebase/fpnv/internal/FpnvTokenVerifierTest.java
boikoa-gl: Yes. we check `issuer` and `audience`

src/test/java/com/google/firebase/fpnv/internal/FpnvTokenVerifierTest.java
boikoa-gl: Previously we had this test and logic for this case
```
@Test
public void testVerifyToken_Claims_Expired() throws Exception {
  Date past = new Date(System.currentTimeMillis() - 10000); // Expired

  JWTClaimsSet expiredClaims = new JWTClaimsSet.Builder()
      .issuer(ISSUER)
      .audience(ISSUER)
      .subject("+1555")
      .expirationTime(past)
      .build();

  String tokenString = createToken(header, expiredClaims);

  // Mock processor returning the expired claims
  when(mockJwtProcessor.process(any(SignedJWT.class), any())).thenReturn(expiredClaims);

  FirebasePnvException e = assertThrows(FirebasePnvException.class, () ->
      verifier.verifyToken(tokenString)
  );

  assertEquals(FirebasePnvErrorCode.TOKEN_EXPIRED, e.getFpnvErrorCode());
}
```
And check for this 
```
    // TODO: i guess this is redundant
    //  jwtProcessor.process did this already
    // Verify Expiration
    Date now = new Date();
    Date exp = claims.getExpirationTime();
    if (exp == null || now.after(exp)) {
      throw new FirebasePnvException(FirebasePnvErrorCode.TOKEN_EXPIRED, "Token has expired.");
    }
```
But, as expected, I received a comment.
```
This manual expiration check is redundant. The DefaultJWTProcessor already validates the token's expiration time and will throw an ExpiredJWTException if it has expired. This exception is already being caught and handled in the verifyToken method. You can safely remove this block of code, as the TODO comment suggests.
````
Therefore i've removed it at all

src/main/java/com/google/firebase/fpnv/FirebasePnvException.java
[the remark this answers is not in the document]
boikoa-gl: Done

lahirumaramba:
Superseded by #1203

PR #1202, opened 2026-04-07
chore: Fix broken integrations tests

jonathanedey:
* Replaces broken test links with a fake local server.
* Removes ReadTimeout and  integration tests as these are difficult to simulate with a mock server.

The previous test links caused failures since their targets were moved to a new link. These tests were already flaky due to hitting an external live endpoint and to mitigate this we are converting to using a local server to validate these tests instead. Using a local server makes testing the Read and Write timeouts difficult and as a result these were removed and rely on unit tests for coverage.


jonathanedey:
Closing this PR for now since the flaky tests have auto resolved themselves. If these test begin to flake frequently again in the future we will re-address the value of these tests.

PR #1203, opened 2026-05-05
feat(pnv): Add support for Firebase Phone Number Verification

lahirumaramba:
Add support for Firebase Phone Number Verification


lahirumaramba:
@gemini-code-assist review

lahirumaramba:
@gemini-code-assist review

lahirumaramba:
@gemini-code-assist review

lahirumaramba:
@gemini-code-assist review

pom.xml
```diff
@@ -446,6 +446,11 @@
             <artifactId>httpclient5</artifactId>
             <version>5.3.1</version>
         </dependency>
+        <dependency>
+            <groupId>com.nimbusds</groupId>
+            <artifactId>nimbus-jose-jwt</artifactId>
+            <version>10.6</version>
```
jonathanedey: Can we bump this to latest 10.9?

src/main/java/com/google/firebase/phonenumberverification/FirebasePhoneNumberVerificationErrorCode.java
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Error codes that are used in {@link FirebasePhoneNumberVerification}.
+ */
+public enum FirebasePhoneNumberVerificationErrorCode {
+  INVALID_ARGUMENT,
+  INVALID_TOKEN,
+  TOKEN_EXPIRED,
+  INTERNAL_ERROR,
+  SERVICE_ERROR,
+}
```
jonathanedey, line 48: Can we add doc strings to these?

src/main/java/com/google/firebase/phonenumberverification/FirebasePhoneNumberVerificationErrorCode.java
```diff
@@ -0,0 +1,28 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.firebase.phonenumberverification;
+
+/**
+ * Error codes that are used in {@link FirebasePhoneNumberVerification}.
```
jonathanedey: We should match the error doc sting we use to the other error codes. Something along the lines of:
```suggestion
 * Error codes that can be raised by the Phone Number Verification APIs. 
```

src/test/java/com/google/firebase/phonenumberverification/internal/FirebasePhoneNumberVerificationTokenVerifierTest.java
```diff
@@ -0,0 +1,501 @@
...
+        + "\\nxk7rLcIEXovagJC82ZCGujo5joJ3fam9/q9I5ju5xw13yMOHyeyzsErCpSP/Xr8f"
+        + "\\nr5uOncBw2twGqOZ+FlQtCdE1Dg==\\n-----END PRIVATE KEY-----\\n\",\n"
+        + "  \"client_email\": \"mock-project-id-none@mock-project-id.iam.gserviceaccount.com\",\n"
+        + "  \"client_id\": \"1234567890\",\n"
+        + "  \"auth_uri\": \"https://accounts.google.com/o/oauth2/auth\",\n"
+        + "  \"token_uri\": \"https://accounts.google.com/o/oauth2/token\",\n"
+        + "  \"auth_provider_x509_cert_url\": \"https://www.googleapis.com/oauth2/v1/certs\",\n"
+        + "  \"client_x509_cert_url\": \"https://www.googleapis.com/robot/v1/metadata/x509/"
+        + "mock-project-id-none%40mock-project-id.iam.gserviceaccount.com\"\n"
+        + "}";
```
jonathanedey: Think this is the fake key we define in the test resources `none.json`. If so, can we reference it instead. 


```suggestion
    String jsonString = ServiceAccount.NONE.asString();
```

src/main/java/com/google/firebase/phonenumberverification/FirebasePhoneNumberVerificationException.java
```diff
@@ -0,0 +1,81 @@
...
+      case TOKEN_EXPIRED:
+      case INVALID_TOKEN:
+        return ErrorCode.UNAUTHENTICATED;
+      case SERVICE_ERROR:
+        return ErrorCode.UNAVAILABLE;
+      case INTERNAL_ERROR:
+      default:
+        return ErrorCode.INTERNAL;
+    }
+  }
```
jonathanedey: It looks like we are using a different pattern for determining the base Firebase Error code (Using a map rather than defining it manually) than what we use for the other services. I don't have a strong opinion against it since it think this it makes modifying these easier. But wanted to get your thoughts on it.

jonathanedey (approved):
LGTM with a few comments, Thanks!

src/test/java/com/google/firebase/phonenumberverification/internal/FirebasePhoneNumberVerificationTokenVerifierTest.java
lahirumaramba: Good catch! I didn't change much from the initial implementation, but this doesn't look like it is necessary to test no project ID test case. Updated it to match the unit tests for other existing services.

src/main/java/com/google/firebase/phonenumberverification/FirebasePhoneNumberVerificationException.java
lahirumaramba: You are right! This was also left intact from the initial PR. I feel it makes constructing new errors easier, so I'm leaning towards keeping this approach.

lahirumaramba: Actually, I ended up changing the error handling to match the rest of the services :)

src/main/java/com/google/firebase/phonenumberverification/FirebasePhoneNumberVerification.java
```diff
@@ -0,0 +1,84 @@
...
+import com.google.firebase.FirebaseApp;
+import com.google.firebase.ImplFirebaseTrampolines;
+import com.google.firebase.internal.FirebaseService;
+import com.google.firebase.phonenumberverification.internal.FirebasePhoneNumberVerificationTokenVerifier;
+
+/**
+ * This class is the entry point for the Firebase Phone Number Verification service.
+ *
+ * <p>You can get an instance of {@link FirebasePhoneNumberVerification} via {@link #getInstance()},
+ * or {@link #getInstance(FirebaseApp)} and then use it.
```
kevinthecheung:
```suggestion
 * or {@link #getInstance(FirebaseApp)}.
```

kevinthecheung (approved):

PR #1205, opened 2026-05-13
chore: Fix issue in the new RC unit tests

rathovarun1032:
fix(rc): Throw IllegalArgumentException for invalid template JSON

The `ServerTemplateImpl` constructor was catching template parsing 
exceptions and only printing the stack trace to stderr. This resulted 
in noisy console output during unit tests, even when the tests passed. 

This change updates the constructor to throw an `IllegalArgumentException` 
when a provided `cachedTemplate` fails to parse. 

Updated `testEvaluateWithInvalidCacheValueThrowsException` to assert 
this new behavior.

### DIFF
--- a/src/test/java/com/google/firebase/remoteconfig/ServerTemplateImplTest.java
+++ b/src/test/java/com/google/firebase/remoteconfig/ServerTemplateImplTest.java
@@ -200,7 +200,7 @@ public class ServerTemplateImplTest {
     assertEquals("Unable to parse JSON string.", error.getMessage());
     // Verify the cause is the original FirebaseRemoteConfigException
     assertTrue(error.getCause() instanceof FirebaseRemoteConfigException);
- }
+  }
 
   @Test
   public void testEvaluateWithInAppDefaultReturnsEmptyString() throws Exception {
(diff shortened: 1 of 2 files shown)


lahirumaramba (approved):
Thanks! LGTM

src/test/java/com/google/firebase/remoteconfig/ServerTemplateImplTest.java
```diff
@@ -188,20 +189,17 @@ public void testEvaluateWithoutDefaultValueReturnsEmptyString()
...
-        new ServerTemplateImpl.Builder(null)
-            .defaultConfig(defaultConfig)
-            .cachedTemplate(invalidJsonString)
-            .build();
-
-    FirebaseRemoteConfigException error =
-        assertThrows(FirebaseRemoteConfigException.class, () -> template.evaluate(context));
-
-    assertEquals(
-        "No Remote Config Server template in cache. Call load() before " + "calling evaluate().",
```
lahirumaramba, line 203: Do we test this use-case anywhere else?

lahirumaramba (approved):

src/test/java/com/google/firebase/remoteconfig/ServerTemplateImplTest.java
rathovarun1032: yes, we test this in other testcase

PR #1232, opened 2026-08-12
chore: update GitHub Actions to Node 24 versions and refactor workflow permissions for zizmor security compliance

yvonnep165:
Updates GitHub Action dependencies across workflows to Node 24-compatible versions to resolve Node 20 runner deprecation warnings, and refactors workflow permissions for zizmor security compliance (following the pattern implemented in https://github.com/firebase/firebase-admin-python/pull/956).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,10 +24,10 @@ jobs:
         java-version: [8, 11, 17]
 
     steps:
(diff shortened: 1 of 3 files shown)


jonathanedey (approved):
LGTM! Thanks for fixing the zizmor warnings too, lets add that to the PR title as well.

yvonnep165:
> LGTM! Thanks for fixing the zizmor warnings too, lets add that to the PR title as well.

Ok! Added.

PR #1233, opened 2026-08-24
feat(appcheck): Add App Check token verification support with replay protection

yvonnep165:
This PR adds support for App Check standard token verification and one-time token verification for replay protection. This enables backend servers to verify standard App Check JWTs locally and optionally verify/consume limited-use (replay-protected) tokens via backend RPC.

- Adds the `FirebaseAppCheck` service entry point with synchronous and asynchronous `verifyToken` methods.
- Creates the data models and options (`DecodedAppCheckToken`, `VerifyAppCheckTokenResponse`, `VerifyAppCheckTokenOptions`, `FirebaseAppCheckException`) for token verification results and options.
- Verifies tokens locally using App Check public keys and calls the backend to consume one-time tokens.
- Adds full unit test suites

### DIFF
--- a/src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
+++ b/src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
@@ -81,9 +81,13 @@ public class DecodedAppCheckToken {
     if (audience instanceof String) {
       return ImmutableList.of((String) audience);
     } else if (audience instanceof List) {
-      @SuppressWarnings("unchecked")
-      List<String> audienceList = (List<String>) audience;
-      return ImmutableList.copyOf(audienceList);
+      ImmutableList.Builder<String> builder = ImmutableList.builder();
+      for (Object item : (List<?>) audience) {
+        if (item instanceof String) {
+          builder.add((String) item);
+        }
+      }
+      return builder.build();
     }
     return ImmutableList.of();
   }
--- a/src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
+++ b/src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
@@ -88,14 +88,14 @@ public class DecodedAppCheckToken {
   }
 
   /**
-   * Returns the expiration time as an {@link Instant}, or {@code null} if not present.
+   * Returns the expiration time as an {@link Instant}.
    */
   public Instant getExpirationTime() {
     return toInstant(claims.get("exp"));
   }
 
   /**
-   * Returns the issued-at time as an {@link Instant}, or {@code null} if not present.
+   * Returns the issued-at time as an {@link Instant}.
    */
   public Instant getIssuedAt() {
     return toInstant(claims.get("iat"));
--- a/src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
+++ b/src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
@@ -37,7 +37,7 @@ public class DecodedAppCheckToken {
    *
    * @param claims A map of JWT claims.
    */
-  DecodedAppCheckToken(Map<String, Object> claims) {
+  public DecodedAppCheckToken(Map<String, Object> claims) {
     checkNotNull(claims, "Claims map must not be null");
     checkArgument(claims.containsKey("sub"), "Claims map must contain sub");
     this.claims = ImmutableMap.copyOf(claims);
--- a/src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
+++ b/src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
@@ -59,6 +59,28 @@ public class DecodedAppCheckToken {
   }
 
   /**
+   * Returns the App ID for which this token was issued.
+   * This is an alias for {@link #getSubject()}.
+   */
+  public String getAppId() {
+    return getSubject();
+  }
+
+  /**
+   * Returns the JWT ID ('jti') of the token, or {@code null} if not present.
+   */
+  public String getJti() {
+    return (String) claims.get("jti");
+  }
+
+  /**
+   * Returns the attestation provider for this token, or {@code null} if not present.
+   */
+  public String getProvider() {
+    return (String) claims.get("provider");
+  }
+
+  /**
    * Returns the audience for which this token is intended.
    */
   public List<String> getAudience() {
--- a/src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
+++ b/src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
@@ -1,0 +1,103 @@
+/*
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.firebase.appcheck;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Represents a verified Firebase App Check token.
+ */
+public class DecodedAppCheckToken {
+
+  private final Map<String, Object> claims;
+
+  /**
+   * Creates an instance of {@link DecodedAppCheckToken} from a map of JWT claims.
+   *
+   * @param claims A map of JWT claims.
+   */
+  DecodedAppCheckToken(Map<String, Object> claims) {
+    checkNotNull(claims, "Claims map must not be null");
+    checkArgument(claims.containsKey("sub"), "Claims map must contain sub");
+    this.claims = ImmutableMap.copyOf(claims);
+  }
+
+  /**
+   * Returns the issuer identifier for the token.
+   */
+  public String getIssuer() {
+    return (String) claims.get("iss");
+  }
+
+  /**
+   * Returns the subject claim ('sub') of the token.
+   */
+  public String getSubject() {
+    return (String) claims.get("sub");
+  }
+
+  /**
+   * Returns the audience for which this token is intended.
+   */
+  public List<String> getAudience() {
+    Object audience = claims.get("aud");
+    if (audience instanceof String) {
+      return ImmutableList.of((String) audience);
+    } else if (audience instanceof List) {
+      @SuppressWarnings("unchecked")
+      List<String> audienceList = (List<String>) audience;
+      return ImmutableList.copyOf(audienceList);
+    }
+    return ImmutableList.of();
+  }
+
+  /**
+   * Returns the expiration time in seconds since the Unix epoch.
+   */
+  public long getExpirationTime() {
+    Object exp = claims.get("exp");
+    if (exp instanceof Date) {
+      return ((Date) exp).getTime() / 1000L;
+    }
+    return exp instanceof Number ? ((Number) exp).longValue() : 0L;
+  }
+
+  /**
+   * Returns the issued-at time in seconds since the Unix epoch.
+   */
+  public long getIssuedAt() {
+    Object iat = claims.get("iat");
+    if (iat instanceof Date) {
+      return ((Date) iat).getTime() / 1000L;
+    }
+    return iat instanceof Number ? ((Number) iat).longValue() : 0L;
+  }
+
+  /**
+   * Returns the entire map of claims.
+   */
+  public Map<String, Object> getClaims() {
+    return claims;
+  }
+}
--- a/src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
+++ b/src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
@@ -20,9 +20,10 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableMap;
 import java.time.Instant;
+import java.util.Collections;
 import java.util.Date;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
@@ -41,7 +42,7 @@ public class DecodedAppCheckToken {
   public DecodedAppCheckToken(Map<String, Object> claims) {
     checkNotNull(claims, "Claims map must not be null");
     checkArgument(claims.containsKey("sub"), "Claims map must contain sub");
-    this.claims = ImmutableMap.copyOf(claims);
+    this.claims = Collections.unmodifiableMap(new HashMap<>(claims));
   }
 
   /**
--- a/src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
+++ b/src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
@@ -59,14 +59,6 @@ public class DecodedAppCheckToken {
   }
 
   /**
-   * Returns the App ID for which this token was issued.
-   * This is an alias for {@link #getSubject()}.
-   */
-  public String getAppId() {
-    return getSubject();
-  }
-
-  /**
    * Returns the JWT ID ('jti') of the token, or {@code null} if not present.
    */
   public String getJti() {
--- a/src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
+++ b/src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
@@ -21,6 +21,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import java.time.Instant;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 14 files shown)


src/main/java/com/google/firebase/appcheck/internal/AppCheckTokenVerifier.java
```diff
@@ -0,0 +1,304 @@
...
+  private static final String VERIFY_TOKEN_URL_FORMAT =
+      "https://firebaseappcheck.googleapis.com/v1/projects/%s:verifyAppCheckToken";
+
+  private final FirebaseApp app;
+  private final String projectId;
+  private final HttpRequestFactory requestFactory;
+  private final JsonFactory jsonFactory;
+  private volatile DefaultJWTProcessor<SecurityContext> jwtProcessor;
+
+  public AppCheckTokenVerifier(FirebaseApp app) {
```
weixifan, line 81: In the Admin SDK, do we use Guice or Dagger to manage the dependency injection of parameters? If so, a singleton-scoped JWK parameter can be injected and managed by Guice or Dagger instead.

src/main/java/com/google/firebase/appcheck/internal/AppCheckTokenVerifier.java
```diff
@@ -0,0 +1,304 @@
...
+          new JWSVerificationKeySelector<>(JWSAlgorithm.RS256, keySource);
+      processor.setJWSKeySelector(keySelector);
+    } catch (MalformedURLException e) {
+      throw new IllegalStateException("Invalid JWKS URL", e);
+    }
+    return processor;
+  }
+
+  protected JWKSource<SecurityContext> createKeySource() throws MalformedURLException {
+    return JWKSourceBuilder.create(URI.create(JWKS_URL).toURL()).retrying(true).build();
```
weixifan: Can we please cache this for 6 hours, using `.cache(...)`? The refresh timeout can be left at default (15 seconds).

https://www.javadoc.io/doc/com.nimbusds/nimbus-jose-jwt/latest/com/nimbusds/jose/jwk/source/JWKSourceBuilder.html#cache(long,long)

src/main/java/com/google/firebase/appcheck/internal/AppCheckTokenVerifier.java
```diff
@@ -0,0 +1,304 @@
...
+    this.app = checkNotNull(app, "FirebaseApp must not be null");
+    this.projectId = getProjectId(app);
+    this.requestFactory = checkNotNull(requestFactory, "HttpRequestFactory must not be null");
+    this.jsonFactory = checkNotNull(jsonFactory, "JsonFactory must not be null");
+    this.jwtProcessor = jwtProcessor;
+  }
+
+  private DefaultJWTProcessor<SecurityContext> getJwtProcessor() {
+    DefaultJWTProcessor<SecurityContext> processor = this.jwtProcessor;
+    if (processor == null) {
```
weixifan: Is there a reason this initialization logic can't be put into the constructor? Doing so avoids the subtle and error-prone synchronization logic. For example, the necessity of the double null-check is not immediately obvious.

src/main/java/com/google/firebase/appcheck/internal/AppCheckTokenVerifier.java
```diff
@@ -0,0 +1,304 @@
...
+  private void verifyClaims(JWTClaimsSet claims) throws FirebaseAppCheckException {
+    checkNotNull(claims, "JWTClaimsSet claims must not be null");
+    String issuer = claims.getIssuer();
+
+    if (Strings.isNullOrEmpty(issuer)) {
+      throw new FirebaseAppCheckException(
+          ErrorCode.INVALID_ARGUMENT, "App Check token has no 'iss' (issuer) claim.");
+    }
+
+    if (!issuer.startsWith(APP_CHECK_ISSUER)) {
```
weixifan, line 233: I think in Admin SDK, we don't have access to the developer's project number. This is actually a deficiency, as https://google.aip.dev/cloud/2510 states that the project number is the canonical identifier, and there are issues in working with project IDs in general.

In the future, if we ever address this issue and we finally get access to the project number, it's actually better to check for strict equality to `https://firebaseappcheck.googleapis.com/<project_number>` here -- for example, `https://firebaseappcheck.googleapis.com/12345678`.

Perhaps a comment explaining this would be appropriate here.

src/main/java/com/google/firebase/appcheck/internal/AppCheckTokenVerifier.java
```diff
@@ -0,0 +1,304 @@
...
+    } catch (ParseException e) {
+      throw new FirebaseAppCheckException(
+          ErrorCode.INVALID_ARGUMENT, "Failed to parse App Check JWT token: " + e.getMessage(), e);
+    } catch (ExpiredJWTException e) {
+      throw new FirebaseAppCheckException(
+          ErrorCode.INVALID_ARGUMENT, "Firebase App Check token has expired.", e);
+    } catch (BadJOSEException e) {
+      throw new FirebaseAppCheckException(
+          ErrorCode.INVALID_ARGUMENT,
+          "Check your project: " + projectId + ". Firebase App Check token is invalid: "
```
weixifan: I think this exception type is the one corresponding to an invalid signature, while the `JOSEException` is thrown when an internal error happens.

https://www.javadoc.io/doc/com.nimbusds/nimbus-jose-jwt/latest/com/nimbusds/jwt/proc/DefaultJWTProcessor.html#process(com.nimbusds.jwt.SignedJWT,C)

src/main/java/com/google/firebase/appcheck/internal/AppCheckTokenVerifier.java
```diff
@@ -0,0 +1,304 @@
...
+  }
+
+  /**
+   * Performs local JWT signature and claims verification.
+   */
+  private DecodedAppCheckToken verifyTokenLocally(String token) throws FirebaseAppCheckException {
+    try {
+      SignedJWT signedJwt = SignedJWT.parse(token);
+      verifyHeader(signedJwt.getHeader());
+      JWTClaimsSet claims = getJwtProcessor().process(signedJwt, null);
```
weixifan: Tink actually doesn't even allow the token to be parsed before verifying its signature. In general, it's good practice to verify the cryptographic signature as early as possible before doing anything with the object.

Reference: https://developers.google.com/tink/create-jwts-signature#jwts.

src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
```diff
@@ -0,0 +1,103 @@
...
+public class DecodedAppCheckToken {
+
+  private final Map<String, Object> claims;
+
+  /**
+   * Creates an instance of {@link DecodedAppCheckToken} from a map of JWT claims.
+   *
+   * @param claims A map of JWT claims.
+   */
+  public DecodedAppCheckToken(Map<String, Object> claims) {
```
weixifan: Ah, it's too bad that this class requires a public constructor due to being in a different class. Depending on how you feel about this class referring to a class in `appcheck/internal`, you could use the "Romeo and Juliet" mechanism to restrict access of this constructor to just `AppCheckTokenVerifier` (and other internal classes as needed), and it may actually be worth doing so to prevent misuse of this class.

This is optional and I don't feel strongly about this either way.

src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
```diff
@@ -0,0 +1,103 @@
...
+    checkNotNull(claims, "Claims map must not be null");
+    checkArgument(claims.containsKey("sub"), "Claims map must contain sub");
```
weixifan, line 44: We should probably also check the other required claims `iss`, `aud`, `exp`, and `iat`.

src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
```diff
@@ -0,0 +1,103 @@
...
+      List<String> audienceList = (List<String>) audience;
+      return ImmutableList.copyOf(audienceList);
+    }
+    return ImmutableList.of();
+  }
+
+  /**
+   * Returns the expiration time in seconds since the Unix epoch.
+   */
+  public long getExpirationTime() {
```
weixifan: I think `java.time.Instant` is a better candidate to hold instants in time, since we have access to Java 8.

src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
```diff
@@ -0,0 +1,103 @@
...
+    if (exp instanceof Date) {
+      return ((Date) exp).getTime() / 1000L;
+    }
+    return exp instanceof Number ? ((Number) exp).longValue() : 0L;
+  }
+
+  /**
+   * Returns the issued-at time in seconds since the Unix epoch.
+   */
+  public long getIssuedAt() {
```
weixifan: Same here.

src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
```diff
@@ -0,0 +1,103 @@
...
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Represents a verified Firebase App Check token.
+ */
+public class DecodedAppCheckToken {
```
weixifan, line 33: If this class is part of the public API, please also consider exposing `provider` and the optional claim `jti`.

src/main/java/com/google/firebase/appcheck/VerifyAppCheckTokenOptions.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * Sets whether to consume the token during verification.
+     *
+     * @param consume Optional boolean value.
+     * @return This builder.
+     */
+    public Builder setConsume(Optional<Boolean> consume) {
+      this.consume = consume != null ? consume : Optional.<Boolean>empty();
```
weixifan: In my opinion, allowing `null`s into an optional parameter re-exposes the exact same problem that `Optional<>` was intended to solve. I think it's better for us to check for nullness and throw in this case.

weixifan (commented):
Thank you for this PR! Some minor comments.

src/main/java/com/google/firebase/appcheck/internal/AppCheckTokenVerifier.java
yvonnep165: We don't use Guice or Dagger. Instead we use manual dependency injection via package-private `@VisibleForTesting` constructors and the services are singletons scoped to the `FirebaseApp` lifecycle.

src/main/java/com/google/firebase/appcheck/internal/AppCheckTokenVerifier.java
yvonnep165: Just tried to mirror the pattern from FirebasePhoneNumberVerificationTokenVerifier, which used double-checked locking for its `jwtProcessor`. But that's a good point! I updated `jwtProcessor` to be initialized directly in the constructor as a final field and removed the double checking.

src/main/java/com/google/firebase/appcheck/internal/AppCheckTokenVerifier.java
yvonnep165: Sure! added.

src/main/java/com/google/firebase/appcheck/internal/AppCheckTokenVerifier.java
yvonnep165: Comment added.

src/main/java/com/google/firebase/appcheck/internal/AppCheckTokenVerifier.java
yvonnep165: Thanks for catching that!

src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
yvonnep165: This followed the convention from `FirebaseToken` (auth) and `FirebasePhoneNumberVerificationToken`, where only `sub`  is strictly required in the constructor and the full claims validation (`iss`, `aud`, `exp`, `iat`) is handled by `AppCheckTokenVerifier` before the object is instantiated. Try to keep `DecodedAppCheckToken` as a flexible claims wrapper allows the getters to provide fallbacks (e.g. getAudience() returning []) and make creating mock tokens in tests much simpler. But let me know if you'd still prefer enforcing all of them here though!

src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
yvonnep165: Added `getJti()` and `getProvider()`.

src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
weixifan: I don't feel too strongly about this topic. Note that for these required claim names, the App Check backend is quite disciplined and will never return empty or null for them. For testing, it's reasonable to have an additional test-only constructor where these checks are not performed.

src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
```diff
@@ -57,6 +58,28 @@ public String getSubject() {
     return (String) claims.get("sub");
   }
 
+  /**
+   * Returns the App ID for which this token was issued.
+   * This is an alias for {@link #getSubject()}.
```
weixifan: Actually, I would prefer not promising that the `sub` claim is equal to the app ID, even though it is the case today.

src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
```diff
@@ -73,25 +96,17 @@ public List<String> getAudience() {
   }
 
   /**
-   * Returns the expiration time in seconds since the Unix epoch.
+   * Returns the expiration time as an {@link Instant}, or {@code null} if not present.
```
weixifan: You can actually promise that this is never `null`, if we perform the check for it. Similarly for `getIssuedAt` below.

src/test/java/com/google/firebase/appcheck/internal/AppCheckTokenVerifierTest.java
weixifan, line 107: Is it possible to write a couple of real token tests? Creating JWTs with a simulated public/private RSA key pair shouldn't be too difficult with `nimbusds`. You can create JWT with a private key, and you can mock the JWKS URL to return a JWKS containing your public key.

This isn't too urgent, and can definitely be added in a follow-up PR, but the test will be a *lot* stronger for having real token tests.

weixifan (approved):
Approved with only minor comments!

src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
yvonnep165: Ok it's removed.

src/test/java/com/google/firebase/appcheck/internal/AppCheckTokenVerifierTest.java
yvonnep165: Sure! added some real end-to-end cryptographic token verification tests.

src/main/java/com/google/firebase/appcheck/internal/AppCheckTokenVerifier.java
```diff
@@ -0,0 +1,301 @@
...
+        ApiClientUtils.newAuthorizedRequestFactory(app),
+        ApiClientUtils.getDefaultJsonFactory(),
+        null);
+  }
+
+  /**
+   * Package-private constructor designed explicitly for dependency injection
+   * of mock HTTP request factories, JSON factories, or mock JWT processors
+   * during isolated unit testing flows.
+   */
```
jonathanedey: This comment feels slightly misleading since it makes it seem that this is only used for testing.

src/main/java/com/google/firebase/appcheck/FirebaseAppCheckException.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+public class FirebaseAppCheckException extends FirebaseException {
+
+  public FirebaseAppCheckException(
+      @NonNull ErrorCode errorCode,
+      @NonNull String message,
+      @Nullable Throwable cause,
+      @Nullable IncomingHttpResponse response) {
+    super(errorCode, message, cause, response);
+  }
```
jonathanedey, line 49: Thinking we should also define app check specific client error codes like APP_CHECK_TOKEN_EXPIRED in node. WDYT?

src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
```diff
@@ -0,0 +1,120 @@
...
+  public Instant getExpirationTime() {
+    return toInstant(claims.get("exp"));
+  }
+
+  /**
+   * Returns the issued-at time as an {@link Instant}.
+   */
+  public Instant getIssuedAt() {
+    return toInstant(claims.get("iat"));
+  }
```
jonathanedey, line 107: I think it's good to use a more modern type here but we should make a note track this in future to eventually be consistent in other apis.

jonathanedey (commented):
Thanks for putting this together, this looks good with a few comments!

src/main/java/com/google/firebase/appcheck/internal/AppCheckTokenVerifier.java
yvonnep165: ok, removed.

src/main/java/com/google/firebase/appcheck/FirebaseAppCheckException.java
yvonnep165: Good point! I originally just stuck with the base ErrorCode to keep the exception API surface as minimal as possible and tried to avoid introducing extra enums if not strictly needed. But having dedicated codes like APP_CHECK_TOKEN_EXPIRED is better so developers can handle expired tokens programmatically without parsing error messages. I just added `AppCheckErrorCode` and changed the codes using it.

src/main/java/com/google/firebase/appcheck/DecodedAppCheckToken.java
yvonnep165: Ok, I'll file a ticket in our backlog to track adopting java.time.Instant across older Admin SDK APIs.

jonathanedey (approved):
LGTM, Thanks! Let's get the docs reviewed next!

kevinthecheung (approved):

PR #1239, opened 2026-09-15
 chore(deps): Bump org.apache.httpcomponents.client5:httpclient5

lahirumaramba:
Fixes the CI build failure in #1238 when bumping `org.apache.httpcomponents.client5:httpclient5` from `5.3.1` to `5.6.3`.

1. **Enforce Read Timeout in `ApacheHttp2Request`:**
   - **Issue:** Starting in HttpClient 5.4+, `InternalHttpAsyncExecRuntime` no longer sets socket-level timeouts on the endpoint when the protocol is HTTP/2 (`version >= HTTP_2`) because multiple multiplexed streams share a single connection. Consequently, HTTP/2 requests in `ApacheHttp2Request` were waiting indefinitely on `responseFuture.get()`, causing read timeout tests (`ApacheHttp2TransportIT.testReadTimeoutAuthorizedGet` and `testReadTimeoutAuthorizedPost`) to fail.
   - **Fix:** Stored the configured `readTimeout` in `ApacheHttp2Request` and applied it to `responseFuture.get(readTimeout, TimeUnit.MILLISECONDS)`. On timeout, the future is cancelled (`responseFuture.cancel(true)` sends an HTTP/2 stream reset) and an `IOException("Stream exception in request")` is thrown to maintain backward compatibility with SDK error contracts. Added a unit test `testReadTimeout` in `ApacheHttp2TransportTest`.

2. **Update Test `HttpServer` Initialization in `ApacheHttp2TransportTest`:**
   - **Issue:** HttpCore 5.4.3 introduced a new mandatory parameter (`HttpConnectionFactory`) to the low-level `HttpServer` constructor, causing a compile failure in `FakeServer`.
   - **Fix:** Migrated `FakeServer` to construct the server using `ServerBootstrap.bootstrap()`.


src/test/java/com/google/firebase/internal/ApacheHttp2TransportTest.java
```diff
@@ -434,31 +433,24 @@ public HttpRequestHandler resolve(HttpRequest request, HttpContext context)
...
+                    HttpRequest request, EntityDetails entity, HttpContext context)
+                    throws HttpException, IOException {
+                }
+
+                @Override
+                public void process(
+                    HttpResponse response, EntityDetails entity, HttpContext context)
+                    throws HttpException, IOException {
+                }
+              })
```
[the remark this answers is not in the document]
lahirumaramba: Addressed

src/main/java/com/google/firebase/internal/ApacheHttp2Request.java
```diff
@@ -142,6 +148,9 @@ public void cancelled() {
       throw new IOException("Request Interrupted", e);
     } catch (CancellationException e) {
       throw new IOException("Request Cancelled", e);
+    } catch (TimeoutException e) {
+      responseFuture.cancel(true);
+      throw new IOException("Stream exception in request", e);
```
jonathanedey, line 153: Should we specify that it was a read timeout here?

jonathanedey (approved):
LGTM with one nit, Thanks!

src/main/java/com/google/firebase/internal/ApacheHttp2Request.java
lahirumaramba: Yeah good call! I thought about it but I wonder if that would break the backward compatibility in some way... even if it did, it would most likely be a minor risk I think

lahirumaramba: Let's leave it like this for now because our priority is to get the dependency bumped. We can address that in a future PR.

PR #1240, opened 2026-09-17
feat(fcm): Migrate topic management to FCM v1 API

lahirumaramba:
Migrates topic subscription and management from the legacy Instance ID (IID) service (`iid.googleapis.com`) to the FCM v1 REST API (`fcm.googleapis.com/v1/projects/{projectId}/registrations/{token}/topicSubscriptions`).


OrlandriaH-G (approved):

src/main/java/com/google/firebase/messaging/FirebaseMessagingClientImpl.java
```diff
@@ -182,17 +223,187 @@ public void initialize(HttpRequest request) throws IOException {
...
+        ? topic.substring("/topics/".length()) : topic;
+
+    List<CompletableFuture<TopicResult>> futures = new ArrayList<>();
+    for (int i = 0; i < registrationTokens.size(); i++) {
+      final int index = i;
+      final String token = registrationTokens.get(i);
+      futures.add(CompletableFuture.supplyAsync(
+          () -> sendSingleTopicRequest(token, topicName, isSubscribe, index),
+          this.executor));
+    }
```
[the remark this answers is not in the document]
lahirumaramba: Addressed

src/main/java/com/google/firebase/messaging/FirebaseMessagingClientImpl.java
```diff
@@ -87,6 +104,20 @@ private FirebaseMessagingClientImpl(Builder builder) {
...
+
+  private static ExecutorService createDefaultExecutor(ThreadFactory threadFactory) {
+    ThreadPoolExecutor pool = new ThreadPoolExecutor(
+        100, 100,
+        60L, TimeUnit.SECONDS,
+        new LinkedBlockingQueue<Runnable>(),
+        threadFactory != null ? threadFactory : Executors.defaultThreadFactory());
+    pool.allowCoreThreadTimeOut(true);
+    return pool;
   }
```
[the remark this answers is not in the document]
lahirumaramba: Addressed

src/main/java/com/google/firebase/messaging/FirebaseMessagingClientImpl.java
```diff
@@ -87,6 +106,27 @@ private FirebaseMessagingClientImpl(Builder builder) {
...
+        .setNameFormat("firebase-messaging-topics-%d")
+        .setDaemon(true)
+        .build();
+    ThreadPoolExecutor pool = new ThreadPoolExecutor(
+        100, 100,
+        60L, TimeUnit.SECONDS,
+        new LinkedBlockingQueue<Runnable>(),
+        factory);
+    pool.allowCoreThreadTimeOut(true);
+    return pool;
```
jonathanedey, line 129: I'm hesitant on adding a nested thread executor here. The existing implementation already uses an app level thread manager which developers can configure through options. Adding another layer of threads without developer control could have be a negative side effect of more thread overhead than the developer is expecting.

src/main/java/com/google/firebase/messaging/FirebaseMessagingClientImpl.java
```diff
@@ -182,17 +232,192 @@ public void initialize(HttpRequest request) throws IOException {
...
+    int successCount = 0;
+    List<TopicManagementResponse.Error> errors = new ArrayList<>();
+    for (CompletableFuture<TopicResult> future : futures) {
+      TopicResult result = future.join();
+      if (result.isSuccess()) {
+        successCount++;
+      } else {
+        errors.add(new TopicManagementResponse.Error(result.getIndex(), result.getReason()));
+      }
+    }
```
jonathanedey, line 277: I think we solved a similar issue without a new thread pool with sendEachAsync. Are there advantages to this method over the sendEachAsync way? 

https://github.com/firebase/firebase-admin-java/blob/db48c53599397561382d92fc31228d69709e803b/src/main/java/com/google/firebase/messaging/FirebaseMessaging.java#L236-L247

src/main/java/com/google/firebase/messaging/FirebaseMessagingClientImpl.java
```diff
@@ -182,17 +232,192 @@ public void initialize(HttpRequest request) throws IOException {
...
+        case 500:
+          return "INTERNAL";
+        case 503:
+          return "DEADLINE_EXCEEDED";
+        default:
+          return "UNKNOWN_ERROR";
+      }
+    }
+    return "UNKNOWN_ERROR";
+  }
```
jonathanedey, line 369: Same as others languages. Not blocking but I don't think we need to reparse the firebase error. We can keep the custom mapping if nessasary for matching the status codes error codes correctly.

jonathanedey (commented):
LGTM, but with some push back on the new thread pool.

src/main/java/com/google/firebase/messaging/FirebaseMessagingClientImpl.java
```diff
@@ -182,17 +232,192 @@ public void initialize(HttpRequest request) throws IOException {
...
+        case 403:
+          return "PERMISSION_DENIED";
+        case 404:
+          return "NOT_FOUND";
+        case 429:
+          return "RESOURCE_EXHAUSTED";
+        case 500:
+          return "INTERNAL";
+        case 503:
+          return "DEADLINE_EXCEEDED";
```
yvonnep165, line 363: is 503 "Service Unavailable" and 504 "DEADLINE_EXCEEDED"?

src/main/java/com/google/firebase/messaging/FirebaseMessagingClientImpl.java
lahirumaramba: Thanks for point this here and in the above comment. I will change it to a similar method we used in sendEach

jonathanedey (approved):
LGTM, Thanks!

